Issue #16, opened 2019-12-18
Tests and test infrastructure


chriseth:
This repository has no tests. This is a no-go.

cgewecke:
Close-able via #31, #43

edisinovcic:
@cgewecke I agree on this. @chriseth do you think this is enough?

PR #31, opened 2020-01-20
Add initial test fixtures and CI

cgewecke:
#16 

PR adds:

+ CircleCI 
+ Test fixtures: solidity files and contract artifacts for building test cases
+ Mocha and stub tests for most files. 
+ Development notes in README

There are two passing tests:
+ a cborDecode test which shows how to synthesize the IPFS hash of a metadata string and extract the IPFS hash from deployedBytecode. 
+ a test which deploys a contract to ganache and calls `injector.inject` to verify & store the source in the repository

Also updated injector.js to decode IPFS.

[1]: https://circleci.com/gh/cgewecke/source-verify/5


injector.js
```diff
@@ -142,8 +151,8 @@ let storeData = function(repository, chain, address, compilationResult, sources)
   } else if (cborData['bzzr1']) {
     metadataPath = '/swarm/bzzr1/' + Web3.utils.bytesToHex(cborData['bzzr1']).slice(2)
   } else if (cborData['ipfs']) {
-    throw "TODO: Implement ipfs"
-    //metadataPath = '/ipfs/' + Web3.utils.bytesToHex(cborData['ipfs']).slice(2)
+    const bytecodeLink = new dagPB.DAGLink('', 0, cborData['ipfs']);
```
chriseth: What does this result in?

test/injector.js
```diff
@@ -0,0 +1,88 @@
...
+
+      const options = {
+        data: Simple.compilerOutput.evm.bytecode.object,
+        gasPrice: '1',
+        gas: 4000000,
+      };
+
+      // Deploy contract
+      const simple = new web3.eth.Contract(abi, options);
+      const instance = await simple.deploy().send({from: accounts[0]});
```
chriseth: This is only needed so that the bytecode query returns the right value, is that true?
If so, might it make sense to just mock that json-rpc call?
My concern is that actually deploying a cotnract might also trigger the monitor and not just the injector.

package.json
```diff
@@ -28,10 +31,19 @@
     "express": "^4.17.1",
     "express-fileupload": "^1.1.6-alpha.6",
     "html": "^1.0.0",
+    "ipld-dag-pb": "^0.18.2",
     "request": "^2.88.0",
     "request-promise-native": "^1.0.7",
     "serve-index": "^1.9.1",
     "solc": "^0.5.7",
```
chriseth: We should increase this to 0.6.0.

chriseth:
Looks good otherwise!

injector.js
cgewecke: The link is an object like this:
```
DAGLink {
  _name: '',
  _nameBuf: null,
  _size: 0,
  _cid:
   CID {
     version: 0,
     codec: 'dag-pb',
     multihash:
      <Buffer 12 20 27 dd 4c f1 4e 24 03 b8 43 ff 81 26 5c ce a9 0b d3 be 07 ef c0 d0 bb 7c 2d 8c e2 76 fa c7 df 58>,
     multibaseName: 'base58btc' } 
}
```

It looks like the conversion could be done more simply (without creating a link) with: 
```javascript
multihashes.toB58String(cborData['ipfs']);

> QmR2GL3J5ggeEKusoFgtkRNzKd5BXJt43nYQoWN9Ay9hXD
```

**Libs**
+ [multihashes](https://github.com/multiformats/js-multihash#js-multihash)
+ [ipld-dab-pb DAGLink docs](https://github.com/ipld/js-ipld-dag-pb#daglink-functions).

test/injector.js
cgewecke: Yes, true. The monitor is not running in these tests. Agree it's important to isolate them though. 

I'd rather not mock unless necessary because it requires that the test know about internal implementation details of the method it's testing.

package.json
cgewecke: Upgraded to ^0.6.1 in 754add8a7569cf7bcfdf1ec4d8600a63f51cac03

injector.js
cgewecke: Simplified to use only the multihashes form in 754add8a7569cf7bcfdf1ec4d8600a63f51cac03

test/sources/compiler.json
```diff
@@ -0,0 +1,17 @@
+{
+  "contracts": "*",
+  "solcVersion": "0.6.0",
```
marcgarreau: Should this value and all contract pragma values track with solc in package.json? i.e. `0.6.1`

cgewecke: Hi @marcgarreau   

It could - intended this config to be a convenience that people can adjust the values of when they produce a test case. The version specified here gets fetched from a remote - there's no conflict with the package.json solc - it's separate. 

The pragmas float on `pragma solidity ^0.6.0` - think this should be ok for any 6 series compiler.

Are you seeing an error?

marcgarreau: nope, just double-checking :shipit:

test/injector.js
chriseth: Ok!

PR #38, opened 2020-01-22
Adding health endpoint

edisinovcic:
This is a first version of health endpoint and docker-compose and dockerfile. It's not optimized version but it's quick and dirty first implementation. Please test it and let me know if there are any problems. 

Health endpoint has been added, will soon also add a service that pings it every 60 seconds.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.idea
+.vscode
 .DS_Store
 db/
 node_modules/
--- a/Dockerfile.healthcheck
+++ b/Dockerfile.healthcheck
@@ -1,0 +1,4 @@
+FROM byrnedo/alpine-curl
+WORKDIR /usr/src/app
+COPY health_check.sh ./health_check.sh
+CMD ["./health_check.sh"]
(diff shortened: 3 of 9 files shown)


chriseth:
The target branch seems wrong.

edisinovcic:
Now it's targeting master and I've resolved merge conflicts.

chriseth:
Can you rebase and squash a little?

edisinovcic:
Definitely. :)

injector.js
```diff
@@ -205,5 +205,13 @@ exports.inject = async function(repository, chain, address, files) {
...
+<<<<<<< HEAD
 
 exports.cborDecode = cborDecode;
+=======
+<<<<<<< HEAD
+=======
+
+exports.cborDecode = cborDecode;
+>>>>>>> b70ee4a... Add initial test fixtures and CI
+>>>>>>> 0a02b60... Adding docker-compose for the server and monitor and adding health endpoint to server
```
cgewecke: There are a few stray merge conflict markers here :)

chriseth:
The issue with the tests looks like a problem on CircleCI's side (password-protected ssh key).

In other projects, we use the policy to never have a merge commit as part of a PR because it complicates the review - shall we adopt this policy also here?

injector.js
edisinovcic: Tnx!

edisinovcic:
We could, how do you solve conflicts when they appear? Do you manually change code that is conflicting without pulling branch that it should be merged into?

Issue #30, opened 2020-01-14
Nicer file upload


chriseth:
The look and feel of the file upload page should be improved. We should have a drag and drop area and in general, better input fields.

marcgarreau:
In flight. Got a minimalist react app started with file drag and drop. Nothing being pulled from CDNs. Still wiring it up and touching up the styling, but will hold off on PRs until #31 is merged to establish a baseline.

FYI, don't have any permissions on this repo, so unable to move items on the project kanban board manually.

Issue #29, opened 2020-01-14
Stable multi-file upload


chriseth:
The multi-file upload is currently rather unstable, it works on some browsers sometimes, but not always. It might be necessary to run custom ajax posts.

chriseth:
@marcgarreau maybe we should chat about this issue.

PR #39, opened 2020-01-25
Introduces React UI

marcgarreau:
## What does it do?
This introduces a small React app for the uploader UI. This includes packages for some UI niceties, e.g. files drag and drop functionality and a slick select input. Addresses #30, #29.

## Testing workflow items
- To run the UI alone: `npm run dev:ui`
- To run the UI and server locally: `npm run dev:server`
- To run the UI as a part of the docker env: `npm run dev`

## TODOs
- Finish wiring up; potentially an issue with sample files I'm using, but unable to get an end-to-end success yet. Help testing, crossing the finish line appreciated, cc/ @cgewecke, @edisinovcic 
- Have docker handle the UI build step. cc/ @edisinovcic I know you've got a compose workflow in another PR, so I didn't invest much attention here.
- File explorer feature. Routing hasn't been introduced yet, so this feature is absent for now.
- Small UI enhancements, e.g., disabled button state, etc.
- Clean up some comment cruft, etc.
- Note: some styling changes were performed by my system instance of [prettier](https://prettier.io). For consistency's sake, I always advocate for its adoption.

### before:
Screen Shot 2020-01-24 at 2 59 13 PM

### after:
Happy path:
Screen Shot 2020-01-29 at 4 56 58 PM

Error state:
Screen Shot 2020-01-25 at 12 20 06 PM

Multiple:
Screen Shot 2020-01-29 at 5 27 45 PM

Loading state:
Screen Shot 2020-01-29 at 5 11 52 PM 1


injector.js
```diff
@@ -96,9 +96,9 @@ let recompile = async function(metadata, sources) {
 let findMetadataFile = function(files) {
   for (let i in files) {
     try {
-      let m = JSON.parse(files[i])
-      if (m['language'] === 'Solidity') {
-        return m
+      const file = JSON.parse(JSON.parse(files[i]))
```
cgewecke: Is this double parse required?

README.md
```diff
@@ -55,18 +61,20 @@ modifications have to be prepared at deploy time, but it is still a possibility.
...
+- perform source verification given only an address instead of the bytecode
+  or the metadata
 
 ## Development
 
 ### Launch service
+
 ```
-./run_sh 80 repository db.
+./run.sh 80 repository db
```
cgewecke: 👍

chriseth:
Could you maybe split this into multiple PRs? At least the style changes could be a different commit.

injector.js
marcgarreau: In my testing, unfortunately yes, but that was an item to revisit. Presumably that can get cleaned up further up the stack.

marcgarreau:
@chriseth unfortunately this change set doesn't lend itself to multiple PRs very well. The conversion to a single-page app can't help but touch a large surface area. Re: styling, I can apply some styling to the existing html page, but a decent number of elements being styled are introduced by the React components.

Taking a step back: if you think an SPA is overkill and what you really want is just some styling, we can throw this out. We can chat about that at our meeting in 25 min.

injector.js
marcgarreau: @cgewecke From what I'm seeing (using your example test files), metadata files come in "over-stringified", i.e., the first pass of JSON.parse() still returns a string. 

The existing, deployed proof of concept on verification.komputing.org appears to produce the same results when using these files; my tests produce the same `Metadata file not found` error. Possibly @chriseth has tested with files that have been generated via another method?

In any event, I reverted this change, but moved some conditional logic into `server.js` to parse the file if it's able. I believe there's no "harm" in double parsing, but its arguably cleaner now. Open to other suggestions.

cgewecke: @marcgarreau  I think your solution is the right one... my understanding of this is that:

when the metadata is generated using solc's JSON-IO interface and saved to file as part of an artifact, it's stringified and looks like:

```
"{\"compiler\":{\"version\":\"0.6.0+commit.26b70077\"},\"langua....
```

If you load the artifact and query its type
```js
typeof compilerOutput.metadata
> string
```

Uploadeding the string to IPFS with `ipfs.add` will locate it at the same hash as the one encoded in deployedBytecode. The monitor uses it to find metadata for newly deployed contracts.

On the other hand, when metadata is generated at the cli with `solc --metadata -o metadata.json`, the file contents look like:
```
{"compiler":{"version":"0.6.0+commit.26b70077"},"langua...
```

The cli version is used to organize recompilation and compare its outcome with the deployedBytecode at the user supplied contract address. 

So it looks like there are two use cases and two formats - the second is normal for the UI but maybe it makes sense to support both versions there anyway?

localchain.js
```diff
@@ -0,0 +1,78 @@
...
+  log(`====================`);
+}
+
+// Main
+main()
+  .then()
+  .catch(err => {
+    console.log(err);
+    process.exit(1);
+  })
```
cgewecke:
```suggestion
  })

```

(Missing newline, mea culpa)

ui/App.js
```diff
@@ -0,0 +1,166 @@
+import React, { useState } from 'react'
+import { useDropzone } from 'react-dropzone'
+import Select from 'react-select'
+
+const options = [
+  { value: 'mainnet', label: 'Ethereum Mainnet' },
```
edisinovcic: Moving this to constants file (or json) could be better so contents can be updated more easily in the future. It's nothing crucial, just a style of coding.

marcgarreau:
@cgewecke I dug around for support for your sequential adding scenario. I'm fairly sure this could be custom implemented, but I assume native support doesn't exist, because the functionality mimics a raw html file input (testable at verification.komputing.org).

ui/App.js
marcgarreau: Agree the practice is a good one. My instinct is that adding another file is overkill for now, since there's no foreseeable reuse - but I think a good middle ground is to bring it inside of the component to clearly convey it is only used there.

edisinovcic: Okay, sounds good

Issue #35, opened 2020-01-20
Add test for monitor.js 


cgewecke:
Would be nice to have a check of this before restructuring the project and converting to typescript. Might look like:

1. Launch testrpc
2. Launch monitor, watching localhost
3. Launch local ipfs node --offline
4. Deploy contracts to testrpc
5. Publish contracts to local ipfs node
6. Verify address is put in addressDB
7. Verify sources are published to repository

cgewecke:
#40

PR #40, opened 2020-01-27
Make monitor JS class, add monitor test

cgewecke:
#35

PR 
+ Re-structures `monitor.js` as a JS class with stop and start methods. Enables IPFS.
+ Rewrites `run_monitor.sh` to use a node script that starts/stops the monitor
+ Adds a multi-source / IPFS-based test for the monitor where:
  +  a contract is deployed to testrpc, 
  + its sources and metadata are published to a local offline IPFS node,
  + the monitor discovers everything and saves it. 

Several unused methods are removed from `monitor.js` (these might need to be restored /wired in.)
+ recompile
+ reformatMetadata
+ getBytecodeMetadataAndSources
+ retrieveSingleSource
+ retrieveSources


run_monitor.sh
```diff
@@ -1,4 +0,0 @@
-#!/bin/sh
-
-echo "Starting monitor..."
-while true; do node ./monitor.js /repository/; done
```
chriseth: The benefit of this shell script is that it would re-start the process even in out-of-memory situations.

cgewecke: Ah ok!

cgewecke: Restored in 15232cd9bf408aeb759bb36239521d81f5e1c897

PR #44, opened 2020-02-04
Only allow localhost network when dev ENV variable is set

cgewecke:
As discussed in the Feb 3 meeting, the localhost:8545 client should not be a valid option in the production UI build.

PR 
+ makes the ui's localhost option conditional on `NODE_ENV === 'development'` 
+ fixes a syntax error `ui/App.js` that looks like it might have been accidentally introduced while renaming `options` to `chainOptions.` 
+ adds a README section about UI testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,4 +120,26 @@ npx sol-compiler
 
 Compilation artifacts will be written to an `artifacts` folder.
 
+### UI Testing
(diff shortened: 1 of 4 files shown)


cgewecke:
Superceded by #46

Issue #32, opened 2020-01-20
Restructure repo


cgewecke:
One way the project could be restructured prior to converting to TS might be:
```
.circleci
Dockerfile
public /
src/
  address-db.js
  injector.js
  monitor.js
  server.js
  utils.js  # Duplicate fns
scripts/
  run.sh
  run_monitor.sh
  run_server.sh
test /
  .. etc ..
```
Some files seem experimental (or incidental) and might be removed / stored away.
+ index.js (?)
+ loop.sh
+ verify.js 

Shell scripts could be aliased to npm scripts
```json
"scripts": {
    "docker:all": "./scripts/run.sh",
    "monitor": "./scripts/run_monitor.sh",
    "server": "./scripts/run_server.sh",
}
```

chriseth:
Sounds good! Yes, index, loop and verify can be removed, although verify could be a stand-alone tool.

cgewecke:
#45

PR #45, opened 2020-02-04
Initial file re-org & convert injector to class

cgewecke:
#32


cgewecke:
From meeting (Feb 14): move `ui/` into `src`

scripts/localchain.js
```diff
@@ -10,8 +10,8 @@ const ganache = require('ganache-cli');
 const pify = require('pify');
 const Web3 = require('web3');
 
-const Simple = require('./test/sources/pass/simple');
-const SimpleWithImport = require('./test/sources/pass/simpleWithImport')
+const Simple = require('./../test/sources/pass/simple');
+const SimpleWithImport = require('./../test/sources/pass/simpleWithImport')
```
edisinovcic: Should work without ./ at the begining

cgewecke:
@edisinovcic This is rebased against `staging` now, and #53 is rebased as well - it should go into staging clean. 

But this is a weird intermediate PR - there are changes in #53 to some of the paths rewritten here because TS is getting written to dist, etc. 

I don't know if you'd prefer to just merge this into staging (or even close) and review #53 instead? It's up to you.

cgewecke:
Rebased for #57 (Continuous deployment)

Issue #36, opened 2020-01-20
Convert code to Typescript


cgewecke:

cgewecke:
via #53. (The main application files are all in TS - the tests remain in JS for the time being)

PR #53, opened 2020-02-17
Add typescript

cgewecke:
#36 Typescript

+ Migrates the main application files to typescript using a permissive tsconfig which allows javascript. Test files, ui and js scripts remain in js. 
+ Adds tslint step to CI
+ All the types / ts dependencies are added as dev deps. Did this on assumption that project will be published as a docker. `tsc` is run during the docker initialization & the application consumes its code as transpiled JS from the`dist/` folder.
+ Unit tests are run using ts-node, directly from `./src`

Running: 
```
docker-compose -f docker-compose-testing.yaml up --build
```
...launches a server available at `localhost:1234` and a local ganache instance with deployed contracts that can be used for manual testing. Docker output including contract addresses is visible in the terminal.


docker-compose-testing.yaml
```diff
@@ -43,20 +43,20 @@ services:
...
-      dockerfile: Dockerfile.localchain    
+      dockerfile: Dockerfile.localchain
     ports:
     - "${LOCALCHAIN_EXTERNAL_PORT}:${LOCALCHAIN_PORT}"
     networks:
     - source-verify-network
     restart: always
   repository:
-    image: ethereum/source-verify:repository-${TAG}
+    image: blockchainit/source-verify-repository:${TAG}
```
cgewecke:
```suggestion
    image: ethereum/source-verify:repository-${TAG}
```

README.md
```diff
@@ -75,7 +81,7 @@ Docker-compose (https://docs.docker.com/compose/install/)
 
 ### How to run
 Prepare environment and start by running
-`STAGE=prod ./prepare_and_start.sh` for production 
+`STAGE=prod ./prepare_and_start.sh` for production
 `STAGE=testing ./prepare_and_start.sh` for testing.
```
cgewecke: @edisinovcic Possibly this `How to run` section should be about production only? Mostly because prepare_and_start copy files around in the repo and these end up in the git diff. 

In this PR, I modified docker-compose-testing.yaml to use env.testing directly and using the original `up` command as described below seemed to launch everything correctly for manual testing...

edisinovcic: The only problem with it is that docker-compose uses .env file to enter values in docker-compose and that .env.testing environment is send to dockerfiles during build. There is no way to say docker-compose that it needs to use .env.something while setting up parameters in docker-compose.yaml file. We can maybe add to .gitignore docker-compose.yaml and .env so they can be created. 

There is no difference as ports are the same on .env.prod and .env.testing currently. 

Diff is definitely a problem and those changes are not needed there. I'm open to proposals here if you think this can be done better. :)

edisinovcic: Would something like this be better? https://github.com/ethereum/source-verify/pull/55

src/address-db.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+// TODO make the file name configurable.
+
+import fs from 'fs';
+import util from 'util';
+import Web3 from 'web3';
+
+const log = console.log;
+const dbFile : string = '/db/addressDB.txt';
```
edisinovcic: Move this to environment?

README.md
cgewecke: > There is no way to say docker-compose that it needs to use .env.something while setting up parameters in docker-compose.yaml file. 

Ah I see..

src/injector.ts
```diff
@@ -0,0 +1,237 @@
...
+   * If environment variable TESTING is set to true, localhost:8545 is also available.
+   */
+  private initChains(){
+    for (const chain of ['mainnet', 'ropsten', 'rinkeby', 'kovan', 'goerli']){
+      this.chains[chain] = {};
+      this.chains[chain].web3 = new Web3(`https://${chain}.infura.io/v3/${this.infuraPID}`);
+    }
+
+    // For unit testing with testrpc...
+    if (process.env.TESTING){
```
cgewecke: This should be moved to a variable in the Injector config and set via process.env in the unit tests and server.

src/injector.ts
```diff
@@ -0,0 +1,237 @@
...
+      this.chains['localhost'] = {
+        web3: new Web3(<string>process.env.LOCALCHAIN_URL)
```
cgewecke: Same...

src/address-db.ts
cgewecke: I've ended up just commenting out the address-db references in injector.ts because there's no "write" operation to the db anywhere. 

I think this can be deferred until the second iteration - we'll need to add tests, make sure the volume is mounted etc etc..

cgewecke:
@edisinovcic This is all rebased and should be ready for a final look through I guess...

cgewecke:
Rebased for #57 via #45

src/address-db.ts
cgewecke: Tracking this #58

cgewecke:
@edisinovcic I'm merging this to staging per gitter conversation but it that's not what you intended we can always revert :)

Issue #27, opened 2020-01-13
Check which tools use absolute paths for sending sources to the compiler


chriseth:
Related to #24

cgewecke:
List of a few ... 

(Where the tools do not write metadata to their artifacts, have gone into the code and added it to the output selection to see what would happen.)

| Tool | Company    | Version    | Path Type   |  Outputs metadata |
| ---------- | ------- | ------- | ------- | ------ |
|  [buidler][1]  | NomicLabs | 1.0.2  | relative | No  |
|  [oz/cli][4] | OpenZeppelin | 2.6.0 | relative | No |
|  [truffle][2]  | TruffleSuite| 5.1.2  | absolute | Yes  |
|  [sol-compiler][3] | 0xProject | 4.0.3 | absolute | Yes (configurable) |

[1]: https://buidler.dev/guides/compile-contracts.html
[2]: https://www.trufflesuite.com/docs/truffle/reference/configuration#compiler-configuration
[3]: https://sol-compiler.com/
[4]: https://docs.openzeppelin.com/cli/2.6/compiling#running_the_compiler

Issue #24, opened 2020-01-09
Paths in metadata


cgewecke:
(Following up on discussion in the Jan 8 source-verify meeting)

Am wondering about how the metadata is hashed and the importance of the source file paths. (Apologies in advance if I've misunderstood this...)

In the solidity docs there is a warning:

> Since the bytecode of the resulting contract contains the metadata hash by default, any change to the metadata might result in a change of the bytecode. This includes changes to a filename or path, and since the metadata includes a hash of all the sources used, a single whitespace change results in different metadata, and different bytecode.

For example, A.sol, compiled with `truffle compile` and deployed using the data in the truffle artifact, has metadata source keys like:

```json
"sources": {
  "/Users/cgewecke/code/ef/sv-truffle/contracts/A.sol": {
   "keccak256": "0xfc68a9ffdb0d7d3ec5456f5e2d2c4f855e684892f6c78fcb0dc1e482696d9bb7",
   "urls": [
    "bzz-raw://4d22457fe6215d956a57fcd0cb63f745d8fd718ba8d85a20d65c72232038779f",
    "dweb:/ipfs/QmPmUahXAjtq5mJTW9Qc6oNUVGEuocxb1un5371zLJAAb5"
   ]
  }
},
```
 
Running `solc --metadata contracts/A.sol` over the same source produces

```json
"sources": {
  "contracts/A.sol": {
   "keccak256": "0xfc68a9ffdb0d7d3ec5456f5e2d2c4f855e684892f6c78fcb0dc1e482696d9bb7",
   "urls": [
    "bzz-raw://4d22457fe6215d956a57fcd0cb63f745d8fd718ba8d85a20d65c72232038779f",
    "dweb:/ipfs/QmPmUahXAjtq5mJTW9Qc6oNUVGEuocxb1un5371zLJAAb5"
   ]
  }
},
```

It seems like the file system context and tooling of the deployment and metadata compilations are not allowed to differ even when the sources are identical.

Not sure this would work but perhaps if the file's keccak hash was used as a key instead, metadata could be matched to sources from anywhere as long as paths were associated with the files when uploading to the service for JSON-IO compilation. 

If metadata was pathless, you could:
+ have a source-verify cli tool set up to run in CI on merge to a release branch
+ have a bot that searched Github for Solidity projects, cloning and compiling them, uploading all the metadata to SV to build a large sources DB.

chriseth:
[...]

Issue #23, opened 2020-01-09
CLI 


cgewecke:
(Following up on discussion in the Jan 8 source-verify meeting)

Source submission could be simplified by publishing a cli utility. Buidler and Truffle both support plugins and collecting the sources / generating the metadata would be easy to implement in that format. 

**Install**
```shell
$ npm install --save-dev source-verify
```

**Configure**
```javascript
// truffle-config.js
module.exports = {
  plugins: ['source-verify'],
  networks: {...}
}
```

**Run**
```shell
$ truffle run source-verify

> Generating metadata... 
> ======================
> - A.sol
> - B.sol
> 
> Uploading
> =========
> Your sources should be verified in a few minutes. Check at:
> Metadata: https://gateway.pinata.cloud/ipfs/QmYVd8qstdXtTd1quwv4nJen6XprykxQRLo67Jy7WyiLMB
> Sources: https://gateway.pinata.cloud/ipfs/QmYVd8qiwendlgdsfnJen6XprykxQRLo67Jy(WyiLMC

$ truffle run source-verify --fetch QmYVd8qstdXtTd1quwv4nJen6XprykxQRLo67Jy7WyiLMB

> {
>  ...metadata
> }
```

**Other**

Check if the local version of a contract matches one on chain (to validate a simulation).
```shell
truffle run source-verify \
  --diff \
  --local build/contracts/ENSRegistry.json \
  --network mainnet \
  --address 0xabc....de \

> - 0.4.23+commit.124ca40d // Mainnet
> + 0.5.12+commit.7709ece9 // Local
```

chriseth:
The minimal requirement here is that truffle uploads the metadata and source files to ipfs or swarm around the time the contract is deployed to mainnet. The monitor service in source-verify will pick it up itself. Of course, truffle can also manually upload that data directly to source-verify. This might be more reliable but can only be done after the contract has been deployed to avoid DoS-ing source-verify (it only accepts metadata files and sources of contract that are on chain).

What we can provide to external services by means of an npm package is the re-compilation: Just the fact that the hash in the bytecode matches the metadata file does not mean that the bytecode itself has been produced by the source files linked from the metadata file. To be 100% sure you have to recompile the sources with the correct compiler version and compare the bytecode.

As you already show in the example, "matching sources" is not the only fact that is necessary - another important aspect is that the used and configured compiler versions are the same, as well as all the compiler settings.

edisinovcic:
[...]

PR #56, opened 2020-02-24
[WIP] Feature validator

edisinovcic:
Current naive implementation of the validator. Implementation still needs to be discussed.

DO NOT MERGE!

### DIFF
--- a/injector.js
+++ b/injector.js
@@ -163,6 +163,14 @@ let storeData = function(repository, chain, address, compilationResult, sources)
   }
 }
 
+exports.validate = async function(files) {
+  let metadata = findMetadataFile(files)
+  let sources = rearrangeSources(metadata, files)
+  // Starting from here, we cannot trust the metadata object anymore,
+  // because it is modified inside recompile.
+  return await recompile(metadata, sources)
+}
+
 exports.inject = async function(repository, chain, address, files) {
   if (address) {
     address = Web3.utils.toChecksumAddress(address)
--- a/validator.js
+++ b/validator.js
@@ -1,0 +1,40 @@
+#!/usr/bin/env node
+
+'use strict';
+
+if (process.argv.length < 3) {
+    console.log("Usage: validator.js <file paths>")
+    process.exit(1)
+}
+
+const fs = require('fs');
+
+let injector = require('./injector');
+
+function loadFiles(filePaths){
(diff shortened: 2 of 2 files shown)


edisinovcic:
@cgewecke can you please try to test it locally, as when I remove JSON.parse on loadFiles level I can't seem to get positive results from this function: https://github.com/ethereum/source-verify/blob/03a6bf4a7b3a9cd3e2eea092732894f7a567d3b8/injector.js#L96

I'm using files from test/sources/all/simple.meta.json and Simple.sol

cgewecke:
I read through the [relevant thread][1] about this at Solidity and it looks like the person wants to verify a contract on-chain by cloning a Truffle project on Github, compiling the contracts and checking if the metadata produced matches the hash of the bytecode deployed to chain. 

There are difficulties with this because of the way Truffle compiles using absolute paths as source keys - something discussed in #23, #24, and #27. I'm not sure this PR is really the validation they're looking for unfortunately ... 

Perhaps we should investigate this in more depth as part of a second iteration where we try to make a truffle plugin or general lib for cli tools that could handle this use case. 

[1]: https://gitter.im/ethereum/solidity?at=5e53c4a3fa9f20553b4a0e88

Reproducing part of the thread here for reference:
```
So here is how it works-:
I compiled contracts on my machine and deployed it on the chain. 
Which means metadata associated with the deployment is related to my machine
Now user wants to verify that the contract that they are going to interact with actually 
corresponds to our open source repository
1. They clone the repository
2 .They compile the contracts using their machine. Now in this case metadata will change
3. Now they verify both bytecodes

The verification process should work out of the box, irrespective of the machine and environment used for the actual compilation and deployment
Though the compiler version and other compiler settings should remain same
```

edisinovcic:
I definitely agree. Let's keep it open and discuss it after ethcc to see how to solve this.

edisinovcic:
This was merged by mistake :( @cgewecke should I remove it from staging?

cgewecke:
@edisinovcic I undid it in an earlier rebase it's all fine :)

edisinovcic:
Awesome, tnx!

aaronmboyd:
> I definitely agree. Let's keep it open and discuss it after ethcc to see how to solve this.

Is there a solution for this?

PR #68, opened 2020-03-02
Use branch specific env files when publishing

cgewecke:
**PR targets 59**

Uses env.latest when publishing to the staging branch, env.build when publishing master.

### DIFF
--- a/scripts/build_and_publish_docker_images.sh
+++ b/scripts/build_and_publish_docker_images.sh
@@ -4,19 +4,20 @@ set -e
 # If not staging and master branch are existing
 export TAG="$CIRCLE_BRANCH"
 
-if [ "$CIRCLE_BRANCH" == "staging" ]; then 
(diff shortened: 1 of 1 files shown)


scripts/build_and_publish_docker_images.sh
```diff
@@ -4,19 +4,20 @@ set -e
...
+if [ "$CIRCLE_BRANCH" == "staging" ]; then
     export TAG="latest"
+    cp .env.latest .env
     echo $TAG
 fi
 
 if [ "$CIRCLE_BRANCH" == "master" ]; then
-    export TAG="stable"; 
+    export TAG="stable";
+    cp .env.build .env
```
cgewecke: @edisinovcic Do you think this should this be `build` or `stable`? At the moment staging is published with build.

edisinovcic: `cp .env.${TAG} .env
should fix it

PR #86, opened 2020-03-16
Adding verify.js and npm publish

edisinovcic:
Adding verify.js back to the project and also an ability to publish to npmjs.

Resolves this: https://github.com/ethereum/source-verify/issues/79


src/verify/verify.js
```diff
@@ -0,0 +1,167 @@
...
+            return
+        }
+    }
+    console.log("Function with selector " + payload.substr(0, 8) + " not found!")
+    console.log("Be careful with this transaction!")
+    process.exit(1)
+}
+
+run(process.argv[2], process.argv[3])
+    .catch(console.log)
```
cgewecke: @edisinovcic Just leaving some notes about what "verify.js" V2 could be like since we had further discussion about this in the March 15 & March 23 meetings. 

Think there were two applications:
+ a transaction validator with optional recompilation. This would be a mostly useful as an API method for wallets. It translates raw tx data into human readable output via ABI and userdocs. @ligi has done some work on this already and that could be developed into a lib method published here (via `dist/` 

+ a cli tool that recompiles (and saves?) sources using metadata as an input. It's purpose is to provide on non-centralized (local) version of the deployed service (without running a full monitor).  

Most methods in the in old verify.js file have been extracted to `src/utils.ts`. 

+ reformatMetadata
+ recompile
+ getBytecode
+ cborDecode

Old `verify.js` is missing support for ipfs.  There's relevant code in the Monitor at 
+ [retrieveCode][2]
+ [retrieveMetadataByStorageProvider][3]

[2]: https://github.com/ethereum/source-verify/blob/ede2477cf14c090f862011e2697fc668c448fe0b/src/monitor.ts#L245-L293
[3]: https://github.com/ethereum/source-verify/blob/ede2477cf14c090f862011e2697fc668c448fe0b/src/monitor.ts#L333-L379
Cli interface might be like?
```
metadata-verifier --metadata <path> [--outputDir <path>]
metadata-verifier --address <address> --chain <chain>  [--outputDir <path>]
metadata-verifier --ipfsHash <ipfs-hash> [--ipfsProvider <url>]
```
 
We could use [yargs][1] to help with writing this...

[1]: https://www.npmjs.com/package/yargs

scripts/publish_npm.sh
```diff
@@ -0,0 +1,24 @@
...
+NPM_VERSION=$(npm view verify.js dist-tags.latest)
+
+if [ $NPM_VERSION = $PACKAGE_VERSION ]; then
+    echo "Latest npm version is equal to current package version. Up the version to publish to npm."
+    exit 0
+fi
+
+# publish
+npm set //registry.npmjs.org/:_authToken=${NPM_TOKEN}
+npm publish
```
cgewecke: This is for publishing in CI?

edisinovcic: Yes.

PR #89, opened 2020-03-16
Updating docker-compose environments

edisinovcic:
Fixes incorrect url on UI to repository.


docker-compose-testing.yaml
```diff
@@ -58,7 +58,7 @@ services:
   repository:
     image: ethereum/source-verify:repository-${TAG}
     env_file:
-    - .env.testing
+    - .env
```
cgewecke: @edisinovcic Does this mean none of the testing specific env variables are exposed?

edisinovcic: Basically I've reverted to older variant because docker-compose doesn't take currently sourced .env files (source .env.testing for example) wasn't working so I had to revert it back. This also means that everytime you use an environment you have to copy all the stuff from .env.specific to .env as only .env loads into docker-compose unfortunately. This is also in their documentation but I thought I found a bypass but it seems not :(

cgewecke: Ah ok - so as a practical matter - does a local development instance still work?

edisinovcic: Also on build .env has to be copied in ui folder so react also loads it (that's why here was undefined on UI).

cgewecke: Do you think the simplest thing now for local development is just to launch without docker? That's totally ok as far as I'm concerned.

edisinovcic: What could be faster is running service you are working on without docker and everything else in docker as build time is still not a small factor. So for the things that you are working on, for example UI definitelly run it without docker and everything else with docker-compose. That's usually setup that I use in development.

edisinovcic: Well it all depends how much do you change the environment, but basically you have to setup .env once and use docker-compose as currently is. That's basically it.

edisinovcic: And the development instance is still working. This just loads all environments from .env not from .env.custom.

PR #91, opened 2020-03-18
Adding s3 and ipfs backups of repository

edisinovcic:
This will resolve missing files from repositories.
Resolves:
* https://github.com/ethereum/source-verify/issues/87
* https://github.com/ethereum/source-verify/issues/92
* https://github.com/ethereum/source-verify/issues/108

This way we have 2 ways of backuping files:
* s3 public bucket (centralized) - done
* ipfs (decentralized) - TODO

S3 bucket is will be used in append only mode. No files will be able to be deleted. 

In case that ipfs doesn't work s3 bucket can be always accessed. 

Secrets will be managed through .env which won't be pushed to repo.


src/ipfs/crontab
```diff
@@ -0,0 +1,2 @@
+1 * * * * /app/login_sync_s3.sh >> /var/log/cron.log 2>&1
```
edisinovcic: @edisinovcic TODO: remove crontabs

edisinovcic:
This will be fixed here: https://github.com/ethereum/source-verify/pull/89

edisinovcic:
Due to bugs on cirleci services have now be seperated and this can be merged to staging to fix CI issues. 

On another PR ipfs publish and deploy with secrets will be covered.

package.json
```diff
@@ -16,17 +16,16 @@
     "tslint": "tslint -p tsconfig.json",
     "coverage": "node ./node_modules/nyc/bin/nyc.js --reporter=lcov --reporter=text-summary ./node_modules/tape/bin/tape ./test/index.js",
     "coveralls": "npm run coverage && node ./node_modules/coveralls/bin/coveralls.js <coverage/lcov.info",
-    "localchain": "./scripts/localchain.js",
-    "localchain:test": "NODE_ENV=testing LOCALCHAIN_PORT=8545 ./scripts/localchain.js",
-    "dev": "npm run build:ui && ./scripts/run.sh 80 repository db",
+    "localchain": "./src/localchain.js",
+    "localchain:test": "NODE_ENV=testing LOCALCHAIN_PORT=8545 ./src/localchain.js",
```
cgewecke: Are you moving these to `src` because they're JS or...?

package.json
```diff
@@ -16,17 +16,16 @@
...
+    "dev": "npm run build:ui && ./src/run.sh 80 repository db",
     "dev:ui": "cd ./ui/ && npm install && npm start",
     "dev:server": "npm run build:ui && ./scripts/run_server.sh",
     "dev:server:local": "npm run build:ui && TESTING=true ./scripts/run_server.sh",
     "build": "tsc",
     "build:ui": "cd ./ui/ && npm install && npm run build",
-    "monitor": "tsc && ./scripts/run_monitor.sh",
-    "server": "tsc && node ./dist/server.js ./repository",
-    "server:test": "tsc && TESTING=true LOCALCHAIN_URL=http://localhost:8545 SERVER_PORT=2000 node ./dist/server.js ./repository"
```
cgewecke: Could we leaving this in...I use this server:test command.

src/run_monitor.js
```diff
@@ -1,14 +1,13 @@
 #!/usr/bin/env node
-
 const death = require('death');
-const Monitor = require('../dist/monitor.js').default;
+const Monitor = require('../dist/src/monitor.js').default;
```
cgewecke: Why is this path changing?

ui/App.js
```diff
@@ -11,7 +11,7 @@ export default function App() {
         { value: 'goerli', label: 'Görli' }
     ]
 
-    if (process.env.NODE_ENV === 'testing'){
+    if (process.env.TESTING === 'true'){
```
cgewecke: WDYT about using `||` here instead of getting rid of NODE_ENV altogether? When I'm testing locally the current setup works and I think the `TESTING` is not visible or will need additional configuration...

package.json
edisinovcic: Basically yes, but would probably be more correct to leave them in scripts. Will move that back there.

src/run_monitor.js
edisinovcic: Tsc creates a path in dist like this, so dist/src/filename.js

Maybe some aditional configuration is needed here so that it "flattens" the path?

cgewecke: Oh if it works, great. I'm just confused about how it resolved before without this...

edisinovcic: I'm not sure to be honest, it just stopped working... :/

PR #96, opened 2020-03-24
Validate server inputs / add server unit tests

### LINKED ISSUE #6: Getting Error: Metadata file not found. Did you include "metadata.json"
I upload just one file called metadata.json (is the name important though ?)

The same one mentioned in https://github.com/ethereum/source-verify/issues/5

And I get the following error :

```
Error: Metadata file not found. Did you include "metadata.json"
```

cgewecke:
**PR targets #84**
 
 + Fixes the bug discovered in #6, where submitting a single file always errored with 'Metadata not found'
 + Adds validation checks for `address` and `chain` parameters in injector
 + Adds status codes to server error responses
 + Adds unit tests for `server.ts`

### DIFF
--- a/src/injector.ts
+++ b/src/injector.ts
@@ -15,10 +15,16 @@ import {
   getBytecode,
   recompile,
   RecompilationResult,
+  getBytecodeWithoutMetadata as trimMetadata
 } from './utils';
 
 declare interface StringMap {
   [key: string]: string;
+}
+
+declare interface BytecodeMatch {
+  address: string | null,
+  status: 'perfect' | 'partial' | null
 }
 
 export interface InjectorConfig {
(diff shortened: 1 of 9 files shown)


src/injector.ts
```diff
@@ -271,6 +299,9 @@ export default class Injector {
     files: string[]
   ) : Promise<void> {
 
+    this.validateAddresses(addresses);
```
chriseth: What if the address was left out on mainnet?

cgewecke: We don't have an exception for mainnet in the current version. What made that possible in the first PoC?

cgewecke: Ah sorry - that's what the addressDB is about. Going to open a separate issue about wiring that in.

src/server.ts
```diff
@@ -4,25 +4,37 @@ import fileUpload from 'express-fileupload';
...
+  }]
+});
 
 const repository = './repository/';
 const port = process.env.SERVER_PORT;
 
 
 app.use(express.static('ui/dist'))
+
+// TODO: 52MB is the max file size - is this right?
```
edisinovcic: Open this as issue?

cgewecke: Opened #99

Issue #95, opened 2020-03-23
Add repository fallback for contracts with imperfect hash match


cgewecke:
e.g for cases where re-compilation results in a bytecode which matches everything *except* the metadata hash. 

At the March 23 meeting we discussed how this should be added to the repository and suggestion was it should be its own compartment, possibly something like.

```

contract/
  goerli/
    <address>/

partial_match/
  contract/
    goerli/
      <address>/
```

Partial matches might be mutable.

cgewecke:
#98

Issue #76, opened 2020-03-09
Organize documentation


cgewecke:
At some point it could make sense to have a simple ReadTheDocs explaining
+ What metadata is
+ How the services / endpoints work
+ What tools are available to help upload sources & metadata to the repository
+ How to access and consume data in the repository
  + perfect matches vs. partial_matches
+ How to deploy your own version of the service

edisinovcic:
Should we publish this as a part of github pages? From readme on master?

cgewecke:
@edisinovcic Not sure, when I've used RTD before I've just had a link in the README to their hosted subdomain. I think Solidity is like this as well...

https://solidity.readthedocs.io/en/v0.6.0/metadata.html

What are you imagining re: Github pages?

edisinovcic:
RTD looks better generally so yes it's a better option. 
Example of github pages: https://blockchain-it-hr.github.io/ink-remix-plugin/

edisinovcic:
Addressing: https://github.com/ethereum/sourcify/issues/104

edisinovcic:
https://github.com/ethereum/sourcify/issues/154

kuzdogan:
Closing, see https://docs.sourcify.dev

PR #98, opened 2020-03-25
Save contracts when metadata does not match

cgewecke:
**PR targets #96**

#95 

Adds logic and tests for tolerant bytecode comparison when injecting. 

If the bytecodes match up to but not including the metadata component, the sources and metadata are saved to a fallback folder called `partial_matches` as below:

**Example**
```
repository/
  partial_matches/
    goerli/
      0x26EA1B6a99F491B5700D104CaD55c6Db07AAAFA2/
        sources/
        metadata.json
```

### DIFF
--- a/src/injector.ts
+++ b/src/injector.ts
@@ -15,10 +15,16 @@ import {
   getBytecode,
   recompile,
   RecompilationResult,
+  getBytecodeWithoutMetadata as trimMetadata
 } from './utils';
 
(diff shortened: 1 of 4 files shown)


cgewecke:
> Maybe just add to the README an explanation about the folder structure.

I've added this topic to #76, I'm going to start work on that today or tomorrow.

Issue #37, opened 2020-01-20
Add test sources for libraries and Swarm hashes


cgewecke:
In the Jan 20 meeting, @chriseth noted that there might be a problem correctly matching the metadata hash for deployed libraries.

We should have tests for this case.

chriseth:
Small correction: The problem is matching the deployed bytecode, not the metadata hash.

PR #101, opened 2020-03-26
Restore injector support for bzzr0

cgewecke:
Part of #37 and should also make  #6 close-able.

I accidentally removed bzzr0 hash support from the injector during the initial refactor. 

PR restores it and adds tests for injecting bytecodes with swarm hashes.


cgewecke:
@edisinovcic Thanks!

PR #105, opened 2020-03-27
Add support for contracts with library links

cgewecke:
Completes #37

Contracts with linked libraries look like they need a little extra help. Linking is often done manually during deployment which means that the metadata in many compilation artifacts will not contain the relevant addresses.

The metadata library object also missing an outer source file path key that solc requires. 

PR adds a convenience method to the `src/utils` which takes
+ a mapping of library names to addresses 
+ an `evm.bytecode.linkReferences` object 
+ a metadata string

...and synthesizes a new metadata string with a `libraries` object solc can consume. 

NB: the metadata hash generated during recompilation is different than the one retrieved from deployed bytecode - these contracts are partial matches. 

There might be something I'm missing here... @chriseth do you know of a better way to do this?
 
[EDIT - per discussion in March 30 mtg this needs to save link references (see #112)


haltman-at:
Question, does this also account for the delegatecall guard?  Since obviously that's going to be necessary for the libraries themselves (assuming they're compiled with Solidity 0.4.20 or later).

cgewecke:
@haltman-at 

No, this is just about reconciling metadata that compiles with library link placeholders with the bytecode on chain (which has Library addresses embedded in it). 

(37a0fd0 are d6d493f are the only relevant commits - think this PR should be closed in favor of one based on a more recent state).

haltman-at:
Hm.  Seems a little useless to be able to verify contracts linked to libraries, if the libraries themselves can't be verified...

cgewecke:
@haltman-at 

This PR just addresses one component of the problem. To really manage Libraries correctly we need to get contract creation codes. That's being tracked via #111.

Issue #90, opened 2020-03-17
ENS support in address field


ligi:
Great suggestion on twitter: https://twitter.com/gh1dra/status/1239962423341322241

corydickson:
Give me a few days and I'll pick this up

corydickson:
Strange times ahead :)

corydickson:
As a user I would like to be able to enter an ENS address and it will resolve the associated on-chain ethereum address with its source code and metadata. If such an entry already exists in the repository notify the user, otherwise prompt the user to upload said files to IPFS and verify.

To achieve this we will need to interact with the ENS resolver contract on a given chain. This mapping could be maintained in the front-end. Once we have the address, check it's runtime bytecode against entries in the `address-db`, if the value at a given key match then it already exists inside the repository. 

A consideration is that this will only work for TLD if my understanding is correct? Furthermore is it necessary to have the user upload these files, or would a better approach is to leverage this [EIP](https://github.com/ethereum/EIPs/pull/205), ensuring that the retrieved ABI hash is already pinned to IPFS. I believe the source code would still need to be provided though...

I'll get started mocking up the UI, hopefully I'm thinking about this the right way

ligi:
ah good that we talked about it ;-)
I understood your idea differently. With ENS support I thought you mean instead of entering an hex address in the form you would enter the ENS address of the contract.
Actually I was not even aware of EIP-205. But before going this route I would actually check if there are contracts out there that leverage this EIP - have the gut feeling that it is not.
I think we should split these issues a bit and keep this issue around "ENS support in address field" - meaning if the user enters something that does not start with 0x we will try to resolve it via ENS instead of telling the user that it is not an valid address.
But feel free to open issues with your follow up ideas. But I think we should start with the low hanging fruit of what I just outlined. What do you think?

corydickson:
I think we are speaking of similar things for the ENS resolution, and agree that's the best place to start. Do you think that verifying the on-chain bytecode against what is already inside the database to be a fruitful check? The reasoning behind this was so that we could prevent duplicate contracts being added to the database i.e two of the same runtime bytecodes would map to a 0x address and ENS name and would be treated as separate entities

ligi:
AFAIK duplication cannot happen for exact matches which we only support currently. It might become an issue when partial matches are supported

PR #114, opened 2020-03-31
Resolve ENS addresses upon submission

corydickson:
Addresses: #90

The approach laid out here uses the existing submission endpoint to resolve an ENS address. I went this way to avoid bringing in another web3 instance into the project, despite some of the comments in the issue indicating the ability to provide some tooltip or notification of bad user input

Let me know if that would be preferred and we can also sanitize for regular bytecode addresses


src/injector.ts
```diff
@@ -188,18 +189,23 @@ export default class Injector {
...
   public async inject(
     repository: string,
     chain: string,
     address: string,
+    isENS: boolean,
     files: string[]
   ) : Promise<string[]> {
 
     if (address) {
+      if (isENS) {
```
corydickson: Another way this could be achieved is keeping the assumption the input is not sanitized and match domains here, not sure if we should be accepting subdomains...

corydickson:
After doing some digging not sure if this will work for all of the supported chainIds, is it acceptable to limit it to mainnet, rinkeby, ropsten

corydickson:
We can add a corresponding prompt for the user as well

edisinovcic:
@corydickson that's definitely a good start. Not sure if the prompt is needed as we already get the network (chain) so we can just check in isEns is one of the listed networks.

PR #115, opened 2020-04-02
Merging staging to production

edisinovcic:
Due to the changes in how docker compose is now partitioned as one service one yaml file I will have to update it a bit now manually after this is merged.


edisinovcic:
Tested manually the flow, everything works. Ready to be merger.

src/injector.ts
```diff
@@ -190,58 +352,86 @@ export default class Injector {
    * @param  {string}            chain      chain name (ex: 'ropsten')
    * @param  {string}            address    contract address
    * @param  {string[]}          files
-   * @return {Promise<string[]>}            addresses of successfully verified contracts
+   * @return {Promise<void>}
```
cgewecke, line 355: Forgot to update docs comment here...
```suggestion
   * @return {Promise<string[]>}            addresses of successfully verified contracts
```

environments/.env.stable
```diff
@@ -0,0 +1,43 @@
...
+BUCKET_NAME_STAGING=s3://staging-source-verify
+
+# Geth config
+NETWORK_NAME=goerli
+
+# IPFS config
+IPFS_SECRET=xxx
+
+# Other config
+TESTING=true
```
cgewecke, line 41: Maybe...
```suggestion
TESTING=false
```

environments/.env.latest
```diff
@@ -1,22 +1,43 @@
...
 
+# Geth config
+NETWORK_NAME=goerli
+
+# IPFS config
+IPFS_SECRET=xxx
 
 # Other config
-#can be stable for production or latest for development
 TESTING=true
```
cgewecke, line 41: Setting TESTING to true make the monitor listen to localchain **only**. 
```suggestion
TESTING=false
```

PR #120, opened 2020-04-06
Updating injector to contain status in reponse

edisinovcic:
I've updated the response to an array of addresses, status pairs. It's nested inside of result so it doesn't break the frontend. It can be updated additionally if needed.

{"result":[{"address":"0xf510C8C3f0943AD723eb1EC3D5086e01ea33f6a5","status":"perfect"}]}

### DIFF
--- a/src/injector.ts
+++ b/src/injector.ts
@@ -33,6 +33,11 @@ export interface InjectorConfig {
   silent? : boolean
 }
 
+export type Match = {
+  address: string,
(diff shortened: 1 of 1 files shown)


edisinovcic:
Should we merge this then and then later see how to refactor UI?

cgewecke:
@edisinovcic Yes, that's fine with me.

PR #122, opened 2020-04-08
Adding address check

edisinovcic:
This is a simple check that tries to find files in local repository by network name and address.

Do we need more detailed check here?

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -5,6 +5,7 @@ import cors from 'cors';
 import Injector from './injector';
 import Logger from 'bunyan';
 import util from 'util';
+import { findByAddress } from './utils';
 
 const app = express();
 
@@ -31,7 +32,6 @@ const log = Logger.createLogger({
 const repository = './repository/';
 const port = process.env.SERVER_PORT;
 
-
 app.use(express.static('ui/dist'))
 
 // TODO: 52MB is the max file size - is this right?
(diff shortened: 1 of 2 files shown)


src/server.ts
```diff
@@ -53,11 +53,14 @@ app.post('/', (req, res) => {
...
 
   // If address exists in repostory just return it, otherwise recompile
-  const result = findByAddress(req.body.address, req.body.chain, repository);
-  if(result){
-    res.status(200).send({ address: req.body.address, match: "match" }) //TODO:
+  try {
+    const result = findByAddress(req.body.address, req.body.chain, repository);
+    res.status(200).send({ result }) 
+  } catch(err) {
+    console.info("Could not find file in repository, proceeding to recompilation");
```
cgewecke, line 60: Do you think this info is not worth formatting as a JSON-ified server log?

cgewecke:
> Do we need more detailed check here?

I think only getting perfect matches as you've done here seems like the safest thing initially.  What additional detail are you thinking of?

We could add a unit test for this at some point - there are fixtures set up to test the server [here][1]. . You'd submit some files that save and *then* try a request without files and see if it works.

[1]: https://github.com/ethereum/source-verify/blob/master/test/server.js

src/server.ts
edisinovcic: Definitely, will do that now.

edisinovcic:
> > Do we need more detailed check here?
> 
> I think only getting perfect matches as you've done here seems like the safest thing initially. What additional detail are you thinking of?
> 
> We could add a unit test for this at some point - there are fixtures set up to test the server [here](https://github.com/ethereum/source-verify/blob/master/test/server.js). . You'd submit some files that save and _then_ try a request without files and see if it works.

Yeah, basically idea is to just send an address and chain and then you can also get response that it is valid if it exists in repository, if you also send files then recompilation will be done if there are no files in the repository. Only thing to consider here is if someone wants to upload already recompiled files (we have that address that was send in repository). So maybe we should separate this somehow? I'm mostly thinking about upgradable contracts. On the other hand recompilation shouldn't be the same here so possibly we don't have it as a problem?

edisinovcic:
I've merged it here so we can do a update on another PR and add a test for this if needed.

PR #123, opened 2020-04-09
Server api

edisinovcic:
Adding better error handling.
UI no longer requires both address and files to be added, only address is required.

Flow currently is:
Test by address if files exist in the repository and are a perfect match -> return true
If address check fails go towards recompilation.

There are now more status codes thrown if something happens during address check and recompilation.


edisinovcic:
Tnx a lot!

PR #124, opened 2020-04-10
Adding basic postgres configuration

edisinovcic:
Solves: https://github.com/ethereum/source-verify/issues/111
Adding basic configuration for running postgres in docker.


edisinovcic:
@cgewecke @chriseth I'm not sure if this is to much indexes or are we missing some?
Also please check that types are correct.

cgewecke:
@edisinovcic Are addresses part of this DB?

edisinovcic:
> @edisinovcic Are addresses part of this DB?

I forgot the most important part :)

Will add it now.

edisinovcic:
Try blob instead of text.

edisinovcic:
INDEX on chain_id and address.

edisinovcic:
INDEX address

edisinovcic:
INDEX on chain_id, deployed_code_hash, creation_bytecode
INDEX on address, deployed_code_hash, creation_bytecode

src/postgres/schema.sql
```diff
@@ -0,0 +1,15 @@
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
+CREATE TABLE bytecode (
+    id uuid PRIMARY KEY,
+    address text NOT NULL,
+    creation_bytecode text NOT NULL,
```
chriseth, line 6: Can we make these binary? I think text is quite a waste of space.
It has to allow for zero-bytes, though.

src/postgres/schema.sql
```diff
@@ -0,0 +1,15 @@
...
+    chain_id integer NOT NULL,
+    block_number integer NOT NULL,
```
chriseth, line 8: Ah if we add the block number (which is useful information) we have to check to select the largest one for querying the current deployed bytecode hash.

chriseth: Ah sorry, there is another primary key anyway.

src/postgres/schema.sql
edisinovcic: Probably bitea format would be best. 

https://www.postgresql.org/docs/9.0/datatype-binary.html

src/postgres/schema.sql
edisinovcic: We talked about storing all the changes from oldest to latest, but to show users only the latest change. In the future if we want to track changes in time we could easily support this and you could track changed for a contract in time.

edisinovcic:
Adding db seeder: https://github.com/Shard-Labs/source-verify-seeder

edisinovcic:
SQL lite will be used.

PR #125, opened 2020-04-10
Add unit tests for address-only server submissions (plugin)

cgewecke:
Adds some server tests following on #123. 

There are a few logic changes because we weren't exiting the flow correctly in some cases. 

For example, when an address was discovered successfully, we didn't return immediately after response, causing a second response and express to complain about resetting the headers.

When an address was submitted without files, the error was about missing req.files.files (but the files are optional now.) 

A little logic has been re-worked for the error cases as well by wrapping in try/catch blocks etc.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -7,7 +7,7 @@ import Logger from 'bunyan';
 import {
   findInputFiles,
   InputData,
-  sanatizeInputFiles,
+  sanitizeInputFiles,
   findByAddress,
   errorMiddleware,
   Match
@@ -65,17 +65,25 @@ app.post('/', (req, res, next) => {
     chain: req.body.chain,
   }
 
-  // Try to find by address
+  // Try to find by address, return on success.
   try {
     const result = findByAddress(req.body.address, req.body.chain, repository);
-    res.status(200).send({ result })
+    res.status(200).send(result);
+    return;
   } catch(err) {
     const msg = "Could not find file in repository, proceeding to recompilation"
     log.info({loc:'[POST:VERIFICATION_BY_ADDRESS_FAILED]'}, msg);
   }
 
-  inputData.files = sanatizeInputFiles(findInputFiles(req, log), log);
+  // Try to organize files for submission, exit on error.
+  try {
+    const files = findInputFiles(req, log);
+    inputData.files = sanitizeInputFiles(files, log);
+  } catch (err) {
+    return next(err);
+  }
 
+  // Injection
   const promises: Promise<Match[]>[] = [];
   promises.push(injector.inject(inputData));
 
--- a/src/server.ts
+++ b/src/server.ts
@@ -36,7 +36,7 @@ const log = Logger.createLogger({
   }]
 });
 
-const repository = './repository';
+const repository = process.env.MOCK_REPOSITORY || './repository';
 const port = process.env.SERVER_PORT;
 
 app.use(express.static('ui/dist'))
--- a/src/server.ts
+++ b/src/server.ts
@@ -68,7 +68,7 @@ app.post('/', (req, res, next) => {
   // Try to find by address, return on success.
   try {
     const result = findByAddress(req.body.address, req.body.chain, repository);
-    res.status(200).send(result);
+    res.status(200).send({result});
     return;
   } catch(err) {
     const msg = "Could not find file in repository, proceeding to recompilation"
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -237,26 +240,20 @@ export function sanatizeInputFiles(inputs: any, log: Logger): string[] {
...
+    fs.readdirSync(normalizedPath);
+  } catch(e){
+    throw new Error("Address not found in repository");
   }
 
-  throw new Error("Address not found in repository");
+  return {
+    address: address,
+    status: "perfect"
+  }
```
cgewecke, line 256: I've changed this to only return a single object, because there aren't multiple addresses to be found.

src/server.ts
```diff
@@ -65,17 +65,25 @@ app.post('/', (req, res, next) => {
     chain: req.body.chain,
   }
 
-  // Try to find by address
+  // Try to find by address, return on success.
   try {
     const result = findByAddress(req.body.address, req.body.chain, repository);
-    res.status(200).send({ result })
+    res.status(200).send(result);
```
cgewecke, line 71: I think this change might be wrong....but what is supposed to be sent here?

In the current version it looks like this?  
```
send( { [ { status: 'perfect', address: '0x....' } ] } )
```

src/utils.ts
cgewecke: Reverted in 72ecb77

src/server.ts
cgewecke: Reverted in 72ecb77

src/server.ts
```diff
@@ -65,17 +65,25 @@ app.post('/', (req, res, next) => {
     chain: req.body.chain,
   }
 
-  // Try to find by address
+  // Try to find by address, return on success.
   try {
     const result = findByAddress(req.body.address, req.body.chain, repository);
-    res.status(200).send({ result })
+    res.status(200).send({result});
+    return;
```
edisinovcic, line 72: Good catch!

src/utils.ts
```diff
@@ -198,16 +199,18 @@ export function findInputFiles(req: Request, log: Logger): any {
...
   }
 
-  const msg = 'Request missing expected property: "req.files.files"';
-  log.info({loc: '[POST:REQUEST_MISFORMAT]', err: msg})
-  throw new BadRequest(msg);
+  // If we reach this point, an address has been submitted and searched for
+  // but there are no files associated with the request.
+  const msg = 'Address for specified chain not found in repository';
+  log.info({loc: '[POST:ADDRESS_NOT_FOUND]', err: msg})
+  throw new NotFound(msg);
```
edisinovcic, line 206: It's definitely better to throw not found instead of bad request.

test/server.js
```diff
@@ -188,4 +195,72 @@ describe("server", function() {
...
+        .field("address", simpleInstance.options.address)
+        .field("chain", 'localhost')
+        .end(function (err, res) {
+          assert.equal(res.status, 200);
+          done();
+        });
+    });
+
+    afterEach(function(){
+      try { exec(`rm -rf ${mockRepo}`) } catch(err) { /*ignore*/ }
```
edisinovcic, line 216: Maybe just deleting created files?

Currently, when I have something in local repository folder everything gets deleted after running test.

edisinovcic:
LGTM! 

There are a couple of comments, but generally good to be merged.

src/utils.ts
```diff
@@ -238,25 +241,19 @@ export function sanatizeInputFiles(inputs: any, log: Logger): string[] {
...
-  });
-
-  if(files.length > 0){
-    matches.push({
-      address: address,
-      status: "perfect"
-    });
-    return matches
+  try {
+    fs.readdirSync(normalizedPath);
```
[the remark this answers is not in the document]
cgewecke: > should we at least check extensions of the files?

Yes was wondering about this...do you think its enough if we check that there is a `metadata.json`? e.g
```
${repository}/contract/${chain}/${address}/metadata.json
```

test/server.js
cgewecke: This should be fixed by a976d5b. 

Opted for configuring the repository name with an environment variable.

src/utils.ts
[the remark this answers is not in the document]
edisinovcic: Yeah, it seems good enough for now. If needed we can update it further in the future.

test/server.js
edisinovcic: Nice

src/utils.ts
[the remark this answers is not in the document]
cgewecke: f6771b1

PR #126, opened 2020-04-11
Add Monitor E2E

cgewecke:
Adds an E2E test for the monitor vs. the goerli network which runs after the deploy job on staging and master. 

Test uses Truffle's example Metacoin project
+ clones [cgewecke/metacoin-source-verify][1]
+ uploads sources and metadata to IPFS
+ deploys contracts to goerli
+ waits 45s
+ queries staging repository for metadata.json at the deployed address.

An example of the test completing successfully (from an early commit) can be seen here:

https://circleci.com/gh/ethereum/source-verify/1172

[1]: https://github.com/cgewecke/metacoin-source-verify


scripts/monitor_ci.js
```diff
@@ -0,0 +1,56 @@
...
+ * monitor.
+ */
+
+const assert = require('assert');
+const fetch = require('node-fetch');
+const util = require('util');
+const log = console.log;
+
+// TODO: Check if multiple repositories and toggle by staging or master branch
+const root = 'https://contractrepostaging.komputing.org/';
```
edisinovcic, line 16: There is a FQDN in config and we could also add repository endpoint.

For example: https://github.com/ethereum/source-verify/blob/be2a631b8200df28ec96d778d3e2f63529d5a596/environments/.env.latest#L2

cgewecke: @edisinovcic Ah cool - I saw `REPOSITORY_URL` in the envs and it's "staging" for both latest and stable. Is it safe to make those different? If so I can add that here.

https://github.com/ethereum/source-verify/blob/cccf260c229cee2b4850ece83734e3be7b5197b1/environments/.env.stable#L25

edisinovcic: Definitely, latest is for staging and stable is for master so please update it.

scripts/monitor_ci.js
```diff
@@ -12,8 +12,10 @@ const fetch = require('node-fetch');
 const util = require('util');
 const log = console.log;
 
-// TODO: Check if multiple repositories and toggle by staging or master branch
-const root = 'https://contractrepostaging.komputing.org/';
+const root = (process.env.CIRCLE_BRANCH === 'staging')
+  ? 'https://contractrepostaging.komputing.org/'
```
edisinovcic, line 16: Shouldn't be easier here to just use: process.env.REPOSITORY_URL instead of the logic?

cgewecke: @edisinovcic Is it visible here though? We're not in a docker container when running this script, only in Circle.

scripts/monitor_ci.js
```diff
@@ -0,0 +1,58 @@
...
+
+  console.log(util.inspect(metadata));
+};
+
+main()
+  .then(() => process.exit(0))
+  .catch((err) => {
+    console.log(err);
+    process.exit(1);
+  })
```
cgewecke, line 58:
```suggestion
  })

```

scripts/monitor_ci.sh
```diff
@@ -0,0 +1,18 @@
...
+git clone https://github.com/cgewecke/metacoin-source-verify.git
+cd metacoin-source-verify
+npm install
+
+# Publishes sources to IPFS (via Infura) and deploys contracts to Goerli
+# Account key and Infura project ID are Circle CI env variable settings.
+npm run deploy:goerli
+
+# Give monitor a chance to detect and save.
+sleep 45
```
cgewecke, line 14:
```suggestion
sleep 145
```

I think this will need more time when recompilation for the monitor is hooked up...

scripts/monitor_ci.js
cgewecke: @edisinovcic  I'm going to merge this just to see if works on staging (because it probably won't ha ha!!). Will revisit this question...

cgewecke: It looks like it worked actually...

https://circleci.com/gh/ethereum/source-verify/1201

Deployments to Goerli via Infura/Truffle are a little flaky though - have seen it fail a couple times with a `contract couldn't be stored` error. The job has to be re-run. 

I think this is because of a [long-standing issue][1] with some clients (Parity?) returning a transaction receipt before they've successfully stored the bytecode, causing a race-condition crash in Web3 1.2.1.

Will look at using something besides truffle migrations to deploy if this becomes too unstable/annoying.

[1]: https://github.com/trufflesuite/truffle/issues/2257

edisinovcic: Sorry, just now received a notification. Environment could be loaded, but if this works this way then it's good. :) Maybe adding todo to check this later could be useful.

edisinovcic: I have to get up and running full nodes for goerli and rinkeby on server and we can then use them for this instead of infura?

There is still some config there that I would like to add. Also need to do stop geth for mainnet to resize the server, but I'm not sure if that will corrupt the database... Anyways we could have this fixed in relatively short time. We could just create an issue for it so we know to address it in the future.

cgewecke: > I have to get up and running full nodes for goerli and rinkeby on server and we can then use them for this instead of infura?

Oh yeah that would be awesome! I'll make an issue for this and for the env thing.

PR #135, opened 2020-04-17
Adding file retrieving on server

edisinovcic:
This serves as "raw repository". It returns either URLs to files in folder and subfolders or actual files. 

Files are not send via sendFile because there are more then one and express doesn't support sending multiple files in single response, maybe having some better of doing this would be good?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1863,6 +1863,11 @@
         "randombytes": "^2.0.0"
       }
     },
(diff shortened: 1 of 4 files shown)


edisinovcic:
@cgewecke merging now so I can test with the remix plugin on staging, but please take a look after merge also so we can fix if something is done incorrectly. Tnx :)

PR #139, opened 2020-04-21
Updating injector to new configuration with chains.json

edisinovcic:
Resolves: 
https://github.com/ethereum/source-verify/issues/145
https://github.com/ethereum/source-verify/issues/143


edisinovcic:
The current branch has been deployed on: 
https://verify.shardlabs.io/
https://contractrepo.verify.shardlabs.io/
https://server.verify.shardlabs.io/

edisinovcic:
@cgewecke this changes the whole flow from named based to id based storage and routing.

So either fetching file by name (for example: mainnet) or by id (1) you will get the same result.

Also, injector also writes files by id without byChainId path.

cgewecke:
@edisinovcic 

> Should I also update monitor to recompile in this PR?

I think we should do this separately after this goes in. It needs to manage library links correctly and there's some complexity there. 

Not certain why the server tests are failing here but think the default chainId for ganache-core might be `1` whereas the `.json` has it as 0. Do you know if this is correct?

Web3 has [a method][1] to query the client for this:
```javascript
const chainId = await web3.eth.getChainId()
```

[1]: https://web3js.readthedocs.io/en/v1.2.7/web3-eth.html#getchainid

edisinovcic:
> @edisinovcic
> 
> > Should I also update monitor to recompile in this PR?
> 
> I think we should do this separately after this goes in. It needs to manage library links correctly and there's some complexity there.
> 
> Not certain why the server tests are failing here but think the default chainId for the ganache client might be `1` whereas the `.json` has it as `0`. Do you know if this is correct?
> 
> Web3 has [a method](https://web3js.readthedocs.io/en/v1.2.7/web3-eth.html#getchainid) to query it:
> 
> ```js
> const chainId = await web3.eth.getChainId()
> ```

Nice, tnx.

PR #180, opened 2020-05-28
Ipfs 0.5

edisinovcic:
- Updates ipfs to version 0.6

- Adds logic for ens updater

TODO:
Manage an increase of a gas price dynamically.

### DIFF
--- a/services/ipfs/Dockerfile.ipfs
+++ b/services/ipfs/Dockerfile.ipfs
@@ -1,10 +1,14 @@
-FROM openjdk:8
-RUN apt-get update && apt-get install -y cron curl unzip git 
+FROM openjdk:11
+RUN apt-get update && apt-get install -y cron curl unzip git unzip
 WORKDIR /app
 RUN git clone https://github.com/komputing/source_verify_ens_updater.git /app
-RUN ./gradlew build --no-daemon 
-RUN curl https://dist.ipfs.io/go-ipfs/v0.4.23/go-ipfs_v0.4.23_linux-amd64.tar.gz > go-ipfs.tar.gz && tar xvfz go-ipfs.tar.gz && cd go-ipfs && ./install.sh
-ADD services/ipfs/publish_to_ipfs.sh ./publish_to_ipfs.sh
-ADD services/ipfs/entrypoint.sh ./entrypoint.sh
+RUN ./gradlew distZip
+RUN unzip ./build/distributions/source_verify_ens_updater.zip
+RUN chmod +x source_verify_ens_updater/bin/source_verify_ens_updater
+RUN ls -al ./build/distributions/
+RUN curl https://dist.ipfs.io/go-ipfs/v0.6.0/go-ipfs_v0.6.0_linux-amd64.tar.gz > go-ipfs.tar.gz && tar xvfz go-ipfs.tar.gz && cd go-ipfs && ./install.sh
+ADD src/ipfs/publish.sh ./publish.sh
+ADD src/ipfs/entrypoint.sh ./entrypoint.sh
+ADD src/ipfs/cron.job ./cron.job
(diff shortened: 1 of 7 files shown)


src/ipfs/publish_to_ipfs.sh
```diff
@@ -1,4 +1,11 @@
...
-/usr/local/bin/ipfs name publish `/usr/local/bin/ipfs add -Q -r /app/repository`
+export PATH=~$PATH:/usr/local/bin:/usr/local/openjdk-11/bin:/usr/bin
+export JAVA_HOME=/usr/local/openjdk-11/
+
+hash=$(ipfs add -Q -r /app/repository)
+echo $hash
+curl -X POST "https://ipfs.komputing.org/api/v0/pin/add?arg=$hash&recursive=true"
+ipfs name publish $hash
+
+/app/source_verify_ens_updater/bin/source_verify_ens_updater $IPFS_SECRET /app/repository
```
ligi, line 11: the ENS updater should get no IPFS secret

edisinovcic: Removed.

src/ipfs/publish_to_ipfs.sh
```diff
@@ -1,4 +1,11 @@
...
-/usr/local/bin/ipfs name publish `/usr/local/bin/ipfs add -Q -r /app/repository`
+export PATH=~$PATH:/usr/local/bin:/usr/local/openjdk-11/bin:/usr/bin
+export JAVA_HOME=/usr/local/openjdk-11/
+
+hash=$(ipfs add -Q -r /app/repository)
+echo $hash
+curl -X POST "https://ipfs.komputing.org/api/v0/pin/add?arg=$hash&recursive=true"
+ipfs name publish $hash
+
+/app/source_verify_ens_updater/bin/source_verify_ens_updater /app/repository
```
ligi, line 11: Now you removed the IPFS secret here which is correct.
But the parameter still needs to be there - just the ENS key not the IPFS key. As Far as I see this way it would just fail.
Also ENS publishing should not be in a file called publish_to_ipfs.sh - this is not really separation of concerns and we (or other people) might stumble over this later

edisinovcic: So we would have two scripts one for updating ens and one ipfs uploader? And ens will run after ipfs is finished?

ligi: yes. Another option would just to give it another (more generic) name like publisher. But I lean on just having 2 scripts. IMHO this is cleaner. Separation of concerns FTW.

edisinovcic: Agreed, will seperate in two.

edisinovcic:
@ligi what about gas prices? 

Should we have some option to pay higher gas price then 1 gwei if it's not successfully updated for longer time than x days/weeks?

ligi:
yea - we should do this. Maybe increase it by 1gwei per day since last update and next time start at 1 again? How does this sound to you?
But would love if I would not pay it out of my pocket in this case. Maybe as @franzihei registered the sourcify ENS name (and I think she is better organizing reimbursements) we can do it this way?
maybe we should also target this name then?

edisinovcic:
Yeah, definitely it should be part of the sourcify project funding.

franzihei:
@ligi can you open an issue w.r. to the funding issue (or let me know how I could help)?

ligi (changes requested):
In the meantime we are at go-ipfs 0.6.0 and should use this version

edisinovcic:
@ligi is there any substantial changes there that I should be aware of?

ligi:
it became faster AFAIR due to QUIC by default - but interface wise AFAIR nothing has changed - so you do not need to change anything else than the version number

Issue #181, opened 2020-06-01
Directory structure


ligi:
as discussed just on the call having `contracts` and `partial_matches` on the same level is suboptimal. 2 possible solutions I see:
 * 1 have `/contracts/partial_match` and `/contracts/full_match`
 * 2 have `/partial_match` and `/full_match`

I lean to 1 - but not a strong opinion on this
we need to decide on this before #159 so that from having an initial version in the manifest we have a stable directory structure

chriseth:
Note that there is also "match with deployed bytecode but not with creation bytecode" - or is that the same thing as partial_match?

edisinovcic:
@chriseth all of those should go to partial_match currently but that can also be expanded if needed. Not sure on this part. I'm okay with both 1 and 2 @ligi

edisinovcic:
1

edisinovcic:
It's merged in staging. @ligi please check if format is working for you.

Here is URL to the contractrepo to the manifest.json file.

https://contractrepostaging.komputing.org/manifest.json

PR #185, opened 2020-06-14
Adding repository tag and updating paths for full and partial matches

edisinovcic:
Solves: https://github.com/ethereum/sourcify/issues/181
Affects: https://github.com/Shard-Labs/remix-sourcify/issues/29


edisinovcic:
Add version of 1 in every tag for now.

edisinovcic:
Name should be manifest

src/chains.json
```diff
@@ -19,7 +19,7 @@
       "network": "mainnet",
       "networkId": 1,
       "nativeCurrency": {"name":"Ether","symbol":"ETH","decimals":18},
-      "web3": ["https://mainnet.infura.io/v3/891fe57328084fcca24912b662ad101f"],
```
ligi, line 22: when we just change the api keys here - what will prevent the same thing happens again?

edisinovcic: It will be moved to .env. Still the tests are not working properly so I have to fix them first. :)

edisinovcic:
{
  timestamp: 120000,
  version: 1.0.0
}

edisinovcic:
Solved here: https://github.com/ethereum/sourcify/pull/193

PR #202, opened 2020-07-01
Adding conversion from new metadata format to old format for recompilation

edisinovcic:


edisinovcic:
@cgewecke this are the changes that we spoke about on the call for new solidity metadata format. It's still WIP.

PR #223, opened 2020-07-21
Feature find by address endpoint

Jakic007:
Light endpoint that will be used for automatic checking of already verified contracts when user pastes address list inside text field.
It will try to match with any chain.


src/server.ts
```diff
@@ -123,14 +123,36 @@ app.post('/', (req, res, next) => {
...
   Promise.all(promises).then((result) => {
-    res.status(200).send({result});
+    res.status(200).send({ result });
   }).catch(err => {
     next(err); // Just forward it to error middelware
   })
 })
 
+// Only checks if it is already verified in the repository
+app.get('/lightCheck', (req: any, res) => {
```
edisinovcic, line 133: Would rename this endpoint to something like "checkByAddresses"

PR #230, opened 2020-07-24
add quick check addresses feature

igor-lid:


igor-lid:
- added address validation
- added address verification (only if address or addresses are valid)
- display notifications based on a given response

PR #238, opened 2020-08-03
Repository cors enable

Jakic007:


Jakic007:
This is blocked until we solve metadata.json naming.

PR #239, opened 2020-08-07
server can use dappnode instead of infura

Jakic007:
Added dappnode endpoints to chains.json that are used if infuraID is not provided.
Docker image for server is now node-alpine which reduced it size for 1/3.

### DIFF
--- a/src/chains.json
+++ b/src/chains.json
@@ -20,6 +20,7 @@
       "networkId": 1,
       "nativeCurrency": {"name":"Ether","symbol":"ETH","decimals":18},
       "web3": ["https://mainnet.infura.io/v3/${INFURA_ID}"],
+      "fullnode": {"dappnode": "http://geth.dappnode:8545"},
       "faucets": [],
       "infoURL": "https://ethereum.org"
     },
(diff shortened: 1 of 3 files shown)


src/chains.json
```diff
@@ -20,6 +20,7 @@
       "networkId": 1,
       "nativeCurrency": {"name":"Ether","symbol":"ETH","decimals":18},
       "web3": ["https://mainnet.infura.io/v3/${INFURA_ID}"],
+      "dappnode": "http://geth.dappnode:8545",
```
ligi, line 23: better use fullnode.dappnode

Jakic007: If you agree, I think that the best option is to have web3.infura and web3.fullnode.dappnode since all of those are actualy web3 providers.

ligi: not sure I get what you mean - don't we need a JSON RPC endpoint and not a web3 provider?

edisinovcic: @ligi yes sure, it's just the naming. @Jakic007 leave it then fullnode.dappnode.

Issue #252, opened 2020-09-28
Validation Script


chriseth:
Provide lightweight stand-alone npm package that can be used as a commandline interface and from nodejs scripts that can consume metadata and source files and produces standard-json structures such that people can re-compile contracts to verify them.

Feature list:

- run the script with a list of files as commandline arguments, potentially also directories or zip files
- scan these files for metadata files and return a list of them
  - scan each file in the provided directories
  - if the file is a json file and contains the required keys for metadata files, consider it a metadata file
  - if the file is not a json file / is too large / a binary file, then scan for `"{\"compiler\":{\"version\".*?},\"version\":1}"`, un-quote this and consider this as a metadata file
- each metadata file is / can be identified by the contract it represents (and should be reported to the user like that): `metadata['settings']['compilationTarget']' - there should be a single key that contains the file name and the value is the contract name. The file name can be used to disambiguate if there are multiple contracts of the same name.
- for each metadata file / contract file look for the referenced source files

For each metadata file / contract, we can have a success (all files found) or a failure case (not all files found, or something malformatted).

Output (given for the commandline interface, suitably formatted as json if used via nodejs directly - the instructional texts are only given on the commandline):

```
MyContractName (my/file/name):
  Error: Malformed json

MyContractName2 (my/file/name):
  Error: Missing sources:
  The following files were not part in the list of files and directories provided.
  Please retrieve the files (potentially via ipfs) and re-run the script.
    - my/sourcefileA.sol (keccak256: 0x1234... bzzr://... ipfs://...)
    - my/otherfile.sol (keccak256: 0x1234... bzzr://... ipfs://...)
  7 other source files were found successfully.
 
TheOtherContractName (my/file/othername):
  Success!
  Compiled with Solidity 0.6.5
  https://solc-bin.ethereum.org/wasm/soljson-v0.6.6+commit.6c089d02.js
  https://solc-bin.ethereum.org/linux-amd64/solc-v0.6.6+commit.6c089d02
  To re-compile, use:
  echo "{...}" |   solc --standard-json -
  ...
```

Jakic007:
> if the file is not a json file / is too large / a binary file, then scan for "{\"compiler\":{\"version\".*?},\"version\":1}", un-quote this and consider this as a metadata file

@chriseth What is the case when metadata is inside a file that is not json parsable? We are not checking for the extension, we are trying to JSON.parse the file to pull out metadata out of it.

chriseth:
[...]

PR #282, opened 2020-10-19
Updated documentation, supported standard-json

edisinovcic:
Updating documentation and updating support for standard-json


services/validation/README.md
```diff
@@ -20,6 +92,7 @@ Savings (browser/Savings.sol):
   Compiled with Solidity 0.6.11
   https://solc-bin.ethereum.org/wasm/soljson-v0.6.11+commit.5ef660b1.js
   https://solc-bin.ethereum.org/linux-amd64/solc-linux-amd64-v0.6.11+commit.5ef660b1
+  To re-compile use: `sourcify-validation --prepare-json browser/Savings.sol:Savings Escrow.sol Main.sol Owned.sol provableAPI_0.6.sol Savings.sol metadata.json | solc --standard-json`
```
chriseth, line 95:
```suggestion
  To re-compile use: `sourcify-validation --prepare-json browser/Savings.sol:Savings Escrow.sol Main.sol Owned.sol provableAPI_0.6.sol Savings.sol metadata.json | solc --standard-json -`
```

chriseth: Or does it also work without?

FabijanC: It was tested and only works without the "-".

chriseth: ok, great!

services/validation/README.md
```diff
@@ -1,15 +1,128 @@
...
+
+Several tools, including [Truffle](https://www.trufflesuite.com/) and [Buidler](https://buidler.dev) store this file with each compilation.
+
+## Usage as a script
 
 * `npm install -g sourcify-validation`
-* `sourcify-validation path/to/file1 path/to/file2 path/to/dir path/to/zip`
-    * the script will scan through all the provided files, directories and zips searching for metadata and the belonging source files
-* If everything goes well, proceed here: [Sourcify](https://verification.komputing.org/)
+* `sourcify-validation my/repository`
```
chriseth, line 15: Is it clear that this is a directory relative to the current directory and not e.g. a github project name?

services/validation/README.md
```diff
@@ -1,15 +1,128 @@
...
+
+You can specify a sequence of files, directories and even zip files.
+The script will scan through all the provided files, searching for metadata and the belonging source files.
+
+If everything goes well, you can ask it to provide a "standard-json input file"
+that you just need to send to the compiler, etherscan, or whatever service you use
+for verification.
+
+Verification is also achievable through the use of [Sourcify](https://verification.komputing.org/). All you need to do is upload the files
```
chriseth, line 24: Shouldn't we use `sourcify.dev`?

services/validation/README.md
```diff
@@ -1,15 +1,128 @@
...
+you had provided to this script.
+
+## Options
+* -j, --prepare-json <path:name>
+  * Output only a JSON object to be used as standard-json input to solc.
+  * The expected argument ought to be of the form path:name
+  * E.g. `sourcify-validation --prepare-json foo/bar/Contract.sol:ContractName path/to/dir`
+  * Omitting either the path or the name is tolerated if unambiguous.
+* -S, --no-settings
+  * Omit the original settings when preparing a standard-json.
```
chriseth, line 34: What does that mean?

services/validation/README.md
```diff
@@ -1,15 +1,128 @@
...
+```
+
+If you don't have the required version of the compiler, run:
+* `wget https://solc-bin.ethereum.org/linux-amd64/solc-linux-amd64-v0.6.6+commit.6c089d02 -O solc-0.6.6`
 
-## Example - multiple sources
+Then run:
+* `sourcify-validation -j browser/1_storage.sol:Storage . | solc-0.6.6 --standard-json`
+
+If you use `jq`, the output will have a nice format:
```
chriseth, line 69: we can omit that with the pretty-printing option.

services/validation/README.md
```diff
@@ -61,17 +175,75 @@ ConvertLib (/home/user/dir/truffle-example/contracts/ConvertLib.sol):
...
+
+## Usage as a module
+* `npm install sourcify-validation`
+#### Process paths
+```typescript
+import { ValidationService } from 'sourcify-validation';
+
+const validationService = new ValidationService();
+const paths = ["path/to/file1", "path/to/file2"];
+const checkedContracts = validationService.checkPaths(paths);
```
chriseth, line 204: Can you add a comment that this does the main work?

services/validation/README.md
```diff
@@ -61,17 +175,75 @@ ConvertLib (/home/user/dir/truffle-example/contracts/ConvertLib.sol):
...
+
+function work(files) {
+  for (const file in files) {
+    console.log("\t" + files[file]);
+  }
+}
+
+checkedContracts.forEach(contract => {
+  if (contract.isValid()) {
```
chriseth, line 213: Is it possible that `checkPaths` just returns a json structure without funcitons?

services/validation/README.md
FabijanC: This was copied from #281; it can be switched back to the old line or it can be changed to e.g. `sourcify-validation path/to/repo/dir` if that is more intuitive

services/validation/README.md
FabijanC: That makes sense; this old link was present here even before the documentation was handed over to me.

services/validation/README.md
FabijanC: Regarding the omission of the path or the name, it is as you suggested here: https://github.com/ethereum/sourcify/issues/252#issuecomment-707670866
Examples of the omission can be added.
The -S flag covers the case when a user might not want to re-compile the contracts in a fresh manner; that is - without using the old settings. But if that is something that we don't want to have, it will be removed. I am aware of the motivational part of documentation which mentions the desire of "re-compiling the contract into the exact same binary".

services/validation/README.md
FabijanC: Actually the pretty-printing option of sourcify-validation only affects its output. Here the jq can be used to pretty-print the output of solc.

services/validation/README.md
```diff
@@ -61,17 +175,75 @@ ConvertLib (/home/user/dir/truffle-example/contracts/ConvertLib.sol):
...
+
+```
+
+## Usage as a module
+* `npm install sourcify-validation`
+#### Process paths
+```typescript
+import { ValidationService } from 'sourcify-validation';
+
+const validationService = new ValidationService();
```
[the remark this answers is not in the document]
FabijanC: All services of the sourcify project are implemented as classes with public methods. In this case, sourcify-validation service exports a ValidationService with the method to checkPaths.

services/validation/README.md
FabijanC: Sure.

services/validation/README.md
FabijanC: Do you mean the same json that the script outputs when run with the --prepare-json option?

PR #288, opened 2020-10-23
New version

Jakic007:
- monorepo structure
- services deployment CI
- separated s3 and ipfs between staging and prod env
- monitor containers resource access restricted


services/core/src/services/FileService.ts
```diff
@@ -0,0 +1,120 @@
...
+        const tag: Tag = {
+            timestamp: timestamp,
+            repositoryVersion: repositoryVersion
+        }
+        fs.writeFileSync(filePath, JSON.stringify(tag));
+    }
+
+}
+
+type Tag = {
```
edisinovcic, line 117: @Jakic007 Maybe we can move this to types?

Issue #277, opened 2020-10-16
Research sourcify support for layer 2 solutions


edisinovcic:
Solutions that are interesting in this area:
* optimism (https://optimism.io/)
* xDAI (https://www.xdaichain.com/)

edisinovcic:
@Jakic007 can you do a quick research on optimism and how can we support it via sourcify.
I will do one on xdai.

ligi:
xDAI should be easy - just another chain/chainId - can be treated like we treat the other networks as far as I see
with optimism I would wait until they really launch.

edisinovcic:
https://www.xdaichain.com/for-developers/developer-resources#json-rpc-endpoints

Jakic007:
Optimism has still not launched mainnet and they are currently only doing some small private testnets.
I am following them on their discord server to see when they launch.

PR #294, opened 2020-10-27
Add xdai support; refactor chain retrieval; add build:lerna

FabijanC:
Added xDai support on both backend and frontend. This was suggested in https://github.com/ethereum/sourcify/issues/277. There was code duplication with listing available chains in monitor and injector; this was solved by adding a new method to core/utils. A new script was added to package.json that enables building with lerna.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,21 +47,22 @@ aliases:
         name: Versions
         command: npm version
     - checkout
-    - restore_cache:
-        key: dependency-cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "package.json" }}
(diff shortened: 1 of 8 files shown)


FabijanC:
Should the checks be passing if the new implementation imports a function from sourcify-core that has not yet been published to npm?

Jakic007:
> Should the checks be passing if the new implementation imports a function from sourcify-core that has not yet been published to npm?

I have replaced the `npm install` command in the CI workflow with `lerna bootstrap` to replicate the same state that we have inside our docker containers that are running sourcify.

Jakic007 (approved):

PR #297, opened 2020-11-02
Using local Solc

FabijanC:
Before resorting to fetching solc in a remote manner, a check is made to see whether the requested solc version exists locally. The path where a check is made is specified as an environment variable SOLC_REPO. This required Injector to be modified, because it had issues loading asynchronously. Loading solc locally seems to have been too fast so Infura checking and subsequent chain initialization could not be be executed in time (i.e. injection process would work with an empty chain list). This was fixed by awaiting Infura check and chain initialization.

## Consequences
- This also required new methods for creating Injector, which might require changes in Readme (new Injector -> Injector.create).
- While this sped up the recompilation process by using local solc, a task (infura check) has to be waited for.

## Proposals
- A more efficient way would be not to instantiate Injectors at all, but try solutions such as the singleton pattern.
- Should the Infura checking be done in the same manner in future versions?

## Other changes
- There were some changes in tsconfig files to enable compilation.


services/core/package.json
```diff
@@ -34,6 +34,7 @@
     "cbor": "^5.1.0",
     "directory-tree": "^2.2.4",
     "fs-extra": "^9.0.1",
+    "solc": "^0.7.4",
```
Jakic007, line 37: As agreed we should remove this dependency from the core module.

services/core/src/utils/utils.ts
```diff
@@ -63,3 +68,53 @@ export function cborDecode(bytecode: number[]): any {
...
+ * 
+ * e.g.: "0.6.6+commit.6c089d02"
+ * 
+ * defaults to "latest"
+ * 
+ * @param log a logger to track the course of events
+ * 
+ * @returns the requested solc instance
+ */
+export function getSolc(version: string = "latest", log: {info: any}): Promise<any> {
```
Jakic007, line 88: This should go to @ethereum-sourcify/verification utils

services/core/src/utils/utils.ts
```diff
@@ -63,3 +68,53 @@ export function cborDecode(bytecode: number[]): any {
...
+            resolve(soljson);
+        });
+    }
+
+    log.info({loc: "[GET_SOLC]", version}, "Searching for solc remotely");
+
+    return new Promise((resolve, reject) => {
+        solc.loadRemoteVersion(version, (error: Error, soljson: any) => {
+            if (error) {
+                log.info({loc: "[GET_SOLC]", version}, "Could not find solc remotely");
```
Jakic007, line 112: log.error instead of info to search through logs easier in the future

services/verification/src/services/Injector.ts
```diff
@@ -39,10 +39,31 @@ export class Injector {
...
         this.fileService = config.fileService || new FileService(this.repositoryPath, this.log);
+    }
 
-        if (!this.offline) {
-            this.initChains();
+    /**
+     * Creates an instance of Injector. Does not wait for chains to initialize.
+     * @param config 
+     */
+    public static createFast(config: InjectorConfig = {}) {
```
Jakic007, line 48: This can be removed completely as agreed.

FabijanC:
Pushed fixes as proposed in the review. Reflected the Injector creation changes on monitor.ts and README.md. This will, however, have no effect once monitor.ts is rewritten.

services/core/src/utils/utils.ts
```diff
@@ -1,5 +1,10 @@
 import * as chainOptions from '../chains.json';
 import cbor from 'cbor';
+import fs from 'fs';
```
Jakic007, line 3: This is unused so it can be removed

services/core/src/utils/utils.ts
```diff
@@ -1,5 +1,10 @@
...
+import Path from 'path';
+const solc = require('solc');
```
Jakic007, line 5: This is also unused and a missing dependency so it should be removed.

services/core/src/utils/utils.ts
```diff
@@ -1,5 +1,10 @@
...
+const envPath = Path.join(__dirname, "..", "..", ".env");
```
Jakic007, line 6: This is also not needed here any more.

services/core/src/utils/utils.ts
```diff
@@ -1,5 +1,10 @@
...
+require("dotenv").config({ path: envPath });
```
Jakic007, line 7: Same as above comment

Jakic007, line 7: All of this should be removed. 
```
const solc = require('solc');
￼const envPath = Path.join(__dirname, "..", "..", ".env");
￼require("dotenv").config({ path: envPath });

```

Jakic007 (approved):

PR #310, opened 2020-11-10
Add Solidity syntax highlight

cleanunicorn:
Update `.gitattributes` to enable Solidity syntax highlight on GitHub.


Jakic007:
Hi and thanks for contributing,

According to this, there is no need to manually define that in the `.gitattributes` since this PR has been merged. https://github.com/github/linguist/pull/4936
Also, you can see it working here: https://github.com/ethereum/sourcify/blob/master/services/validation/test/files/single/1_Storage.sol

Are there any additional benefits that I am not aware of?

PR #322, opened 2020-11-25
Use sourcify URL instead of komputing

ligi:
also removed the wasteful slashes from the URLs

PS: we should also make the contractrepo a sourcify.dev subdomain for consistency


edisinovcic:
@ligi looks good, can you please create PR against the staging branch? Tnx

ligi:
base changed

Issue #320, opened 2020-11-19
Report if compilationTarget is malformed


FabijanC:
Both ValidationService and VerificationService make assumptions regarding the size of the compilationTarget property of metadata.   They expect it to contain a single key and value (file name and contract name), and don't report anything if it is not so. Perhaps both services (Validation and Verification) should have checks and report appropriate errors.

Issue #308, opened 2020-11-09
Use already verified contracts and ipfs in verification service


Jakic007:
We can use the ipfs hash from the manifest.json file to try to fetch missing sources.
If that fails, we can try to find missing files by the keccak256 from our repository.

PR #324, opened 2020-11-26
Services enhancement

FabijanC:
Regarding requests from https://github.com/ethereum/sourcify/issues/308:
- Added ipfs-fetch functionality for missing openzeppelin dependencies (the ones declared in .sol files with import "@ openzeppelin...").
- As discussed, implementation still doesn't use locally available verified contracts.
- Moved some common types/classes from services/validation to services/core.

Regarding requests from https://github.com/ethereum/sourcify/issues/320:
- Reporting if not a single compilationTarget in metadata.
- Added appropriate object size assert function in services/core to avoid code duplication: it is used in validation and verification service.

Possibly needs discussion:
- In order to reflect the true semantics, the property `sources` was renamed to `contracts` in types.ts:`InputData`. Its type was enforced from `any` to `CheckedContract[]`.
- Fetching of missing files is performed in a newly introduced method of the existing class `CheckedContract`. Perhaps use interface.


services/core/src/utils/utils.ts
```diff
@@ -62,4 +62,36 @@ export function cborDecode(bytecode: number[]): any {
...
+
+/**
+ * Asserts that the number of keys of the provided object is expectedSize.
+ * If not, logs an appropriate message (if log function provided) and throws an Error.
+ * @param object the object to check
+ * @param objectName the name of the object to use in error messages
+ * @param expectedSize the size that the object should have
+ * @param log optional log function
+ */
+export function assertObjectSize(object: any, expectedSize: number, log?: Function, info?: any) {
```
Jakic007, line 75: Can we make this `object` and `info` parameters type safe if it is expecting some particular form of an object.
Also, `@param objectName` that is listed above does not exist in the argument list, while `info` argument is missing in the params list above.

FabijanC: Regarding the type of `object`, the idea was to make the function generic in the sense that any JavaScript object may be checked. But as proposed, this function will be a private method of `ValidationService` as it will no longer be required by any other part of implementation once its calling is removed from `VerificationService`. The `log` and `info` parameters will then not have to be passed.

Jakic007 (approved):

Issue #321, opened 2020-11-24
Return partial match if full match is not returned


edisinovcic:

FabijanC:
Closed by https://github.com/ethereum/sourcify/pull/325

PR #325, opened 2020-11-26
Fetch partial matches if full not available

FabijanC:
Fetch partial if full match not available: https://github.com/ethereum/sourcify/issues/321
- FileController had two endpoints:
  - /tree/:chain/:address
  - /:chain/:address
- Added two more that return partial matches if full not found (the return object contains the property `status` indicating whether it was a full or a partial match):
  - /tree/any/:chain/:address
  - /any/:chain/:address
- Refactored FileController to reduce code duplication.

Possible discussion points:
- url path choice (/any)

Solves:
* https://github.com/sourcifyeth/remix-sourcify/issues/27


FabijanC:
Updated the documentation.

edisinovcic (approved):

Issue #339, opened 2020-12-16
Handle error when files are provided in different than `files` parameter in /verify endpoint


vbaranov:
Currently, `/verify` endpoint returns 
```
{
    "error": "Cannot read property 'data' of undefined"
}
```

when files are posted in different than `files` parameter

Issue #335, opened 2020-12-14
/checkByAddresses API endpoint returns internal server error if no params provided


vbaranov:
The response in that case:

```
{
    "error": "Cannot read property 'split' of undefined"
}
```

edisinovcic:
Tnx for the notice, this is a bug and will be addressed.

PR #342, opened 2021-01-04
Verification API 2

FabijanC:
- Changes to Verification API 1:
  - by virtue of returning a better message and/or status code:
    - fixes https://github.com/ethereum/sourcify/issues/339
    - fixes https://github.com/ethereum/sourcify/issues/335
  - so far, all POST requests were handled by the same handler, i.e. a POST request to any of the following would behave the same way: "/verify", "/", "/wild-path", "/another-wild-path"
    - now only "/verify" and "/" are legal Verification API 1 paths for POST requests (breaking?)
- Additions (Verification API 2):
  - session functionality
  - compiler selection (prerequisite for standard JSON)
  - every valid request returns an object with the status of currently pending contracts
- Pending:
  - ~the light request use-case (the user only wants to check the contract at a given address)~
  - ~standard-json support (this can be included in a separete pull request)~
  - discuss the the new endpoint structure (i.e. how the old and the new URL paths should be named), w.r.t. to possible breaking changes
- Changes have been reflected in the documentation (.md files and Postman collection)


.vscode/launch.json
```diff
@@ -295,6 +295,22 @@
         "services/validation/test/files/single"
       ]
     },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Validation service (standard-json file input)",
```
Jakic007, line 301: We can comment out this for now since we don't support standard json so it is a bit confusing.

Sourcify.postman_collection.json
```diff
@@ -1,39 +1,819 @@
...
+								"type": "text/javascript",
+								"exec": [
+									""
+								]
+							}
+						}
+					]
+				},
+				{
+					"name": "Verification API 2",
```
Jakic007, line 98: I am getting this error for every request inside this collection for APIv2
```json
{
    "error": "Cannot perform the request for the provided Content-Type"
}
```

src/server/controllers/VerificationController.ts
```diff
@@ -23,48 +27,95 @@ export default class VerificationController extends BaseController implements IC
...
+            res.status(200).send({ result: [result] }); // TODO array is an old expected behavior (e.g. by frontend)
+        }).catch(error => {
+            res.status(500).send({ error: error.message });
+        });
     }
 
-    checkByAddresses = async (req: any, res: Response) => {
-        let resultArray: Array<Object> = [];
+    private checkByAddresses = async (req: any, res: Response) => {
+        const missingParams = ["addresses", "chainIds"].filter(p => !req.query[p]);
```
Jakic007, line 109: We can probably use the express-validator here to have more consistent code.
Example is in the FileController.ts

src/server/controllers/VerificationController.ts
```diff
@@ -84,16 +135,220 @@ export default class VerificationController extends BaseController implements IC
...
+        for (const id in session.inputFiles) {
+            const pc = session.inputFiles[id];
+            inputSize += pc.content.length;
+        }
+
+        pathContents.forEach(pc => inputSize += pc.content.length);
+
+        if (inputSize > VerificationController.MAX_INPUT_SIZE) {
+            const msg = "Too much session memory used. Delete some files or restart the session.";
+            throw new BadRequestError(msg); // TODO 413 Payload Too Large
```
Jakic007, line 266: This return message should be added

Jakic007 (changes requested):
Look into this documentation https://express-validator.github.io/docs/
I think you can make these API endpoints cleaner by using this package. We are already using it in the FileController.ts.

Sourcify.postman_collection.json
FabijanC: Sorry, I forgot to mention this: All requests that upload files expect them to exist at certain locations on your computer. You can achieve this by setting Sourcify root directory as your Postman working directory. Perhaps we can switch to using Postman variables in the paths of the files being uploaded - and use a Potman pre-request script on every request requiring such variables. Modifying the working directory seemed as a simpler solution to me. Whichever option we choose, it should be documented (something I forgot to do). I can't seem to find a way to add a description to a whole Postman collection, so I'll just add the same description to the three requests including file upload. Also, Postman descriptions don't seem very _visible_ to me, so we can try with something better.

src/server/controllers/VerificationController.ts
FabijanC: What do you mean? A PayloadTooLarge error class?

Jakic007: Maybe PayloadTooLargeError can extend BadRequestError? Since that is bad request but it would be nice to return different error code for PayloadTooLargeError. It improves the API when used programmatically because it is easier to just check the status code.

Jakic007 (approved):

PR #372, opened 2021-02-03
Use the default chains.json

### LINKED ISSUE #363: Use unmodified chains.json
It would be easier for maintenance to use unmodified https://chainid.network/chains.json and use only networkID as identificator instead of network names.
Also, we can have another `supported.json` that will define which chainID will be monitored or supported for verification.

FabijanC:
- The chains.json file that provides the project with information about chains is the current version from [this source](https://chainid.network/chains.json). Information about being supported, monitored and dappnode endpoints are moved to the utils.ts file. Whether these should be in separate JSON files is a point of discussion.
- All mentions of networkId are changed to chainId as this is the actual value used.
- Fixes #363


Jakic007 (changes requested):
It would be better to have supported chains and fullnode endpoints inside `sourcify-chains.json` instead of using hardcoded values in `sourcify-core` service.

Also, if I am not mistaken, with this implementation, user can choose only between using infura or regular nodes. We can have environment variables that would enable using both at the same time for different chains.
For example, user wants to verify on mainnet and goerli but he is only running goerli node and wants to access mainnet via infura.

Jakic007 (approved):

PR #449, opened 2021-04-01
UI v3 draft

### LINKED ISSUE #440: Draft UI for API v2
UI that will be used for testing API v2 and session based verification.
It does not have to look nice, it just needs to work.

FabijanC:
A draft of the new UI that uses API v2.0.
The TCP port used by the UI draft is 1236.

Closes #440.


chriseth:
Can you do a screengrab video as soon as there is something to show?

FabijanC:
@chriseth https://drive.google.com/file/d/16e5r-ovwuCq-LsBdtzem6UtvrJTmgnWM/view?usp=sharing

chriseth:
Ah nice! This already looks really good!
Some comments:
 - can you display all files uploaded, not just the unused ones? Otherwise people might upload files multiple times. The "uploaded file" UI element can have a height limit and an internal scroll bar. If possible, display the files uploaded most recently at the top (or the bottom, depending on the default scroll position).
 - If the status is "Not verified", it would be great to give an indication about how this can be changed: The address input field and the chain selector could be red. Also it would be nice to have a short text explaining the current situation.

FabijanC:
@chriseth, that makes sense. However, all files can be seen: some in the unused section, some in the found sources section. The missing and found sources are displayed as two collapsable lists.

environments/ui-draft.yaml
```diff
@@ -0,0 +1,34 @@
...
+  env_file:
+  - .env
+  restart: always
+  networks:
+  - source-verify    
+
+services:
+  ui-draft:
+    <<: *project-base
+    build:
```
Jakic007, line 21: This should be replaced by `image` because you are building and pushing that image before and this yaml is only used for deployment. 
This was good for local testing :)

Jakic007 (changes requested):

environments/ui-draft.yaml
FabijanC: ah

Jakic007 (approved):

PR #463, opened 2021-05-06
Fix input in UI draft (close #456)

### LINKED ISSUE #456: Fix selected chain and address when switching between contracts in the UI draft
Steps to reproduce:
1. Upload a metadata file
2. Select a chain
3. Upload another metadata file

The chain selection field of the second contract is the same as of the first contract. The correct state is resolved upon reloading the page.

The same applies for address input.

FabijanC:
Checks whether the entered address is valid or empty; uploads to session only if so; otherwise just display the entered value. Persists the last valid input.


Jakic007 (changes requested):
Pasting address to verifier does not work

PR #473, opened 2021-05-24
Switch to custom nodes

FabijanC:
Switch to using custom nodes instead of Infura nodes for Goerli and Rinkeby.


Jakic007 (approved):

Jakic007:
TODO: use infura for tests because CircleCI can not access our local nodes

PR #478, opened 2021-05-31
Add verification-e2e test

### LINKED ISSUE #477: Add a verification e2e test
- This test is to be run after deployment on staging/master branch.
- It can use the same metacoin-source-verify project as the monitor-e2e test.
- Files shouldn't be pushed to IPFS so they don't get monitored.
- Some kind of cryptographic salt should be added to the source files to generate different IPFS hashes every time (basically append random chars to source).
- Since CI tests won't use the DB once https://github.com/ethereum/sourcify/pull/471 is merged, this will be the only way to test that functionality.

FabijanC:
Close #477.

- New part of staging/master workflow.
- Added to the nightly workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,8 +47,8 @@ aliases:
     steps:
       - checkout
       - run:
-          name: install node-fetch
-          command: npm install node-fetch
+          name: install node-fetch dotenv
+          command: npm install node-fetch dotenv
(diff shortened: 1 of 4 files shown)


Jakic007:
It should use node:12 @FabijanC

Jakic007 (approved):

PR #487, opened 2021-06-24
Adding contracts route and Telos networks

jesseschulman:
Per conversation on gitter, this PR includes 2 changes:

- It adds support for Telos EVM testnet and mainnet (mainnet going live in coming weeks)
- It adds a new API route for getting a list of contract addresses for a given chain so available contracts are discoverable

### DIFF
--- a/services/core/src/sourcify-chains.ts
+++ b/services/core/src/sourcify-chains.ts
@@ -162,21 +162,15 @@ export default {
         "txRegex": getBlockscoutRegex()
     },
     "40": {
-        "fullnode": {
-            "dappnode": "https://mainnet.telos.net/evm"
-        },
         "supported": true,
         "monitored": false,
-        "contractFetchAddress": "https://bscscan.com/" + ETHERSCAN_SUFFIX,
+        "contractFetchAddress": "https://mainnet.telos.net/v2/explore/evm/" + ETHERSCAN_SUFFIX,
         "txRegex": ETHERSCAN_REGEX
     },
     "41": {
-        "fullnode": {
-            "dappnode": "https://testnet.telos.net/evm"
-        },
         "supported": true,
         "monitored": false,
-        "contractFetchAddress": "https://bscscan.com/" + ETHERSCAN_SUFFIX,
+        "contractFetchAddress": "https://testnet.telos.net/v2/explore/evm/" + ETHERSCAN_SUFFIX,
         "txRegex": ETHERSCAN_REGEX
     }
 }
--- a/services/core/src/sourcify-chains.ts
+++ b/services/core/src/sourcify-chains.ts
@@ -160,5 +160,23 @@ export default {
         "monitored": true,
         "contractFetchAddress": "https://cchain.explorer.avax.network/" + BLOCKSCOUT_SUFFIX,
         "txRegex": getBlockscoutRegex()
+    },
+    "40": {
+        "fullnode": {
+            "dappnode": "https://mainnet.telos.net/evm"
+        },
+        "supported": true,
+        "monitored": false,
+        "contractFetchAddress": "https://bscscan.com/" + ETHERSCAN_SUFFIX,
+        "txRegex": ETHERSCAN_REGEX
+    },
+    "41": {
+        "fullnode": {
+            "dappnode": "https://testnet.telos.net/evm"
+        },
+        "supported": true,
+        "monitored": false,
+        "contractFetchAddress": "https://bscscan.com/" + ETHERSCAN_SUFFIX,
+        "txRegex": ETHERSCAN_REGEX
     }
 }
--- a/services/core/src/sourcify-chains.ts
+++ b/services/core/src/sourcify-chains.ts
@@ -4,6 +4,7 @@ const ETHERSCAN_REGEX = "at txn <a href='/tx/(.*?)'";
 const ETHERSCAN_SUFFIX = "address/${ADDRESS}";
 const BLOCKSCOUT_REGEX = "transaction_hash_link\" href=\"${BLOCKSCOUT_PREFIX}/tx/(.*?)\"";
 const BLOCKSCOUT_SUFFIX = "address/${ADDRESS}/transactions";
+const TELOS_SUFFIX = "v2/evm/get_contract?contract=${ADDRESS}";
 
 function getCustomURL(chainName: string, useOwn=false) {
     if (useOwn && process.env.TESTING !== "true") {
@@ -164,13 +165,13 @@ export default {
     "40": {
         "supported": true,
         "monitored": false,
-        "contractFetchAddress": "https://mainnet.telos.net/v2/explore/evm/" + ETHERSCAN_SUFFIX,
-        "txRegex": ETHERSCAN_REGEX
+        "contractFetchAddress": "https://mainnet.telos.net/" + TELOS_SUFFIX,
+        "isTelos": true
     },
     "41": {
         "supported": true,
         "monitored": false,
-        "contractFetchAddress": "https://testnet.telos.net/v2/explore/evm/" + ETHERSCAN_SUFFIX,
-        "txRegex": ETHERSCAN_REGEX
+        "contractFetchAddress": "https://testnet.telos.net/" + TELOS_SUFFIX,
+        "isTelos": true
     }
 }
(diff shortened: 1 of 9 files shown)


Jakic007:
Hi! I have changed the base for this PR to our `staging` branch instead of master. Could you please do a rebase on `ethereum:staging`?

poplexity:
> Hi! I have changed the base for this PR to our `staging` branch instead of master. Could you please do a rebase on `ethereum:staging`?

Rebased and updated.

FabijanC:
@poplexity Are these the recommended chain explorers:
https://telos.bloks.io/
https://telos-test.bloks.io/

Also, could I see an example of a contract page on the recommended explorer? Something like https://goerli.etherscan.io/address/0x5bd2477bf2aaadfeb46b128b80fc109b8e172347

services/core/src/sourcify-chains.ts
```diff
@@ -148,5 +148,22 @@ export default {
         "supported": true,
         "monitored": true,
         "graphQLFetchAddress": "https://rinkeby-indexer.arbitrum.io/graphql"
+    "40": {
```
FabijanC, line 151: I'm guessing there were some conflicts when rebasing, so this is now malformed (a `},` missing).

services/core/src/sourcify-chains.ts
```diff
@@ -148,5 +148,22 @@ export default {
...
+        "fullnode": {
```
FabijanC, line 152: You probably don't need the `fullnode` property. The RPC is already specified in [chains.json](https://github.com/ethereum/sourcify/blob/master/services/core/src/chains.json), which was last updated a few days ago and includes the required data.

services/core/src/sourcify-chains.ts
```diff
@@ -148,5 +148,22 @@ export default {
...
+            "dappnode": "https://mainnet.telos.net/evm"
+        },
+        "supported": true,
+        "monitored": false,
+        "contractFetchAddress": "https://bscscan.com/" + ETHERSCAN_SUFFIX,
```
FabijanC, line 157: [My previous comment](https://github.com/ethereum/sourcify/pull/487#issuecomment-868322676) about explorers concerns this line. In case a contract uses immutable state variables, Sourcify needs to know the chain's block explorer to be able to correctly verify. So I guess this will have to be changed to something other than the Binance Smart Chain explorer.

FabijanC (changes requested):

services/core/src/sourcify-chains.ts
FabijanC: Also, not really necessary, but the entries in this file are numerically sorted by key, so you can change the position too if you want to.

services/core/src/sourcify-chains.ts
poplexity: Removed from both networks.

services/core/src/sourcify-chains.ts
poplexity: Yeah, that must've happened in the rebase, added the line back in.

services/core/src/sourcify-chains.ts
poplexity: Fixed.

poplexity:
> @poplexity Are these the recommended chain explorers:
> https://telos.bloks.io/
> https://telos-test.bloks.io/

Those are explorers for the Telos native chain, the EVM is a smart contract running on native.  While all the EVM transactions will reflect there, it will have limited EVM transaction specific data.  We will have more EVM specific data in the lightweight explorer that I've updated the URLs for in `sourcify-chains.ts`.

> Also, could I see an example of a contract page on the recommended explorer? Something like https://goerli.etherscan.io/address/0x5bd2477bf2aaadfeb46b128b80fc109b8e172347

Part of getting added to Sourcify is so we can further develop the address page to include contract specific info.  It will link/reference/credit Sourcify for the source code and will also use Sourcify to load the ABI and provide unpacked/decoded function data.  Suggestions and/or requirements for this page from your perspective are welcome! :)

The URLs added in `sourcify-chains.ts` still have a version of the explorer that only has Telos native information, you can see [my developer instance](http://poplexity.net:7000/v2/explore/evm/address/0xc51fe232a0153f1f44572369cefe7b90f2ba08a5) which does have the latest.  It may get wiped often but I generally leave online.

We are also working on a more fully featured explorer which will be dedicated to the EVM and have also been in contact with Etherscan, but neither of those are expected to be available at mainnet launch.

FabijanC:
@poplexity In some cases (such as when contracts use `immutable` state variables), Sourcify cannot verify just by recompiling and comparing the recompiled bytecode against what is deployed on the blockchain. In that case the `creation data` (which is equal to `creation bytecode` + `constructor arguments`) is fetched. There are three approaches to this:
1. Lately we've been working on resolving this by querying our creation code database, but there have been issues with this approach.
2. If the chain RPC can tell whether a contract was present at a given block number, then binary search can be performed to quickly obtain the block number that first contained the contract and extract the creation transaction from that block. This approach is only feasible for ETH testnets because only in that case do we have RPC nodes with such historic data.
3. Otherwise, Sourcify utilizes each chain's block explorer to match the contract address (provided from the beginning of verification) to the transaction ID of the creation transaction. The `regex` part mentioned in `sourcify-chains.ts` is needed to actually extract the transaction ID from the page.

So if the transaction ID extraction process cannot be achieved for your chain, you would sacrifice the verification of some contracts (namely those using immutable state variables).

poplexity:
@FabijanC how do you handle internal transactions that deploy contracts via create/create2?  Due to the nature of how our EVM is built, we can do some things easier than others.  We will be able to provide a simple API to return info about a contract if given it's address, which I think can at least respond with the creation transaction and the data from that transaction.  We are currently adding tracking of internal transactions which will include create/create2 contract creation (like AMM pools for instance), and should be able to include their data in the same API with some additional flag to signal that it was internal.

FabijanC:
@poplexity Sourcify doesn't cover the case of contracts deployed with create/create2.

poplexity:
So would those show up as partially verified?  Or not verified at all?

And for the creation data, if we were able to provide an API that accepts the contract address and returns the transaction hash in which that contract was created (or the full transaction receipt), would that work?

FabijanC:
> So would those show up as partially verified? Or not verified at all?

Not verified at all.

> And for the creation data, if we were able to provide an API that accepts the contract address and returns the transaction hash in which that contract was created (or the full transaction receipt), would that work?

Yes, that would work.

poplexity:
@FabijanC I just wanted to follow up here and see what the status is for this to get merged as we plan to build some explorer functionality using the ABIs we'll retrieve from sourcify.  I see that we probably need a rebase to resolve some conflicts, I will do that shortly.  Do you also want to wait for us to have the API mentioned above available on testnet before you merge?  Thanks again!

FabijanC:
@poplexity 
Yes, it seems we should wait for your API to be available. When it becomes available, you can expand this PR by adding the API querying functionality to the [getCreationData method](https://github.com/ethereum/sourcify/blob/bd49dd465d44068dafc667a6f7cf8402d30880d7/services/verification/src/services/Injector.ts#L296). You can see how each case of getting the creation data is covered (unfortunately currently it's `if`s and `else`s). This will probably require adding an appropriate URL to the Telos entry in [sourcify-chains.ts](https://github.com/ethereum/sourcify/blob/03240b5bd7a3f5f17f9093c1c5b29a8aabcaa95d/services/core/src/sourcify-chains.ts) and possibly adding an attribute to the [InjectorChain class](https://github.com/ethereum/sourcify/blob/bd49dd465d44068dafc667a6f7cf8402d30880d7/services/verification/src/services/Injector.ts#L18).

Also, if you'd simply like to contribute, you can separate this PR into two because the part with adding contracts route can already be merged.

poplexity:
@FabijanC the latest commit should add support for querying the Telos API as you described.

I also considered implementing it in a more generic way that just assumes an HTTP GET and a JSON response and provides some JSON path to query the trxHash from, but to do that I probably would install something like [_.get](https://lodash.com/docs/#get) and I was unsure about adding packages just to solve our one use case.

To test this, you can use contract with hash `0xE9c31091868d68598Ac881738D159A63532d12f9`

This should result in a query to `https://testnet.telos.net/v2/evm/get_contract?contract=0xE9c31091868d68598Ac881738D159A63532d12f9` which then should find transaction hash of `0xbafad3f5b64219650a83cc016e9d474bb8ef97244336271c74dbdc003c4012cb` and get the input via web3.  The explorer for that transaction hash is [here](https://testnet.telos.net/v2/explore/evm/transaction/0xbafad3f5b64219650a83cc016e9d474bb8ef97244336271c74dbdc003c4012cb).

The source for that contract is [SoonToken](https://github.com/ethereum/sourcify/files/6943852/SoonToken.txt) (note github didn't like the .sol extension, so I used .txt)

FabijanC:
@poplexity LGTM, although the provided SoonToken test file doesn't use immutable variables, so verifying it doesn't rely on the API. But I've done some local tests and it seems to work, so I'll merge this onto `staging` and it will hopefully be available on http://staging.sourcify.dev when the CI finishes.

PR #503, opened 2021-08-25
Add docs/architecture.md

FabijanC:


docs/architecture.md
```diff
@@ -0,0 +1,230 @@
...
+    - when a new contract is deployed, its metadata's IPFS hash is extracted from the end of the bytecode
+    - the metadata is fetched from IPFS
+    - source files information and compiler version are read from the metadata file
+    - the source files are fetched and recompiled
+    - files are stored to the repo
+- UI:
+  - provides a user friendly interface for communicating with the server
+  - two purposes:
+    - check if an address is verified
+    - verify a contract
```
chriseth, line 25: Plus inspect the source code?

Jakic007: @chriseth That's what repository is for, right? UI just provides a link to the source code in the repository

FabijanC: You mean by the fact that a link to the source code in the repository is generated if the checking says that the contract is indeed verified? That can be added.

chriseth: Ah right!

PR #507, opened 2021-08-26
Add DB access functionality to Injector

FabijanC:
This should allow omitting the address and chain from verification requests. That is, these fields should be optional, and the creationData DB should be queried by recompiled creation bytecode to fetch the actual address and chain. This can result in multiple query results since the same contract bytecode exists on multiple addresses, thus yielding multiple (address, chain) pairs for the same bytecode input. Finally, after adapting the backend, the frontend should also be adapted to allow for address/chain omission.

### DIFF
--- a/services/core/src/services/FileService.ts
+++ b/services/core/src/services/FileService.ts
@@ -142,6 +142,7 @@ export class FileService implements IFileService {
         try {
             const storageTimestamp = fs.statSync(contractPath).birthtime;
             return [{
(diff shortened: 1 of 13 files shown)


Fkhaibari (approved):

Iluminex94 (approved):

marcocastignoli:
Finally closing this since we will implement this feature using the Alliance Database

PR #519, opened 2021-09-06
Bump ipfs to 0.9.1 - closes #514

### LINKED ISSUE #514: Update ipfs to 0.9.1﻿

ligi:


chriseth:
Are the merge commits intentional?

ligi:
@chriseth no - did this one from the github website - saw the diff was clean - and did not see the merge commits - IMHO we should not have merge commits in the first place - then we would also not see them here ..

chriseth:
I'm fine with either way, but I actually like explicit merge commits because it makes it clear which commits belong to a common feature.

Issue #495, opened 2021-07-31
IPNS/IPFS gateway is down


MysticRyuujin:
Can't load the ipns hash of `51qzi5uqu5dll0ocge71eudqnrgnogmbr37gsgl12uubsinphjoknl6bbi41p` from any IPFS gateway currently, even the one on the website `gateway.ipfs.io` which seems weird since I've pinned the ipns to my own nodes personally with pub/sub support.

Jakic007:
We are investigating this issue.
Thank you for letting us know!

ligi:
related #514 #518

more info on what is problematic - this is what `ipfs id` returns:

```
{
	"ID": "12D3KooWCQDnipR4ovmxZGcWhJDBJPUZLgewmhb9uz9R3zKBGXNF",
	"PublicKey": "CAESICZjpXqySMlWlZL5eFda1GJcdn8ouUEnjsFlZVMuhOSA",
	"Addresses": [
		"/ip4/127.0.0.1/tcp/4001/p2p/12D3KooWCQDnipR4ovmxZGcWhJDBJPUZLgewmhb9uz9R3zKBGXNF",
		"/ip4/127.0.0.1/udp/4001/quic/p2p/12D3KooWCQDnipR4ovmxZGcWhJDBJPUZLgewmhb9uz9R3zKBGXNF",
		"/ip4/178.19.221.38/tcp/2456/p2p/12D3KooWCQDnipR4ovmxZGcWhJDBJPUZLgewmhb9uz9R3zKBGXNF",
		"/ip4/178.19.221.38/udp/5561/quic/p2p/12D3KooWCQDnipR4ovmxZGcWhJDBJPUZLgewmhb9uz9R3zKBGXNF",
		"/ip6/::1/tcp/4001/p2p/12D3KooWCQDnipR4ovmxZGcWhJDBJPUZLgewmhb9uz9R3zKBGXNF",
		"/ip6/::1/udp/4001/quic/p2p/12D3KooWCQDnipR4ovmxZGcWhJDBJPUZLgewmhb9uz9R3zKBGXNF"
	],
	"AgentVersion": "go-ipfs/0.9.0/",
	"ProtocolVersion": "ipfs/0.1.0",
	"Protocols": [
		"/ipfs/bitswap",
		"/ipfs/bitswap/1.0.0",
		"/ipfs/bitswap/1.1.0",
		"/ipfs/bitswap/1.2.0",
		"/ipfs/id/1.0.0",
[...]

PR #520, opened 2021-09-07
Fix dockerized ipfs nodes unreachable

kuzdogan:
The previous `server` profile disables localhost discovery. According to [docs](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#profiles) this is for when running a server but breaks things when dockerized. Change should make the ipfs node reachable over the docker network and also from outside.

See #495


ligi (approved):
LGTM

ogwurujohnson (approved):
🚀

PR #528, opened 2021-09-17
Enable pubsub in ipfs daemon

MysticRyuujin:
I think this will help with the IPNS resolution issues? #495 as suggested by @wmitsuda


kuzdogan:
Merged into staging thanks

Issue #538, opened 2021-09-29
Implementing an npm module for contract call decoding


kuzdogan:
Idea
-----------------------
Implement an npm module for decoding a contract call into a more human-readable format using the ABI and userdoc ([NatSpec](https://docs.soliditylang.org/en/latest/natspec-format.html)) in the metadata of the contract.

The main functionality would take raw tx input data and contract address as input and human-readable contract call as output. If available, show the parameters with the NatSpec 

Initial implementation and development is at branch [contract-call-decoder](/ethereum/sourcify/tree/contract-call-decoder)

Example
----------------
Example on contract [0x7492F518e41D8610a677Fe84b39223b746449979](https://rinkeby.etherscan.io/address/0x7492F518e41D8610a677Fe84b39223b746449979) on Rinkeby.:
```solidity
/// @notice Creates `_amount` token to `_to`. Must only be called by the owner (MasterChef).
    function mint(address _to, uint256 _amount) public onlyOwner {
        _mint(_to, _amount);
        _moveDelegates(address(0), _delegates[_to], _amount);
    }
``` 

Calling `mint('0xAA6042aa65eb93C6439cDaeBC27B3bd09c5DFe94', 1000000000000000000)`

Tx input data in hex
```
0x40c10f19000000000000000000000000aa6042aa65eb93c6439cdaebc27b3bd09c5dfe940000000000000000000000000000000000000000000000000de0b6b3a7640000
```

if userdoc is not available should return something like:
```
calling function mint(address _to, uint256 _amount) with:
{
  _to: '0xAA6042aa65eb93C6439cDaeBC27B3bd09c5DFe94'
  _amount: 1000000000000000000
}
```

If available, should return also the documentation in addition to above: 
```
Creates 1000000000000000000 token to 0xAA6042aa65eb93C6439cDaeBC27B3bd09c5DFe94. Must only be called by the owner (MasterChef).
```

Usage
-------------------
The module can be a class that takes two parameters in constructor: an RPC and an optional IPFS/Swarm Gateway (can default to ipfs.io/ipfs/ and swarm-gateways.net).
```js
import Decoder from 'contract-call-decoder'

const decoder = new Decoder('https://rinkeby.infura.io/v3/<Access-Token>', { ipfs: 'https://ipfs.infura.io/ipfs/' })
[...]

PR #565, opened 2021-10-19
Implementing an npm module for contract call decoding

ogwurujohnson:
https://github.com/ethereum/sourcify/issues/538

This will serve as the primary branch for all things related to the wrapper until we complete development, then we can promote to staging


marcocastignoli:
I worked on a contract-call-decoder on a separate package https://github.com/marcocastignoli/contract-call-decoder

After merging https://github.com/ethereum/sourcify/pull/837, I will open a new PR to include the package mentioned above into `./packages` (like I did with the bytecode-utils package)

PR #566, opened 2021-10-19
Fetch metadata from IPFS

kuzdogan:


services/contract-call-decoder/build/index.d.ts
```diff
@@ -1,11 +1,30 @@
+import Web3 from 'web3';
```
ogwurujohnson, line 1: Can you add the build directory to gitignore, we don't need to be in git

kuzdogan: I thought about that, is it sufficient if it is sent to npm? So in `.gitignore` but not in `.npmignore`

ogwurujohnson: Any reason why we want to send it to npm? 
if there is none, i think its best we have just the `gitignore` for now, and specify the build directory in it, npm will ignore it

ogwurujohnson:
LGTM 🚀, resolve failing tests please

ogwurujohnson (approved):

kuzdogan:
The tests are consistently failing without any good info. Need to look into that.

PR #550, opened 2021-10-05
Add support for Optimistic Kovan network

elenadimitrova:
Adds optimistic kovan to the supported networks.

PR #581, opened 2021-11-09
Add support for Optimistic Mainnet network 

### LINKED ISSUE #352: Support for Optimism
Optimism network has launched and it is quite popular. Unfortunately, adding support for this L2 network is not straight forward as with most networks where we just have to add a new RPC URL.
Optimism is not EVM compatible, it uses a slightly modified version called OVM. For that reason, we would have to use a modified [solc](https://github.com/ethereum-optimism/solc-js) compiler for those verifications.

elenadimitrova:
Adds optimistic mainnet to the supported networks in addition to optimistic kovan which was merged in #550 

Pending optimistic mainnet regenesis scheduled for 11th November after which we can provide the sources off OE mainnet as well. 

Resolves #352

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,6 +315,7 @@ In order to fetch from the staging repository, replace https://repo.sourcify.dev
 - Ubiq
 - OneLedger Testnet Frankenstein
 - Syscoin Tanenbaum Testnet
+- Optimistic Mainnet
 
 ## Adding a new chain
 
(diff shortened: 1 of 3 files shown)


kuzdogan:
Congrats on the successful upgrade :clap: We'll be waiting for the mainnet contracts for testing and following bulk verification.

PR #591, opened 2021-11-22
Chain support Optimism & Boba

kuzdogan:


kuzdogan:
For some reason, Velas network commits show up in new commits but the changes are not effective.. Kept them here not to force-push

PR #625, opened 2022-01-12
Implement fetcher UI

ogwurujohnson:
PR implements the fetcher UI


ogwurujohnson:
> * some

- I already added the address verification. fetch request isn't made unless address is valid.
- No reason actually why I had it there, was just a nice to have nudge, to let a visiting user know we intentionally left the button out

kuzdogan:
Yeah I see the verification now sorry. Just then I was testing I've given an address with one character missing and typed a character then I got an error e.g. paste `0x801f3983c7baBF5E6ae192c84E1257844aDb4b4` then type `3`. Anyways, thanks :+1: merging

Issue #640, opened 2022-01-26
Handle `inline` not being parsed by the Solidity compiler between v0.8.2 - v0.8.4


kuzdogan:
Thanks to https://github.com/wighawag/hardhat-deploy/issues/263#issuecomment-1022251783 we found out that verification fails when the optimizer settings contain the `inline` option, due to the compiler not parsing the option. The bug was already noticed https://github.com/ethereum/solidity/issues/11418 and [fixed with v0.8.5](https://github.com/ethereum/solidity/releases/tag/v0.8.5).

For effected versions v0.8.2 - v0.8.4 remove the option `inline` in the standard JSON input before being fed to the Solidity compiler in `reformatMetadata` function at `core/src/utils/utils.ts`.

PR #661, opened 2022-02-16
Fix inliner not being parsed by solidity

ogwurujohnson:
This PR fixes issue #640


kuzdogan:
Can you also add a test with the contract you used for testing under `test/server.js`?

kuzdogan (changes requested):

PR #685, opened 2022-02-23
Turn ipfs filestore on, remove garbage collection

kuzdogan:


ligi:
what's the reason to turn off GC?

kuzdogan:
It was slowing down the publish operation and cluttering the logs *as I thought it was running automatic every hour* as set in the ipfs config:
```
  "Datastore": {
    "BloomFilterSize": 0,
    "GCPeriod": "1h",
```

but it seems it [needs to be enabled](https://docs.ipfs.io/concepts/persistence/#garbage-collection) upon daemon start. So thanks for the heads up @ligi

PR #692, opened 2022-03-02
Implement fixes for new UI

ogwurujohnson:


kuzdogan:
Just added styling to the "Add you project" button to keep consistency, and better keep the `hover:underline` and the clickable elements the same. 

Otherwise looks good :+1: Thanks

PR #706, opened 2022-04-03
add fuse chain

mul53:


kuzdogan:
I see that you deployed the contracts and added the tests thanks.

But you need to add the `sourcify-chains.ts` file as well.

mul53:
@kuzdogan i have added the changes

kuzdogan:
The immutable contract test is still failing. Could you please fix the constructor arg. input? I see you have given `123456` but the contract was apparently deployed with `100000`

mul53:
@kuzdogan changed the value

kuzdogan:
Looks good! We usually release weekly but if something else that needs to be pushed on production comes up, it will get deployed sooner. For now, you can see it on https://staging.sourcify.dev

PR #743, opened 2022-05-20
Add chain 486217935

scream4ik:


kuzdogan:
Turning off support for Gather Devnet as the chain seems to be shut down

PR #750, opened 2022-05-29
add candle to list of supported chains

samisbakedham:
Following chain support instructions, mentions adding the logo to the list of supported chains after verification


kuzdogan:
Sorry but we are featuring only the some of the most used networks. We can't feature candle at this stage.

samisbakedham:
> Sorry but we are featuring only the some of the most used networks. We can't feature candle at this stage.

Oh ok...

samisbakedham:
> Sorry but we are featuring only the some of the most used networks. We can't feature candle at this stage.

Are there a set of metrics that a chain needs to reach to be considered "most used"? Would love to set a goal for it to be featured.

kuzdogan:
Not really, that is more a subjective list and I don't think we will be actively following any metrics. Maybe we can revisit if the project takes off and we start getting a lot of page visits. It is insignificant at this stage

PR #757, opened 2022-06-09
Add chain 246

volodimir-sevastianov:


volodimir-sevastianov:
Hi @ligi @holiman @jpitts @Aldekein , not sure this bug is from my side, please check and let me know if i did smthing wrong

volodimir-sevastianov:
@kuzdogan okay, looks like previous issue is gone, but now there is another

kuzdogan:
Those tests sometimes fail randomly and is not an issue, just haven't have the time to fix it. It's all good :+1:

PR #761, opened 2022-06-21
Add chain 432201 (Dexalot Testnet)

jaimi-io:


services/core/src/sourcify-chains.ts
```diff
@@ -419,5 +419,10 @@ export default {
         "contractFetchAddress": "https://gw-explorer.nervosdao.community/" + BLOCKSCOUT_SUFFIX,
         "txRegex": getBlockscoutRegex()
     },
+    "432201": { // Dexalot Testnet
+        "supported": true,
+        "monitored": false,
+        "contractFetchAddress": `https://subnet-explorer-api.avax.network/v1.1/432201/` + AVALANCHE_SUBNET_SUFFIX,
```
connorch, line 425: If you switch this to this URL instead it should work!
https://subnet-explorer-api.avax-test.network/v1.1/432201/

PR #778, opened 2022-08-12
Import from etherscan

marcocastignoli:


src/server/controllers/VerificationController-util.ts
```diff
@@ -71,6 +86,18 @@ export function isVerifiable(contractWrapper: ContractWrapper) {
         && Boolean(contractWrapper.chainId);
 }
 
+// TODO: Check if we already have the full file path or just the file name.
```
kuzdogan, line 89: Reverted to using your way of looking up and falling back to the first item otherwise.

What do you mean here with the TODO?

marcocastignoli: I meant: "I have to learn if there is a better way to get the file name or ask to etherscan / sol devs if there is right way to extract the information"

services/verification/src/services/VerificationService.ts
```diff
@@ -9,7 +9,7 @@ export interface IVerificationService {
     findByAddress(address: string, chain: string): Match[];
     findAllByAddress(address: string, chain: string): Match[];
     inject(inputData: InputData): Promise<Match>;
-    verifyWithJSON(compilerVersion: string, contractName: string, compilerJson: any) : Promise<void>;
+    verifyWithJSON(compilerVersion: string, contractName: string, compilerJson: any) : Promise<Match>;
 }
 
 export class VerificationService implements IVerificationService {
```
AndikaShidiqPratma, line 15: services/verification/src/services/VerificationService.ts

AndikaShidiqPratma (approved):

Issue #779, opened 2022-08-15
Network error on repo.sourcify.dev/select-contract


vrde:
I wanted to verify my contract on EVMOS from the [contract verification page](https://evm.evmos.org/address/0x383f9F442011496C2555084BbA160182B77a8A19/contract_verifications/new).

That page redirected me to https://repo.sourcify.dev/select-contract/, and this is where I got this error:

![image](https://user-images.githubusercontent.com/134680/184611357-a8a8994f-9268-4cb9-85a2-652cad05fb46.png)

The failing request tries to load `http://sourcify.dev/server/chains`, that is an HTTP resource, not an HTTPS one, thus the *mixed content* error.

kuzdogan:
Thanks for reporting. Fixed with https://github.com/ethereum/sourcify/commit/a31b492aaa594a4fada92239fd496694ffa3bd44 https://repo.sourcify.dev/select-contract/

PR #780, opened 2022-08-15
Release

kuzdogan:
- Remote pinning
- Fix AWS keys
- Fix repo select contract HTTP fetch https://github.com/ethereum/sourcify/issues/779


kuzdogan:
Not a false positive but the secret leak was handled in the commits @Aldekein

Issue #811, opened 2022-09-28
Allow deeplinking of verified contract


ligi:
it would be great if: https://sourcify.dev/#/lookup has a way to deeplink to a contract

someone shared this:

> This is the RSociety contract if anyone wants to take a final look  https://optimistic.etherscan.io/address/0xfbf562a98ab8584178efdcfd09755ff9a1e7e3a2#code

and I wanted to provide an alternative sourcify link - but in the current flow it is not obvious (or possible) to deeplink to a contract in sourcify - only to the repo (but then you miss other possible chains or the remix integration) - so I think it would be nice when looking up that the contract becomes part of the URL to be able to deeplink

kuzdogan:
Agreed 👍 Works better as a query parameter if we just want to show this UI since we don't do server-side rendering. 

<img width="1784" alt="image" src="https://user-images.githubusercontent.com/13069972/192822937-58ce0e7e-39c8-4178-bb50-030c26b73328.png">

PR #813, opened 2022-09-28
Allow deeplinking of verified contract

orenyomtov:
I'm not a react developer, and I have a vague feeling I've been using all kinds of react-anti-patterns to get this to work, but... it does  work 😄

Fixes https://github.com/ethereum/sourcify/issues/811


marcocastignoli:
Hey @orenyomtov, thanks for the PR :tada: I rolled back to HashRouter because otherwise Sourcify will not work if accessed from IPFS. Also I rebased the PR on `staging` so tests will not fail (hopefully).

orenyomtov:
@marcocastignoli Awesome!

Is there a preview link for the staging environment UI?

Also, what do you think about making it conditional so that it uses the HashRouter when on IPFS, and the BrowserRouter on sourcify.dev? (e.g. Using https://www.npmjs.com/package/ipfs-react-router)

kuzdogan:
The staging server is unfortunately still down we had a failed disk there and now working on a new setup. 

I liked the idea but the package is not updated since long and I couldn't find the source code. Maybe we can do a similar workaround

PR #821, opened 2022-10-05
Add chain 99 (POA Network Core)

rimrakhimov:


rimrakhimov:
@kuzdogan There is still a CircleCI test failed, but I'm not sure that it is the problem of network added. If it is, I may research on the problem more

kuzdogan:
No that's expected and because of API keys missing. Thanks @rimrakhimov 👍

kuzdogan:
Turning off support as the chain seems to be shut down (RPCs failing) but I see the block explorer is working https://blockscout.com/poa/core

Please let me know if this is not the case

PR #837, opened 2022-11-11
Bytecode utils

marcocastignoli:


packages/bytecode-utils/README.md
```diff
@@ -0,0 +1,28 @@
...
+### Result
+
+```json
+{
+  "cbor": {
+    "bytes": "0xa2646970667358221220dceca8706b29e917dacf25fceef95acac8d90d765ac926663ce4096195952b6164736f6c634300060b",
+    "length": 51
+  },
+  "ipfs": "QmdD3hpMj6mEFVy9DP4QqjHaoeYbhKsYvApX1YZNfjTVWp",
+  "solcVersion": "0.6.11"
```
kuzdogan, line 26:
```suggestion
  "solcVersion": "0.6.11"
  "experimental": true
```

packages/bytecode-utils/package.json
```diff
@@ -0,0 +1,87 @@
+{
+  "name": "@ethereum-sourcify/bytecode-utils",
+  "version": "1.0.4",
+  "description": "Decode a contract's bytecode",
```
kuzdogan, line 4:
```suggestion
  "description": "Decode the CBOR encoded data at the end of an Ethereum contract's bytecode.",
```

packages/bytecode-utils/package.json
```diff
@@ -0,0 +1,87 @@
...
+  "main": "build/main/index.js",
+  "typings": "build/main/index.d.ts",
+  "module": "build/module/index.js",
+  "repository": "https://github.com/ethereum/sourcify",
+  "license": "MIT",
+  "keywords": [],
```
kuzdogan, line 10:
```suggestion
  "keywords": ["sourcify", "ethereum", "cbor", "contract", "smart-contract", "metadata", "solidity", "ipfs"],
```

packages/bytecode-utils/src/lib/bytecode.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+});
+
+test('bytecode decode cbor with ipfs property', (t) => {
+  t.is(
+    decode(BYTECODE_IPFS).ipfs,
+    'QmdD3hpMj6mEFVy9DP4QqjHaoeYbhKsYvApX1YZNfjTVWp'
+  );
+});
+
+test('bytecode decode cbor with unknow property (bzzr1)', (t) => {
```
kuzdogan, line 44: Is `bzzr1` unknown property?

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -0,0 +1,88 @@
...
+type DecodedObject = {
+  cbor: CBOR;
+  ipfs?: string;
+  solcVersion?: string;
+  [key: string]: string | CBOR | Uint8Array | undefined | boolean;
+};
+
+/**
+ * Decode contract's bytecode
+ * @param bytecode - hex of the bytecode
```
kuzdogan, line 21: With or without `0x`? Let's be explicit

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -0,0 +1,88 @@
...
+ * @returns Object describing the contract
+ */
+export const decode = (bytecode: string): DecodedObject => {
+  if (bytecode.length === 0) {
```
kuzdogan, line 25: If `0x` this will not throw.

kuzdogan: Maybe sanitize the bytecode before this to handle if first two bytes are `0x`

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -0,0 +1,88 @@
...
+      }
+      case 'solc': {
+        result.solcVersion = cborDecodedObject.solc.join('.');
+        break;
+      }
+      case 'experimental': {
+        result.experimental = cborDecodedObject.experimental;
+        break;
+      }
+      default: {
```
kuzdogan, line 63:
```suggestion
      // bzzr0 and bzzr1 are handled by the default case since they are hex encoded
      default: {
```

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -0,0 +1,88 @@
...
+export const get = async (
+  address: string,
+  provider: EthereumProvider
+): Promise<string> => {
+  const bytecode = await provider.request({
+    method: 'eth_getCode',
+    params: [address, 'latest'],
+  });
+  return bytecode as string;
+};
```
kuzdogan, line 88: Hmm do we really need this? I think this is not the job of this library. And I couldn't find the source code of the dependency https://www.npmjs.com/package/ethereum-provider

kuzdogan (changes requested):

packages/bytecode-utils/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Marco Castignoli
```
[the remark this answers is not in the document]
marcocastignoli: it's better to put Sourcify

packages/bytecode-utils/src/lib/bytecode.ts
marcocastignoli: That is just a type, you need to pass a provider when using it. I think it makes sense as long as we don't force specific dependecy to be used (web3, ethers, ...). This works in every environemnt because all the libraries have an Ethereum Provider, it's an EIP I believe.

packages/bytecode-utils/LICENSE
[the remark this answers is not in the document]
marcocastignoli: I used the same license in `./LICENSE`

lerna.json
```diff
@@ -1,14 +1,23 @@
...
-  "packages": [".", "ui", "services/*"],
+  "packages": [
+    ".",
+    "ui",
+    "services/*",
+    "packages/*"
+  ],
   "version": "0.1.5"
-}
+}
```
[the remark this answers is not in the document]
marcocastignoli: indeed

packages/bytecode-utils/src/lib/bytecode.ts
kuzdogan: Where would this be needed? I still don't see any use case or need for this.

marcocastignoli: Here for example: https://github.com/marcocastignoli/contract-call-decoder/blob/6304b2db3d88f80747e74e53d472c15a2e6085a2/src/lib/ContractCallDecoder.ts#L46

packages/bytecode-utils/package.json
```diff
@@ -0,0 +1,87 @@
...
+    "cov:check": "nyc report && nyc check-coverage --lines 100 --functions 100 --branches 100",
+    "reset-hard": "git clean -dfx && git reset --hard && yarn",
+    "prepare-release": "run-s reset-hard test cov:check"
+  },
+  "engines": {
+    "node": ">=10"
+  },
+  "dependencies": {
+    "@ethersproject/bytes": "^5.7.0",
+    "@types/node": "^18.11.4",
```
kuzdogan, line 33: Should be devDep?

marcocastignoli: probably, I'll check

packages/bytecode-utils/src/lib/bytecode.ts
kuzdogan: I still think this is not this package's job and if they need a functioning Provider anyway, they'll likely use the wrapper function e.g. `web3.eth.getCode` anyway. But ok if you think it makes sense.

marcocastignoli: Now I have doubts :S We have to understand if it makes sense to offer a shortcut to fetch the ABI, or we should push developers to use theirs library's option

kuzdogan: When in doubt, I usually opt in for not adding things but adding them when the need arises.

kuzdogan:
@marcocastignoli 
What exactly are you asking here about this solution? Important is that the tests run when we run repo tests i.e. `npx lerna run test --stream`.

Trimming is also something we can move to this package, separate from the `decode` function. Good point.

marcocastignoli:
* ok `npx lerna run test --stream` automatically runs test in each project so tests are not needed, I didn't know that.
* ok for the trim function

packages/bytecode-utils/src/lib/bytecode.ts
marcocastignoli: removed

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -77,18 +68,46 @@ export const decode = (bytecode: string): DecodedObject => {
 };
 
 /**
- * Get contract's bytecode given an address and an rpc
- * @param address - contract's address
- * @param provider - provider used to get the bytecode
- * @returns bytecode
+ * Split bytecode into execution and auxdata
+ * @param bytecode - hex of the bytecode with 0x prefix
+ * @returns array first element is execution bytecode, second element is auxdata
```
kuzdogan, line 73:
```suggestion
 * @returns string[] - [ executionBytecode, auxdata?, cborBytesLength?] all as hexStrings
```

src/server/controllers/VerificationController.ts
```diff
@@ -262,9 +262,12 @@ export default class VerificationController
...
       contract: checkedContract,
     };
-    const result = await this.verificationService.inject(inputData);
-
-    res.send({ result: [result] });
+    try {
+      const result = await this.verificationService.inject(inputData);
+      res.send({ result: [result] });
+    } catch (e) {
+      throw Error("An error occurred while verifying this contract");
```
kuzdogan, line 269: Why are we catching with a custom message here? This means the error messages inside `Injector` won't be forwarded to the user

packages/bytecode-utils/src/lib/bytecode.spec.ts
```diff
@@ -19,22 +18,24 @@ const BYTECODE_EXPERIMENTAL = readFileSync(
...
+  `${BYTECODES_FOLDER}/withoutauxdata.hex`
+).toString();
 
-test("get contract's bytecode by address and provider", async (t) => {
-  const ethereumProvider = provider('https://eth-mainnet.public.blastapi.io');
-  t.is(
-    await get('0x00000000219ab540356cBB839Cbe05303d7705Fa', ethereumProvider),
-    BYTECODE_IPFS
-  );
+test('return bytecode without auxdata while splitting bytecode into execution bytecode and auxadata', (t) => {
```
kuzdogan, line 25: Implies the returned bytecode will be return excluding the auxdata. Didn't get it first.
```suggestion
test('return the whole bytecode when the bytecode that doesn't contain auxdata', (t) => {
```

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -77,18 +68,46 @@ export const decode = (bytecode: string): DecodedObject => {
...
-    method: 'eth_getCode',
-    params: [address, 'latest'],
-  });
-  return bytecode as string;
+export const splitAuxdata = (bytecode: string): string[] => {
+  if (bytecode.length === 0) {
+    throw Error('Bytecode cannot be null');
+  }
+  if (bytecode.substring(0, 2) !== '0x') {
+    throw Error('Bytecode should start with 0x');
```
kuzdogan, line 80: Why not just handle bytecodes without 0x?
```suggestion
    bytecode = "0x" + bytecode
```

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -0,0 +1,113 @@
...
+ * Decode contract's bytecode
+ * @param bytecode - hex of the bytecode with 0x prefix
+ * @returns Object describing the contract
+ */
+export const decode = (bytecode: string): DecodedObject => {
+  if (bytecode.length === 0) {
+    throw Error('Bytecode cannot be null');
+  }
+  if (bytecode.substring(0, 2) !== '0x') {
+    throw Error('Bytecode should start with 0x');
```
kuzdogan, line 27: Also here, why not just handle bytecodes without 0x?
```suggestion
    bytecode = "0x" + bytecode
```

kuzdogan (changes requested):

src/server/controllers/VerificationController.ts
marcocastignoli: Is it ok? Because it's a low level message. Do we want to show it?

marcocastignoli: yeah it's right, I'll remove this catch

PR #840, opened 2022-11-15
Add chain 28528 (Optimism Bedrock Alpha)

rimrakhimov:


kuzdogan:
Turning support for this chain off as it seems it is not used anymore. The RPC haven't been working since weeks

PR #842, opened 2022-11-16
Add contract call decoder

marcocastignoli:


kuzdogan:
I guess it needs a `package-lock`

marcocastignoli:
> I guess it needs a `package-lock`

https://github.com/ethereum/sourcify/blob/71ae3d0d3d2576f5d7204966c50378141e5aebdf/packages/contract-call-decoder/package-lock.json

kuzdogan:
I wonder why the test is failing then

marcocastignoli:
for some reason in the CI there is an error that I don't have locally, now I'm trying to fix the dependencies in question

packages/contract-call-decoder/README.md
```diff
@@ -0,0 +1,40 @@
...
+const tx = {
+  to: '0xD4B081C226Bc8aBdaf111DEf54c09E779ad29428',
+  data: '0xcea299370000000000000000000000000000000000000000000000000000000000000002',
+};
+```
+
+### Get the notice using metadata taken from Sourcify
+
+```ts
+const notice = await evaluateCallDataFromTx(tx, { chainId: 5 }); // Set the new value 4
```
kuzdogan, line 29: I feel the main function should not be named this. It's not evaluating the calldata, evaluate is when filling the NatSpec dynamic expression. It's rather "decoding" the contract call/tx, as the lib name suggests. We can discuss this.
```suggestion
const notice = await decodeContractCall(tx, { chainId: 5 }); // Set the new value 4
```

kuzdogan (changes requested):

packages/contract-call-decoder/src/lib/ContractCallDecoder.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  const functionSignatureHash = tx.data.slice(0, 10);
+
+  const { selector } = findSelectorAndAbiItemFromSignatureHash(
+    functionSignatureHash,
+    metadata.output.abi
+  );
+
+  const evaluatedString = await evaluate(
+    metadata.output.userdoc.methods[selector].notice,
```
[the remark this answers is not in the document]
marcocastignoli: idk, we have to decide which information we want to return:
* parsed notice for sure
* abi?
* devdoc?
* userdoc?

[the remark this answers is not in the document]
marcocastignoli: proposed structure, tell me if it's ok to implement @kuzdogan 
```json
{
    "contract": {
        "author": "",
        "title": "",
        "details": "",
        "@custom:fields": ""
    },
    "method": {
        "name": "",
        "details": "",
        "params": [
            { "key": "", "detail": "", "value": ""}
        ],
        "returns": "",
        "notice": "",
        "@custom:fields": ""
    }
}
```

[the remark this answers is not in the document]
kuzdogan: What I was trying to add in the call is that the information from the ABI JSON is missing. Take this example
https://repo.sourcify.dev/contracts/full_match/1/0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552/metadata.json
```
    function execTransactionFromModuleReturnData(
        address to,
        uint256 value,
        bytes memory data,
        Enum.Operation operation
    ) public returns (bool success, bytes memory returnData) {
```
```json
      {
        "inputs": [
          {
            "internalType": "address",
            "name": "to",
            "type": "address"
          },
          {
            "internalType": "uint256",
            "name": "value",
            "type": "uint256"
          },
          {
            "internalType": "bytes",
            "name": "data",
            "type": "bytes"
          },
          {
            "internalType": "enum Enum.Operation",
            "name": "operation",
            "type": "uint8"
          }
        ],
        "name": "execTransactionFromModuleReturnData",
        "outputs": [
          {
            "internalType": "bool",
            "name": "success",
            "type": "bool"
          },
          {
            "internalType": "bytes",
            "name": "returnData",
            "type": "bytes"
          }
        ],
        "stateMutability": "nonpayable",
        "type": "function"
      },
```

A method can have input params and output params as stated in the ABI JSON. Importantly each param will have a type and an internal type, which are missing above. I think we can add the `value` and the `detail` to this format and include it under the `method`. wdyt?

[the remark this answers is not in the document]
kuzdogan:
```json
{
    "contract": {
        "author": "",
        "title": "",
        "details": "",
        "@custom:fields": ""
    },
    "method": {
        "name": "execTransactionFromModuleReturnData",
        "details": "Allows a Module to execute a Safe transaction without any further confirmations and return data",
        "inputs": [
          {
            "internalType": "address",
            "name": "to",
            "type": "address",
            "detail": "Destination address of module transaction",
          },
          {
            "internalType": "uint256",
            "name": "value",
            "type": "uint256",
            "value": "Ether value of module transaction"
          },
          {
            "internalType": "bytes",
            "name": "data",
            "type": "bytes"
            "detail": "Data payload of module transaction"
          },
          {
            "internalType": "enum Enum.Operation",
            "name": "operation",
            "type": "uint8",
            "detail": "Operation type of module transaction"
          }
        ],
        "outputs": [
          {
            "internalType": "bool",
            "name": "success",
            "type": "bool",
          },
          {
            "internalType": "bytes",
            "name": "returnData",
            "type": "bytes"
          }
        ],
        "returns": "",
        "notice": "",
        "@custom:fields": ""
    }
}
```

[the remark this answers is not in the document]
marcocastignoli: I'm a bit confused by this.. Was it a mistake? `detail` and `value` are always there right? It's not one or the other. The value should contain the actual parameter's value in the call?

```json
          {
            "internalType": "address",
            "name": "to",
            "type": "address",
            "detail": "Destination address of module transaction",
          },
          {
            "internalType": "uint256",
            "name": "value",
            "type": "uint256",
            "value": "Ether value of module transaction"
          },
```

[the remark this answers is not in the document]
kuzdogan: Yes it's a mistake, sorry

packages/contract-call-decoder/package.json
```diff
@@ -0,0 +1,104 @@
+{
+  "name": "@ethereum-sourcify/contract-call-decoder",
+  "version": "0.2.0",
+  "description": "-",
```
kuzdogan, line 4:
```suggestion
  "description": "Library to decode Ethereum smart contract calls into human-readable descriptions using ABI and NatSpec",
```

packages/contract-call-decoder/src/lib/ContractCallDecoder.ts
```diff
@@ -0,0 +1,191 @@
...
+    contract: {
+      author: devdoc.author,
+      title: devdoc.title,
+      details: devdoc.details,
+      custom: customFieldsContract,
+    },
+    method: {
+      selector,
+      abi: abi,
+      details: devdoc.methods[selector]?.details,
```
kuzdogan, line 183: Are we not evaluating the method devdoc?

packages/contract-call-decoder/src/lib/ContractCallDecoder.ts
```diff
@@ -0,0 +1,191 @@
...
+    readonly selector: string;
+    readonly details: string;
+    readonly abi: FunctionFragment;
+    readonly returns: string;
+    readonly notice: string;
+    readonly decodedParams: readonly DecodedParam[];
+    readonly params: { readonly [index: string]: unknown };
+    readonly custom: {
+      readonly [index: string]: string;
+    };
```
kuzdogan, line 123: I guess some of these need to be optional?

packages/contract-call-decoder/src/lib/utils.ts
```diff
@@ -0,0 +1,59 @@
+import { BigNumber } from 'ethers';
+
+function arrayContainsMixedTypeKeys(array): boolean {
```
kuzdogan, line 3: Can you add some documentation to what this does and why

packages/contract-call-decoder/src/lib/ContractCallDecoder.ts
```diff
@@ -0,0 +1,191 @@
...
+  return contractMetadataJSON;
+}
+
+export const evaluate = async function name(
+  expression: string,
+  abi,
+  transaction: Transaction,
+  provider?
+): Promise<string> {
+  return await radspec(expression, abi, transaction, provider);
```
kuzdogan, line 76: I see `provider` in `radspec` is mandatory but optional in the function param. Also `provider` and `abi` are `any` type here. Shouldn't this be not allowed in TS?

packages/contract-call-decoder/tsconfig.json
```diff
@@ -0,0 +1,47 @@
...
+    "outDir": "build/main",
+    "rootDir": "src",
+    "moduleResolution": "node",
+    "module": "commonjs",
+    "declaration": true,
+    "inlineSourceMap": true,
+    "esModuleInterop": true /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */,
+    "resolveJsonModule": true /* Include modules imported with .json extension. */,
+
+    // "strict": true /* Enable all strict type-checking options. */,
```
kuzdogan, line 14: Let's turn strict on
```suggestion
    "strict": true /* Enable all strict type-checking options. */,
```

kuzdogan (changes requested):

kuzdogan:
Requested some changes. I think it makes more sense to do the review after turning on Typescript `strict`

packages/contract-call-decoder/src/lib/ContractCallDecoder.ts
marcocastignoli: we are not, right now radspec is not built to evaluate it, we should edit the lirbary, but I don't know if it is worth it

kuzdogan: What is the difference between evaluating a `notice` string and a `details` string?

marcocastignoli: notice is in userdoc, dev becomes details in devdoc

@notice | Explain to an end user what this does | contract, library, interface, function, public state variable, event
-- | -- | --
@dev | Explain to a developer any extra details | contract, library, interface, function, state variable, event

packages/contract-call-decoder/src/lib/utils.ts
marcocastignoli: Here you displayed also another big problem, I included the entire ethers library while I should have inclueded just the [bignumber package](https://github.com/ethers-io/ethers.js/tree/master/packages/bignumber)

packages/contract-call-decoder/src/lib/ContractCallDecoder.ts
marcocastignoli: I superficially looked at the code last time, radspec acutally supports decoding other strings.

packages/contract-call-decoder/.eslintrc.json
```diff
@@ -0,0 +1,33 @@
...
+  "parser": "@typescript-eslint/parser",
+  "env": { "es6": true },
+  "ignorePatterns": ["node_modules", "build", "coverage"],
+  "plugins": ["import", "eslint-comments", "functional"],
+  "extends": [
+    "eslint:recommended",
+    "plugin:eslint-comments/recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:import/typescript",
+    "plugin:functional/lite",
```
kuzdogan, line 12: Are we using `functional` for a reason? This is causing `functional/no-let` errors

kuzdogan (changes requested):
Some checks need to be added. 

In general, we should throw and not return false I think, if there's no particular reason to do so.

Also an end-to-end example in README would be useful for people without context. I can write that if you want.

packages/contract-call-decoder/src/lib/ContractCallDecoder.ts
```diff
@@ -0,0 +1,210 @@
...
+
+export enum MetadataSources {
+  Sourcify,
+  BytecodeMetadata,
+}
+
+type GetMetadataOptions = {
+  readonly source?: MetadataSources;
+  readonly chainId?: number;
+  readonly address?: string;
```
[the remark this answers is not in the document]
marcocastignoli: it's not mandatory because the field can be passed also in the tx object
https://github.com/ethereum/sourcify/blob/35ba02efc4c152fc550787ad515f4a6659ef2edb/packages/contract-call-decoder/src/lib/ContractCallDecoder.ts#L136

packages/contract-call-decoder/src/lib/ContractCallDecoder.ts
```diff
@@ -0,0 +1,210 @@
...
+  let contractMetadataJSON;
+  if (options.source === MetadataSources.Sourcify) {
+    try {
+      const req = await fetch(
+        `${options.sourcifyProvider}/contracts/full_match/${options.chainId}/${options.address}/metadata.json`
+      );
+      contractMetadataJSON = await req.json();
+    } catch (e) {
+      console.log(e);
+      return false;
```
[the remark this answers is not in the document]
marcocastignoli: I was just following the functional way to do so, but it makes no sense since the rest of the project doesn't.
https://github.com/eslint-functional/eslint-plugin-functional/blob/main/docs/rules/no-throw-statement.md

kuzdogan (approved):
LGTM 🚀

Issue #846, opened 2022-11-17
Support for simulating creation bytecode with constructor arguments


kuzdogan:
For the cases when it is not possible to match a contract with its deployed bytecode (e.g. immutables) it might be possible to execute the creation code of the compiled contract with the constructor arguments to see if we will get the same on-chain deployed bytecode. 

This can also be useful for the verification of contracts created by factories, where the tx that created the contract does not consist of the creation code itself.

PR #864, opened 2022-12-05
Creation code simulation

kuzdogan:
#846 

- Adds the functionality to verify contracts by "simulating" the creation bytecode. 

  This means we execute the creation bytecode compiled from the given source code and metadata with additional `contextVariables` such as the constructor arguments and msg.sender. These are needed if any immutable variables get assigned by these variables. These variables are saved in a file called `contextVariables.json` in the contract repo.

  The bytecode execution is simulated by the [ethereumjs EVM](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/evm)

- Changes the create2 API
  
  Previously the create2 API was accepting the constructor arguments as an array of parameter objects. Changes this to accept `abiEncodedConstructorArguments`. The abi-encoding is generated in the UI. Note that this is a breaking change but since the create2 API is not being used by anyone, I don't think we need a major version upgrade.
  
- Typings for session and non-session verify requests.
  inserts the Session type into Request by "declaration merging".
Declaration merged Session type contains our custom session fields.
No need to typecast into MySession unsafely.

- Add basic validation on routes, but should be done better with full schema validation.

- Creation data scraping:
  
  Request the tx from multiple RPCs sequentially

- Also fixes create2 address calculation logic.
  Previously the hex of the salt was being taken which is incorrect. The create2 calculation does not have the salt hash as input but salt in 32 byte hex string.

- Shared "ConstructorArguments" component to be both used by the normal verification and the create2 verification. It handles the display and the abi-encoding of the constructor arguments. Also possible to input raw abi-encoded constructor arguments instead of each single argument e.g. `0x00000...aa`.

TODO:
- [x] Add a test case for a factory deployed contract (i.e. will use "simulation") with empty constructor arguments.


Bigrig7272gg (approved):

services/verification/src/services/Injector.ts
```diff
@@ -242,87 +257,89 @@ export class Injector {
...
       address,
       chainId: chain,
       status: null,
-      encodedConstructorArgs: undefined,
+      abiEncodedConstructorArguments: undefined,
       libraryMap: undefined,
+      contextVariables: undefined,
     };
 
-    if (deployedBytecode && deployedBytecode.length > 2) {
```
marcocastignoli, line 259: why this check is gone?

services/verification/src/services/Injector.ts
```diff
@@ -356,6 +373,114 @@ export class Injector {
...
+      chainId: parseInt(chainId),
+      defaultHardfork: evmVersion,
+    });
+    const eei = new EEI(stateManager, common, blockchain);
+
+    const evm = new EVM({
+      common,
+      eei,
+    });
+    if (creationBytecode.startsWith("0x")) {
```
marcocastignoli, line 395: maybe we could start using some utility like with0x() or without0x() considering that this kind of check is so common. Not a priority btw, we can do it later

services/verification/src/services/Injector.ts
```diff
@@ -689,44 +838,31 @@ export class Injector {
...
-      contractBytecode: compilationResult.creationBytecode,
-      constructorTypes: constructorArgsTypes,
-      constructorArgs: constructorArgsValues,
-    });
+      compilationResult.creationBytecode,
+      abiEncodedConstructorArguments
+    );
 
-    if (create2Address !== computedAddr) {
+    if (create2Address.toLowerCase() !== computedAddr.toLowerCase()) {
```
marcocastignoli, line 860: I don't know if we want this check... Because the users is expecting that the create2Address they pass are exactly the addresses computed by us. It would be better to do a checksum verification returning an error in the case the passed create2address is not checksummed

test/server.js
```diff
@@ -899,6 +916,173 @@ describe("Server", function () {
...
+        "FactoryTest3.sol"
+      );
+      const sourceBuffer = fs.readFileSync(sourcePath);
+
+      const res = await chai
+        .request(server.app)
+        .post("/")
+        .field("address", childAddress)
+        .field("chain", defaultContractChain)
+        .field("msgSender", factoryAddress)
```
marcocastignoli, line 1002: isn't it better to handle the the global variables as an object in which msgSender is one of the properties?

marcocastignoli: (like I suggested you with the stored file on the repo)

src/server/controllers/VerificationController.ts
```diff
@@ -952,6 +955,16 @@ export default class VerificationController
         .custom(
           (chain, { req }) => (req.chain = this.validateSingleChainId(chain))
         ),
+      body("msgSender")
```
marcocastignoli, line 958: isn't it better to handle the the global variables as an object in which msgSender is one of the properties?

services/core/src/utils/types.ts
```diff
@@ -17,6 +17,8 @@ export interface InjectorInput {
   contract: CheckedContract;
   bytecode?: string;
   creationData?: string;
+  constructorArguments?: string;
+  msgSender?: string;
```
marcocastignoli, line 21: isn't it better to handle the the global variables as an object in which msgSender is one of the properties?

services/verification/README.md
```diff
@@ -36,6 +36,9 @@ interface InjectorInput {
   addresses: string[]; // For now only one address is supported
   contracts: CheckedContract[]; // Solidity and metadata files
   bytecode?: string; // If you already have bytecode of the deployed contract
+  creationData?: string;
+  constructorArguments?: string;
+  msgSender?: string;
```
marcocastignoli, line 41: isn't it better to handle the the global variables as an object in which msgSender is one of the properties?

marcocastignoli:
I love the UI :heart:, I think that my only doubt is about msgSender as the only parameter in the requests. I would rather use a contextVariable object that contains msgSender

services/verification/src/services/Injector.ts
kuzdogan: Yes I'm doing the check non-sparingly. I wonder if there's a usual way to deal with it. I assume always having the 0x is the way to go like `sanitize0x(string)`

services/verification/src/services/Injector.ts
kuzdogan: Checksummed and non-cheksummed addresses are not different than each other. Bytewise `0xAbCdEf` is equivalent to `0xabcdef`. Checksum is generally for avoiding typos etc. when taking an irreversible action like sending a tx. I don't think it applies to our case.

test/server.js
kuzdogan: I agree, also better to keep it consistent with the internal data structure.

services/verification/src/services/Injector.ts
marcocastignoli: In metamask they have two utils: with0x(), without0x()

services/verification/src/services/Injector.ts
marcocastignoli: I know that one usecase of create2 is to have fancy addresses, so maybe the user will search for a specific address like 0xAaAaAa and she thinks she found it but then she has problems for this check, I don't know if I explained myself, kinda a specific edge case I know :D

services/verification/package.json
```diff
@@ -25,13 +25,20 @@
     "@types/node-fetch": "^2.5.7",
     "@types/promise.any": "^2.0.0",
     "@types/semver": "^7.3.9",
+    "lru-cache": "^7.14.1",
```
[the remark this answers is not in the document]
marcocastignoli: why there is that error?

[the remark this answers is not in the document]
kuzdogan: ethereum-js has an older dependency which does not have type declarations.

kuzdogan:
Updated, accordingly. Please have a look @marcocastignoli

marcocastignoli (approved):

PR #877, opened 2022-12-16
refactor: `getFunction` is simpler

therealharpaljadeja:
Get function from ABI directly using `functionSelector`

image


therealharpaljadeja:
sorry I thought it was getting the abi, closing.

PR #886, opened 2023-01-03
Update Celo Mainnet Endpoint

0xCoDeFi:
Updating Celo Mainnet endpoint only


kuzdogan:
Thanks! While looking into it I also noticed a bug https://github.com/ethereum/sourcify/commit/d91fe634081b33e32b5314c83e5a59d6dcc89a37 

Merging

Issue #897, opened 2023-01-18
Add ENS to the landing page example verified contracts


kuzdogan:
AFAIK there are ENS contracts verified on Sourcify. Find the main contract verified and add it to the featured contracts in the UI `featured.js`. 

ENS Contracts: https://etherscan.io/accounts/label/ens

yash251:
This is the base contract of ENS - 
https://etherscan.io/address/0x57f1887a8bf19b14fc0df6fd9b2acc9af147ea85

PR #899, opened 2023-01-20
added ens to featured contracts

yash251:
solves #897


marcocastignoli:
image

thanks for the contribution! Personally I would change a bit how this is rendered because it seems like it has a big margin compared to the others. Or maybe we can rename it ENS insteam of Ethereum Naming System. I believe that for the target of Sourcify ENS in explentarory enough.

yash251:
right! that is because of the large name. I'll change it to ENS.

marcocastignoli:
thanks @yash251 for the contribution!

PR #911, opened 2023-02-08
Update chains

BoThe1K:
Fetch new file from: https://chainid.network/chains.json
via: `npm run update-chains`


marcocastignoli:
I'll do it, thanks for PR but it's simpler for me to do it myself instead of checking :D

PR #915, opened 2023-02-13
add: canto rpc to chain list

jessepinkman9900:


marcocastignoli:
we use the RPC from chains.json, you have to update it here: https://chainid.network

PR #919, opened 2023-02-14
update chains.json

0xSilkweave:
Added two additional Canto (mainnet) RPCs.


marcocastignoli:
we update the chains.json file from https://chainid.network/. Open a PR over there if not yet updated.

PR #920, opened 2023-02-14
Added Elastos chain

irzhywau:


marcocastignoli:
we update the file chains.json regularly from https://github.com/ethereum-lists/chains. open a PR there if you need to update the network

PR #921, opened 2023-02-15
Update chains.json add Bear Network Chain Mainnet 641230

BearNetwork-BRNKC:


marcocastignoli:
this is not needed, we update the file automatically from https://chainid.network/
also, Bear Network is already included

PR #939, opened 2023-02-24
update: explorer endpoint

jessepinkman9900:


kuzdogan:
The immutable test was failing and it seems it is because `https://canto.slingshot.finance` was returning empty response when the tx is requested:

```
curl --location 'https://canto.slingshot.finance/' --header 'Content-Type: application/json' --data '{
	"jsonrpc":"2.0",
	"method":"eth_getTransactionByHash",
	"params":[
		"0x118e78686df742d225a28caf8852ec29269d01effa46c799ed02ff3051dbcc33"
	],
	"id":1
}'
```
returns:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": null
}
```

The second RPC `https://canto.neobase.one/` returns successfully. You might want to look into why canto.slingshot.finance is behaving this way.

This had to be handled by our server but it seems there was a bug. Now we use the successful response from the second RPC. Can you please rebase?

jessepinkman9900:
@kuzdogan rebase completed. please review

PR #942, opened 2023-03-02
User input creator tx

kuzdogan:
Changes enable passing the `creatorTxHash` variable to server. Normally this is scraped from an explorer or fetched with an API but sometimes this fails. This allows users to provide it themselves.

- Adds the variable to both session and legacy verify endpoints
- Adds tests for both
- Additionally refactors how files are saved to remove the redundant code.

TODO:
- [ ] Add the field to the docs once the PR passes.


src/server/services/RepositoryService.ts
```diff
@@ -527,93 +544,40 @@ export default class RepositoryService implements IRepositoryService {
...
-        fileName: "create2-args.json",
+        fileName,
       },
-      JSON.stringify(create2Args)
+      JSON.stringify(contentJSON)
     );
   }
 
-  private storeLibraryMap(
+  private storeTxt(
```
marcocastignoli, line 565: I like this and storeJson

src/server/services/RepositoryService.ts
```diff
@@ -527,93 +544,40 @@ export default class RepositoryService implements IRepositoryService {
...
-      {
-        matchQuality,
-        chainId,
-        address,
-        source: false,
-        fileName: "create2-args.json",
+        fileName,
       },
-      JSON.stringify(create2Args)
+      JSON.stringify(contentJSON)
```
marcocastignoli, line 561: I just noticed that `JSON.stringify(contextVariables, undefined, 2)` is not used. I'm personally ok with that, but there is less readability for a human manually reading the repository. If you also think this is not a problem no need to change it.

marcocastignoli (approved):

src/server/services/RepositoryService.ts
kuzdogan: Yes I haven't added it here and removed it in other places to keep consistency https://github.com/ethereum/sourcify/pull/942/commits/d1034319fdc5fa8fa76e25136e77719a93ad06f7 the metadata.json also does not have spaces. Apart from metadata, the JSON files are not big and would be still readable.

Issue #952, opened 2023-03-09
File not saved after CREATE2 verification


kuzdogan:
When verifying `0x000000000022D473030F116dDEE9F6B43aC78BA3` on mainnet the UI shows a successful verification but upon looking for the contract in Lookup or in the repo link, it can be seen that it's empty.

### Steps to reproduce:
- Import `0x000000000022D473030F116dDEE9F6B43aC78BA3` from Etherscan Mainnet
- Click "Verify create2 contract"
- Add deployer address `0x4e59b44847b379578588920ca78fbf26c0b4956c` and salt `0xd3af2663da51c10215000000`, enter `CLIENT_TOKEN`.
- Click verify
- Enter `0x000000000022D473030F116dDEE9F6B43aC78BA3` in https://staging.sourcify.dev/#/lookup
  - Must be empty
- Go to https://repo.staging.sourcify.dev/contracts/full_match/0/0x000000000022D473030F116dDEE9F6B43aC78BA3/
  - The folder will be empty

PR #957, opened 2023-03-09
#952 File not saved after CREATE2 verification

marcocastignoli:
* the store function was not present in the create2 verification
* add tests to check the existence of the contract in create2 tests


kuzdogan (approved):
LGTM!

PR #959, opened 2023-03-09
Add other Etherscan chains

### LINKED ISSUE #940: Add other chains for import from Etherscan
There are other chains that use Etherscan and that would enable us to use the import feature.

I've added the ones I'm aware of while debugging a contract with the import feature. Added are:
- Arbitrum mainnet & testnet
- Optimism mainnet & testnet
- Avalanche mainnet & testnet

Are there other chains with Etherscan?

The code is pushed to the branch [etherscan-all-chains](https://github.com/ethereum/sourcify/tree/etherscan-all-chains).

Additionally we need to check if we need to add API keys for the respective chains. The API accepts requests without API keys (or with wrong API key?) with a restricted rate limit. Check if the key is being used for each chain, including Ethereum chains. Maybe there's an API-key-only endpoint that allows easily checking this.

kuzdogan:
Fixes #940 that adds other Etherscan instances (Polygonscan etc.) to Import to Etherscan feature.

Also supersedes https://github.com/ethereum/sourcify/pull/966 as it's related. 

This ended up being a larger PR than expected because of cascading issues. In general most of the changes are straightforward and can be hidden by marking "Viewed" during the review. 

- Adds "Etherscan instance tests" with 3 contract types for each Etherscan instance to be run nightly.
- Separate the "Etherscan tests" from server.js.
- Reverts back to the external IPFS Gateway in tests as the local gateway is no longer working.
- Refactors the tests to use shared "assertion..." functions
- Adds validation to the Etherscan endpoints
  - Both endpoints now both accept `chain` and `chainId`
- Adds port parameter to server constructor, to be able to run tests on different ports
- Adds `etherscanAPI` field to the `/chains` endpoint


packages/contract-call-decoder/src/lib/ContractCallDecoder.ts
```diff
@@ -35,7 +35,7 @@ type GetMetadataOptions = {
 const defaultGetMetadataOptions: GetMetadataOptions = {
   source: MetadataSources.Sourcify,
   sourcifyProvider: 'https://repo.sourcify.dev',
-  ipfsGateway: 'https://cloudflare-ipfs.com/',
+  ipfsGateway: 'https://ipfs.io/',
```
marcocastignoli, line 38: I remember I changed this a while ago because it was more stable in general

packages/lib-sourcify/src/lib/verification.ts
```diff
@@ -140,7 +140,6 @@ export async function verifyCreate2(
     address: computedAddr,
     chainId: '0',
     status: 'perfect',
-    storageTimestamp: new Date(),
```
marcocastignoli, line 143: Why is this getting removed?

src/server/controllers/VerificationController.ts
```diff
@@ -619,16 +619,48 @@ export default class VerificationController
...
+        .optional()
+        .custom(
+          (chainId, { req }) =>
+            // Support both `body.chain` and `body.chainId`
+            (req.body.chain = chainId)
+        ),
+      body("chain")
+        .exists()
+        .bail()
+        .custom((chain, { req }) => (req.chain = checkChainId(chain))),
```
marcocastignoli, line 638: `req.chain` or `req.body.chain`?

src/server/controllers/VerificationController.ts
```diff
@@ -619,16 +619,48 @@ export default class VerificationController
...
+        .bail()
+        .custom(
+          (address, { req }) => (req.addresses = validateAddresses(address))
+        ),
+      body("chainId")
+        .optional()
+        .custom(
+          (chainId, { req }) =>
+            // Support both `body.chain` and `body.chainId`
+            (req.body.chain = chainId)
```
marcocastignoli, line 633: why here why are not calling `checkChainId`?

test/etherscan.js
```diff
@@ -0,0 +1,369 @@
...
+          assertEtherscanError(
+            err,
+            res,
+            "This contract is not verified on Etherscan"
+          );
+          done();
+        });
+    });
+
+    it("should fail by exceeding rate limit on etherscan APIs", async () => {
```
marcocastignoli, line 140: I would do it at the end just to be safe that then the other test succeed (I had the same problem before)

test/etherscan.js
```diff
@@ -0,0 +1,369 @@
...
```
marcocastignoli, line 292: Also this at the end

marcocastignoli (changes requested):
Looks good, I added a few comments, maybe some of them are not real problems.

src/server/controllers/VerificationController.ts
kuzdogan: It's already at `req.body.chain` but should've not assigned to `req.chain` like in https://github.com/ethereum/sourcify/blob/d8e7ef2c7735a9d4cee1121243e17b01e2e24979/src/server/controllers/VerificationController.ts#L663

src/server/controllers/VerificationController.ts
kuzdogan: It's effectively checked at the next check `req.body.chain`. If double checked, the error sends double errors

marcocastignoli (approved):

Issue #955, opened 2023-03-09
Set an environment variable for solc compiler platform


kuzdogan:
Right now, when compiling contracts we are trying to use the solc compiler binaries, which are set as the `linux-amd64` binaries. While it works for the servers it doesn't work in Apple M-chips (from my experience) and possibly on other platforms such as Windows. 

An environment variable should be set to let users choose which solc platform they want to use as in https://github.com/ethereum/solc-bin. 

Question: Should it contain multiple platforms @marcocastignoli ? I think it doesn't make sense as you likely choose only one so not, or would there be a case? I don't know much about the emscripten
```
SOLC_PLATFORMS=linux-amd64,emscripten-wasm32
```
or
```
SOLC_PLATFORM=linux-amd64
```

It should always fall back to `solc-js`, which means when the var is empty, it would always use `solc-js`.

This also requires us to change the current `SOLC_REPO` variable to the parent directory and save the downloaded binaries in respective folders:

Currently:
```
SOLC_REPO=/home/data/solc-bin/linux-amd64
```

After
```
SOLC_REPO=/home/data/solc-bin
```
and save the binary in `/home/data/solc-bin/linux-amd64/solc-linux-amd64-v0.4.12+commit.194ff033
```

This should be followed after this is resolved:
https://github.com/ethereum/sourcify/issues/954

kuzdogan:
The final verdict is:
- Check the platform from node with `os.platform()` or similar, and choose the relevant binary. So that we don't have to introduce a new env var.
  - As a note in the future for compatibility: What does it return in browser?
- If it's not available use solc.js

marcocastignoli:
> As a note in the future for compatibility: What does it return in browser?

it will depend on the bundler we'll use because process / os is not a global variable in the browser

marcocastignoli:
we decided to auto detect the environment

PR #964, opened 2023-03-14
Default SolcJs compiler

### LINKED ISSUE #954: Save solc-js executables after download
Currently when the compiler falls back to `solc-js` instead of binaries, it always downloads them and doesn't save. 

Save the solc-js files to avoid downloading each time. This means the solc-js handler first checks if the file exists and downloads otherwise.

marcocastignoli:
- [x] fix #954 force solcjs to use the setupMethods api
  * download the solcjs binary like the amd one
  * use the downloaded binary through setupMethods
  * check existing binaries before downloading new one
- [x] implement #955

### DIFF
--- a/packages/lib-sourcify/src/lib/solidityCompiler.ts
+++ b/packages/lib-sourcify/src/lib/solidityCompiler.ts
@@ -10,6 +10,8 @@ const solc = require('solc');
 
 const GITHUB_SOLC_REPO =
   'https://github.com/ethereum/solc-bin/raw/gh-pages/linux-amd64/';
+const GITHUB_SOLCJS_REPO =
+  'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/';
 const RECOMPILATION_ERR_MSG =
   'Recompilation error (probably caused by invalid metadata)';
 
@@ -58,7 +60,9 @@ export async function useCompiler(version: string, solcJsonInput: JsonInput) {
     compiled = shellOutputBuffer.stdout.toString();
   } else {
     const soljson = await getSolcJs(version);
-    compiled = soljson.compile(inputStringified);
+    if (soljson) {
+      compiled = soljson.compile(inputStringified);
+    }
   }
 
   if (!compiled) {
@@ -142,6 +146,51 @@ async function fetchSolcFromGitHub(
   return false;
 }
 
+async function validateSolcJsPath(soljsonPath: string): Promise<boolean> {
+  try {
+    const solcjsImports = await import(soljsonPath);
+    const soljson = solc.setupMethods(solcjsImports);
+    if (soljson) {
+      return true;
+    }
+    return false;
+  } catch (e) {
+    return false;
+  }
+}
+
+async function fetchSolcJsFromGitHub(
+  solcJsPath: string,
+  version: string,
+  fileName: string
+): Promise<boolean> {
+  const githubSolcURI = GITHUB_SOLCJS_REPO + encodeURIComponent(fileName);
+  const res = await fetchWithTimeout(githubSolcURI);
+  // TODO: Handle nodejs only dependencies
+  if (res.status === StatusCodes.OK) {
+    fs.mkdirSync(path.dirname(solcJsPath), { recursive: true });
+    const buffer = await res.arrayBuffer();
+
+    try {
+      fs.unlinkSync(solcJsPath);
+    } catch (_e) {
+      undefined;
+    }
+    fs.writeFileSync(solcJsPath, new DataView(buffer), { mode: 0o755 });
+    if (await validateSolcJsPath(solcJsPath)) {
+      console.log(
+        `Successfully fetched solcjs ${version} from GitHub: ${githubSolcURI}`
+      );
+      return true;
+    }
+  }
+  console.log(
+    `Failed fetching solcjs ${version} from GitHub: ${githubSolcURI}`
+  );
+
+  return false;
+}
+
 /**
  * Fetches the requested version of the Solidity compiler (soljson).
  * First attempts to search locally; if that fails, falls back to downloading it.
@@ -158,34 +207,25 @@ async function fetchSolcFromGitHub(
  *
  * @returns the requested solc instance
  */
-export function getSolcJs(version = 'latest'): Promise<any> {
+export async function getSolcJs(version = 'latest'): Promise<any> {
   // /^\d+\.\d+\.\d+\+commit\.[a-f0-9]{8}$/
   version = version.trim();
   if (version !== 'latest' && !version.startsWith('v')) {
     version = 'v' + version;
   }
 
-  const soljsonRepo = process.env.SOLJSON_REPO || 'soljson-repo';
-  const soljsonPath = path.resolve(soljsonRepo, `soljson-${version}.js`);
+  const soljsonRepo =
+    process.env.SOLJSON_REPO || path.join('/tmp', 'soljson-repo');
+  const fileName = `soljson-${version}.js`;
+  const soljsonPath = path.resolve(soljsonRepo, fileName);
 
-  if (fs.existsSync(soljsonPath)) {
-    console.log(`Using local solcjs ${version} from ${soljsonPath}`);
-    return new Promise((resolve) => {
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
-      const soljson = solc.setupMethods(require(soljsonPath));
-      resolve(soljson);
-    });
+  if (!fs.existsSync(soljsonPath)) {
+    if (!(await fetchSolcJsFromGitHub(soljsonPath, version, fileName))) {
+      return false;
+    }
   }
 
-  return new Promise((resolve, reject) => {
-    solc.loadRemoteVersion(version, (error: Error, soljson: any) => {
-      if (error) {
-        console.error(error);
-        reject(error);
-      } else {
-        console.log(`Got solcjs ${version}`);
-        resolve(soljson);
-      }
-    });
-  });
+  console.log(`Using local solcjs ${version} from ${soljsonPath}`);
+  const solcjsImports = await import(soljsonPath);
+  return solc.setupMethods(solcjsImports);
 }
--- a/packages/lib-sourcify/src/lib/solidityCompiler.ts
+++ b/packages/lib-sourcify/src/lib/solidityCompiler.ts
@@ -8,13 +8,24 @@ import { JsonInput } from './types';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const solc = require('solc');
 
-const GITHUB_SOLC_REPO =
-  'https://github.com/ethereum/solc-bin/raw/gh-pages/linux-amd64/';
+const GITHUB_SOLC_REPO = 'https://github.com/ethereum/solc-bin/raw/gh-pages/';
 const GITHUB_SOLCJS_REPO =
-  'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/';
+  'https://github.com/ethereum/solc-bin/raw/gh-pages/bin/';
 const RECOMPILATION_ERR_MSG =
   'Recompilation error (probably caused by invalid metadata)';
 
+export function findSolcPlatform(): string | false {
+  if (process.platform === 'darwin' && process.arch === 'x64') {
+    return 'macosx-amd64';
+  }
+  if (process.platform === 'linux' && process.arch === 'x64') {
+    return 'linux-amd64';
+  }
+  if (process.platform === 'win32' && process.arch === 'x64') {
+    return 'windows-amd64';
+  }
+  return false;
+}
 /**
  * Searches for a solc: first for a local executable version, then from GitHub
  * and then using the getSolcJs function.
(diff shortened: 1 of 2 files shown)


marcocastignoli:
Before merging we need to test the new functionalities on win,mac,linux,fallback

- [x] Fallback
- [ ] Linux
- [ ] Windows
- [ ] Mac (intel)

packages/lib-sourcify/src/lib/solidityCompiler.ts
```diff
@@ -140,6 +170,51 @@ async function fetchSolcFromGitHub(
...
+    if (soljson) {
+      return true;
+    }
+    return false;
+  } catch (e) {
+    return false;
+  }
+}
+
+async function fetchSolcJsFromGitHub(
```
kuzdogan, line 186: Do we really need a whole new function for this? It seems to be overlapping a lot with `fetchSolcFromGitHub`

Also can we rename the both functions `fetchAndSave...FromGitHub`. While reviewing it was not clear to me if this returns the executable or saves the file to the disk.

kuzdogan (commented):
Looks good! 

We also need to test these. 

1. Add a unit test for `useCompiler` with a mock json-input.

If it's possible to mock/override `os.process()`

2. A unit test with an invalid `os.process()` for `useCompiler` to see if the fallback to `solcjs` works. 

If it's not possible to mock the `os.process()` we run the `solcjs` part of the `useCompiler` code manually and test

packages/lib-sourcify/src/lib/solidityCompiler.ts
marcocastignoli: I thought about it but the check to make sure that the compiler is working is different between the solcjs and the others. Maybe we can just group the download part. Or pass as a callback the check of the working compiler.

kuzdogan: Do we need to validate it here? The reason we were validating was to check if the binary is executable,  i.e. if you have the Linux solc on a Mac, it will fall back to `solc-js`. I can't think of any case we'd need to "validate" the `solc-js`

packages/lib-sourcify/src/lib/solidityCompiler.ts
```diff
@@ -156,34 +231,25 @@ async function fetchSolcFromGitHub(
...
-    console.log(`Using local solcjs ${version} from ${soljsonPath}`);
-    return new Promise((resolve) => {
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
-      const soljson = solc.setupMethods(require(soljsonPath));
-      resolve(soljson);
-    });
+  if (!fs.existsSync(soljsonPath)) {
+    if (!(await fetchSolcJsFromGitHub(soljsonPath, version, fileName))) {
+      return false;
+    }
```
kuzdogan, line 249: If validation is useful in `solc-js` it would not serve the purpose here.

Think of the case: 
1. solcjs-v0.8.0 does not exist, so we fetch **and save** it
2. The validation in fetch function fails, so we return false
3. Another contract using v0.8.0 asks for the compiler
4. It exists on the disk so this time it won't return false

packages/lib-sourcify/src/lib/solidityCompiler.ts
```diff
@@ -140,6 +170,51 @@ async function fetchSolcFromGitHub(
...
+  // TODO: Handle nodejs only dependencies
+  if (res.status === StatusCodes.OK) {
+    fs.mkdirSync(path.dirname(solcJsPath), { recursive: true });
+    const buffer = await res.arrayBuffer();
+
+    try {
+      fs.unlinkSync(solcJsPath);
+    } catch (_e) {
+      undefined;
+    }
```
kuzdogan, line 202: Any ideas why we `unlink()`? I see this was also in the old `fetcSolc...` code

packages/lib-sourcify/src/lib/solidityCompiler.ts
```diff
@@ -29,40 +42,47 @@ export async function useCompiler(version: string, solcJsonInput: JsonInput) {
...
-    let error: false | Error = false;
-    if (shellOutputBuffer.error) {
-      const typedError: NodeJS.ErrnoException = shellOutputBuffer.error;
-      // Handle compilation output size > stdout buffer
-      if (typedError.code === 'ENOBUFS') {
-        error = new Error('Compilation output size too large');
+  const solcPlatform = findSolcPlatform();
+  if (solcPlatform) {
+    const solcPath = await getSolcExecutable(solcPlatform, version);
+    if (solcPath) {
```
kuzdogan, line 50: If `solcPath` is assigned `false` here, because of the validation, it would silently fail and not fall back. I suggest:

```ts
  const solcPlatform = findSolcPlatform();
  let solcPath;
  if (solcPlatform) {
    solcPath = await getSolcExecutable(solcPlatform, version);
  }
  if (solcPath) {
    ...
  } else {
    const soljson = await getSolcJs(version);
    if (soljson) {
      compiled = soljson.compile(inputStringified);
    }
  }
```

packages/lib-sourcify/src/lib/solidityCompiler.ts
```diff
@@ -29,40 +42,47 @@ export async function useCompiler(version: string, solcJsonInput: JsonInput) {
...
-    compiled = soljson.compile(inputStringified);
+    if (soljson) {
+      compiled = soljson.compile(inputStringified);
+    }
   }
 
   if (!compiled) {
-    throw new Error('Compilation failed. No output from the compiler.');
+    throw new Error(
+      'Compilation failed. No output from the compiler. Try using a different compiler.'
```
kuzdogan, line 84:
```suggestion
      'Compilation failed. No output from the compiler.'
```

You can't compile a contract with a different _version_ of the compiler.

If we mean different _platform_ of the compiler, it wouldn't make sense because we're not giving the option to the user to choose. I suggest removing the last sentence

packages/lib-sourcify/src/lib/solidityCompiler.ts
kuzdogan: Anyways, my suggestion would be:
- Have a shared function to `fetchAndSave`
- Don't validate inside but validate in `getSolc` after `fetchAndSave`
- Don't validate `solc-js`, unless we have a reason to do so. If so, validate again in `getSolcJs`.

kuzdogan (changes requested):
Some more comments. Please read through all of them, I think they make sense as whole.

Also, a rebase is needed

packages/lib-sourcify/src/lib/solidityCompiler.ts
marcocastignoli: Yeah I left it by mistake from the previous implementation

packages/lib-sourcify/src/lib/solidityCompiler.ts
marcocastignoli: Honestly I didn't check how this work, I'll spend some time understanding how this function works.

marcocastignoli: Maybe because it fails to write if the file is already there?

packages/lib-sourcify/src/lib/solidityCompiler.ts
```diff
@@ -158,34 +186,25 @@ async function fetchSolcFromGitHub(
...
+  const soljsonRepo =
+    process.env.SOLJSON_REPO || path.join('/tmp', 'soljson-repo');
+  const fileName = `soljson-${version}.js`;
+  const soljsonPath = path.resolve(soljsonRepo, fileName);
 
-  if (fs.existsSync(soljsonPath)) {
+  if (!fs.existsSync(soljsonPath)) {
+    if (!(await fetchAndSaveSolc('bin', soljsonPath, version, fileName))) {
+      return false;
+    }
```
kuzdogan, line 204: <Repeating the comment>

If validation is useful in `solc-js` it would not serve the purpose here.

Think of the case: 
1. solcjs-v0.8.0 does not exist, so we fetch **and save** it
2. The validation in fetch function fails, so we return false
3. Another contract using v0.8.0 asks for the compiler
4. It exists on the disk so this time it won't return false

packages/lib-sourcify/src/lib/solidityCompiler.ts
```diff
@@ -29,36 +40,44 @@ export async function useCompiler(version: string, solcJsonInput: JsonInput) {
...
-    });
-
-    // Handle errors.
-    let error: false | Error = false;
-    if (shellOutputBuffer.error) {
-      const typedError: NodeJS.ErrnoException = shellOutputBuffer.error;
-      // Handle compilation output size > stdout buffer
-      if (typedError.code === 'ENOBUFS') {
-        error = new Error('Compilation output size too large');
+    if (solcPath) {
```
kuzdogan, line 51: Typo?

packages/lib-sourcify/src/lib/solidityCompiler.ts
```diff
@@ -109,12 +137,14 @@ function validateSolcPath(solcPath: string): boolean {
   return false;
 }
```
kuzdogan, line 139:
```suggestion

/**
 * Fetches a solc binary from GitHub and saves it to the given path.
 * 
 * If platform is "bin", it will download the solc-js binary.
 */
```

packages/lib-sourcify/test/functions.spec.ts
```diff
@@ -24,7 +24,7 @@ describe('Verify Solidity Compiler', () => {
     await getSolcJs('0.8.17+commit.8df45f5f');
   });
   it('Should fetch latest solc from github', async () => {
-    await getSolcExecutable('0.8.9+commit.e5eed63a');
+    await getSolcExecutable('linux-amd64', '0.8.9+commit.e5eed63a');
```
kuzdogan, line 27: Please check the returned result. Otherwise it will fail silently

kuzdogan (changes requested):
Please have a look at the comments. I noticed when we rebase, the previous comments get outdated and seem resolved so maybe we should use rebase in PR reviews more sparingly. 

Also repeating the comment from the first review:

> We also need to test these. 
> 
> 1. Add a unit test for `useCompiler` with a mock json-input.
> 
> If it's possible to mock/override `os.process()`
> 
> 2. A unit test with an invalid `os.process()` for `useCompiler` to see if the fallback to `solcjs` works. 
> 
> If it's not possible to mock the `os.process()` we run the `solcjs` part of the `useCompiler` code manually and test

packages/lib-sourcify/src/lib/solidityCompiler.ts
kuzdogan, line 51: Typo?

marcocastignoli: yes

packages/lib-sourcify/test/functions.spec.ts
```diff
@@ -44,10 +47,41 @@ describe('Verify Solidity Compiler', () => {
...
+          outputSelection: {
+            '*': {
+              '*': ['*'],
+            },
+          },
+        },
+      });
+      expect(compiledJSON?.contracts?.['test.sol']?.C).to.not.equals(undefined);
+    } catch (e: any) {
+      expect.fail('Compiler error:');
```
kuzdogan, line 78: Do we expect it to fail? Doesn't it pass the tests with this?

packages/lib-sourcify/test/functions.spec.ts
```diff
@@ -18,17 +18,20 @@ import { Metadata, MissingSources } from '../src/lib/types';
...
-    await getSolcExecutable('linux-amd64', '0.8.9+commit.e5eed63a');
+    expect(await getSolcJs('0.8.17+commit.8df45f5f')).not.equals(false);
   });
+  if (process.platform === 'linux') {
+    it('Should fetch latest solc from github', async () => {
+      expect(
+        await getSolcExecutable('linux-amd64', '0.8.9+commit.e5eed63a')
+      ).not.equals(null);
+    });
+  }
```
kuzdogan, line 32: Let's complete it with the Mac and Windows tests

kuzdogan (changes requested):
Two more comments.

Plus this test is missing?

> 1. Add a unit test for useCompiler with a mock json-input.

i.e. a passing test with "Should compile with solc binary"

packages/lib-sourcify/test/functions.spec.ts
marcocastignoli: at the end the tests are needed for circleci right? are macx64 and winx64 needed?

packages/lib-sourcify/test/functions.spec.ts
marcocastignoli: typo again

marcocastignoli: no sorry, with `expect.fail()` you are forcing the test to fail, and it's right

marcocastignoli: https://unexpected.js.org/api/fail/

kuzdogan: Ok let's add the error message there then. Because if it will be caught there won't be a way to see why it failed.

marcocastignoli: ok

packages/lib-sourcify/test/functions.spec.ts
kuzdogan: Mostly yes. This will be implicit in the missing test mentioned above anyway, so fine this way.

marcocastignoli: But that is already tested by other tests, that's why is missing here, if you check the coverage that branch is already covered like x15 times

marcocastignoli:
So if now we are ready we can merge

marcocastignoli:
seems test is passing

kuzdogan (approved):

PR #973, opened 2023-03-29
Verify with `immutableReferences`

### LINKED ISSUE #972: Using `immutableReferences` in verification
We haven't really been aware that the compiler outputs a field called `immutableReferences`, which shows where exactly the values of the immutable variables are stored in the runtime bytecode. By leaving out these fields we can basically verify any contract with immutables. 

This could eliminate the need for different types of verification like "simulation" etc. Additionally we would no longer need the extra information `msg.sender` etc. and saving those in a extra files in the repo. This would enable reproducing verification only with the information found in `metadata.json`.

As these fields will only be values it should be ok to verify without looking at these positions.

- Add `evm.deployedBytecode.immutableReferences` to the `outputSettings` in `CheckedContract`
- Return `immutableReferences` from `recompile()`
- Pass these optional parameters to `matchWithDeployedBytecode()`
- Take these fields into considerations when matching in the function.

kuzdogan:
Uses the `immutableReferences` output of the compiler to set the immutable values in the onchain bytecode to `0000...` before matching. Saves them in a file called `immutable-references.json`.

Fix #972 

- Lowers the branch threshold to 60%. I've added some tests to increase coverage but branch coverage is left as something for the future.
- Adds the CREATE2 test to `lib-sourcify` too
- Adds additional `lib-sourcify` tests.
- Removes the "simulation" and `creatorTxHash` tests but adds them as unit tests below.
- Add unit test for the `replaceImmutableReferences` function.
- Skips etherscan rate limit tests as they seem to be inconsistent. Left for the future.
- Remove constructor args checks from monitor tests as immutbles are no longer handled by verifyWithCreatorTx.


marcocastignoli (approved):
After looking at this, I think that the only case in which other verification methods are useful is when the compiler can't generate the ImmutableReferences object, but can it happen?

I'm in favor for keeping this as the only verification method.

kuzdogan:
> After looking at this, I think that the only case in which other verification methods are useful is when the compiler can't generate the ImmutableReferences object, but can it happen?
> 
> I'm in favor for keeping this as the only verification method.

I don't think that would ever happen. That'd be a compiler bug.

Ok, good. Still we should not remove it immediately but after making sure we won't need them

Issue #936, opened 2023-02-23
Investigate what might be causing partial matches from single-file Etherscan contracts


kuzdogan:
Looking into verifying contracts from Etherscan, I notice **almost always** we can verify perfectly if the contracts are with standard-json input, which is great.

This leaves me wondering if we can find a pattern in single file or multi-part contracts that would enable perfect matches. Because opposingly, we **almost always** have partial matches with those.

My initial naive guess was the prepended "Submitted to Etherscan at..." part in the code but apparently that's just added on the UI and not in the API.

A way to find out this might be:

* Deploy a contract to a testnet, save the metadata before.
* Verify it as a single file contract on Etherscan
* Check if it's verified. If not compare the metadata generated by Sourcify during verification vs. the saved metadata of the deployed contract.

Example single file: https://etherscan.io/address/0x3446dd70b2d52a6bf4a5a192d9b0a161295ab7f9#code https://etherscan.io/address/0x4691937a7508860f876c9c0a2a617e7d9e945d4b#code

<sub>View in Huly <a href="https://app.huly.io/guest/w-kaanuzdogan-sourcify-65d30de0-75121c5c6c-fd615e?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NWU1N2U0NzhiNmVjZDA0ZDYyZDI0ZTQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6Incta2FhbnV6ZG9nYW4tc291cmNpZnktNjVkMzBkZTAtNzUxMjFjNWM2Yy1mZDYxNWUiLCJwcm9kdWN0SWQiOiIifQ.Zo4f34jBRmQaYt_kuAkpxlIscyZ1P1s5zcTJJ-Zeb1g">HI-391</a></sub>

sealer3:
I found the solution. Etherscan was replacing my Unix newlines (`\n`) with Windows newlines (`\r\n`). I was able to get the correct source keccak256 hash from the Etherscan output by removing the carriage returns (`\r`).

marcocastignoli:
@kuzdogan if I remember correctly Sourcify already has a function to test the keccak256 changing whitespaces and trailing newlines?

sealer3:
[...]

PR #976, opened 2023-04-05
Etherscan metadata variations

kuzdogan:


marcocastignoli:
@kuzdogan PR is not ready but if you want to check what I did till now and give me some feedback, it's appreciated.

packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -20,10 +25,10 @@ const FETCH_TIMEOUT = parseInt(process.env.FETCH_TIMEOUT || '') || 3000; // ms
  */
 export class CheckedContract {
   /** Object containing contract metadata keys and values. */
-  metadata: Metadata;
+  metadata!: Metadata;
```
marcocastignoli, line 28: In TypeScript, the exclamation mark ! after a property name is called a non-null assertion operator. It is used to tell the TypeScript compiler that a property or variable is not null or undefined, even though the compiler might think otherwise.

Here's an example:

```typescript
class Person {
  name!: string;
  age!: number;
}
```

In this example, the ! after name and age tells the compiler that these properties will be initialized at some point, even though they are not initialized in the constructor or anywhere else in the class. This is useful when dealing with situations where the TypeScript compiler can't infer whether a value is null or undefined, such as when the value is set asynchronously.

Note that using the non-null assertion operator should be done with care, as it can lead to runtime errors if the value is actually null or undefined. It is always better to check if a value is null or undefined before using it, instead of relying on the non-null assertion operator.

test/etherscan.js
```diff
@@ -31,6 +31,10 @@ chai.use(chaiHttp);
 const CUSTOM_PORT = 5678;
 
 describe("Import From Etherscan and Verify", function () {
+  this.beforeEach(async function () {
```
marcocastignoli, line 34: I added this, because lots of Etherscan tests were failing

packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -91,6 +89,81 @@ export class CheckedContract {
...
+      solidity = sources.reduce((sources, source) => {
+        sources[source.path] = source.content;
+        return sources;
+      }, {});
+
+      const compilationResult = await useCompiler(
+        this.compilerVersion,
+        solcJsonInput
+      );
+
```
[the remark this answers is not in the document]
marcocastignoli: I completally missed this while reading your implementation idea.

test/etherscan.js
kuzdogan: I don't think this is a reason for the tests to fail. The rate limit is quite generous. Did you look why it was failing?

marcocastignoli: If some etherscan tests fails then they end sooner than expected, hitting Etherscan's API rate limit

packages/lib-sourcify/src/lib/CheckedContract.ts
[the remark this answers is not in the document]
marcocastignoli: Implemented with the last commit

packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -91,6 +89,81 @@ export class CheckedContract {
...
+        };
+      }
+    );
+
+    const byHash = storeByHash(pathContent);
+
+    const byVariation = groupBy(
+      Array.from(byHash, ([, value]) => value),
+      'variation'
+    );
```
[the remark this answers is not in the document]
marcocastignoli: added comments

marcocastignoli:
Tests are failing because https://ipfs.io is down

marcocastignoli:
@kuzdogan now I think we are ready for merge, wanna review again?

packages/lib-sourcify/src/lib/verification.ts
```diff
@@ -59,7 +59,26 @@ export async function verifyDeployed(
...
+
+    matchWithDeployedBytecode(
+      originalMetadataMatch,
+      recompiled.deployedBytecode,
+      deployedBytecode
+    );
+    if (originalMetadataMatch.status === 'perfect') {
+      return originalMetadataMatch;
+    } else {
+      return match;
```
kuzdogan, line 79: Here we return the previous `match` but we already modified the `checkedContract` in the `tryToFind...`, so the wrong checkedContract will be saved in the repo. If it's not a perfect match, we should not modify the `checkedContract` at all. 

My suggestion would be to return `new CheckedContract` or `null` from the `tryToFind..`. So first check if `match.status === 'partial'`, then `const newCheckedContract = checkedContract.tryToFind`, recompile, try to match, and if perfect `checkedContract = newCheckedContract`

packages/lib-sourcify/test/functions.spec.ts
```diff
@@ -157,4 +159,12 @@ describe('Checked contract', () => {
     expect(sources).lengthOf(1);
     expect(sources[0]).equals('Storage.sol');
   });
+  it('Should tryToFindOriginalMetadata from checked contract', async () => {
+    const contract = new CheckedContract(WrongMetadata as any, {
```
kuzdogan, line 163: Should be `Metadata` not `any`?

kuzdogan, line 163: Doesn't it throw here? Should be Metadata not any?

packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -91,6 +89,134 @@ export class CheckedContract {
...
+                return '';
+              }
+            );
+          }
+        }
+        return sources;
+      }, {});
+
+      if (decodedAuxdata?.ipfs) {
+        const compiledMetadataIpfsCID = ipfsHash(JSON.stringify(metadata));
```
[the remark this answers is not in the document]
marcocastignoli: I think JSON.stringify is enough, no? doesn't it already remove whitespaces and order keys alphabetically?

packages/lib-sourcify/src/lib/hashFunctions/ipfsHash.ts
```diff
@@ -0,0 +1,179 @@
+// This is a typescript translation of the file https://github.com/ethereum/solidity/blob/develop/libsolutil/IpfsHash.cpp
```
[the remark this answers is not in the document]
marcocastignoli: Thanks ChatGPT ❤️

packages/lib-sourcify/src/lib/hashFunctions/swarmHash.ts
```diff
@@ -0,0 +1,68 @@
+import { makeChunkedFile } from '@fairdatasociety/bmt-js';
```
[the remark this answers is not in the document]
marcocastignoli: Here is was more a collab between me and ChatGpt 😆 I wrote swarmBzzr1Hash because ChatGpt couldn't create a working version of it while translating from c++. Everything was translated by ChatGPT.

packages/lib-sourcify/src/lib/verification.ts
marcocastignoli: I think here you misanderstood, or I'm not seeing something. In theory the `initSolcJsonInput` method is called only if we have a perfect match, so this.solidity, this.metadata, this.metadataRaw are not updated by `tryToFindOriginalMetadata`.

marcocastignoli:
> what do you mean? Don't we have verification test cases? Like the one that is our starting point from Etherscan?

packages/lib-sourcify/src/lib/CheckedContract.ts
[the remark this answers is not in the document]
kuzdogan: I does indeed remove the whitespaces but does not change the ordering

kuzdogan:
> > Don't we have verification test cases? Like the one that is our starting point from Etherscan?
> 
> what do you mean?

I see the function is just unit tested but the reason we develop this at the first place is to be able to verify contracts perfectly when they have common and known changes in the metadata that we can try to fix. I was referring to the contracts in this issue https://github.com/ethereum/sourcify/issues/936 We should test the verification on those, as well as find and add other cases that we know.

marcocastignoli:
The contract in the test is actually one of those, the one from this comment: https://github.com/ethereum/sourcify/issues/936#issuecomment-1490680020

packages/lib-sourcify/src/lib/types.ts
```diff
@@ -27,7 +27,14 @@ export interface MissingSources {
   };
 }
 
-// @TODO: Fully define metadata
+export interface MetadataSources {
+  [index: string]: {
+    keccak256: string;
+    content?: string;
+    urls?: string[];
```
kuzdogan, line 34: There's also `license?`

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -341,6 +341,25 @@ describe('lib-sourcify tests', () => {
...
+        accounts[0]
+      );
+
+      const match = await checkAndVerifyDeployed(
+        contractFolderPath,
+        sourcifyChainGanache,
+        deployedAddress
+      );
+      expectMatch(match, 'perfect', deployedAddress);
+    });
```
kuzdogan, line 362: Are you deploying the contract compiled with "Windows EOL(endofline)" and then verifying again with "Windows EOL" sources or is the "artifact.json" compiled with "Unix EOL"?

If former this is not how it works. If latter, we should explicitly comment somewhere that this bytecode ao nd metadata hash belongs to Unix EOL formatted source code.

Also I believe "SimplyLog.json" file is not being used, we can remove it.

Finally there should be a better way to name this case. Simply "wrong" doesn't tell much. In detail what we have here is a contract compiled and deployed with "Unix EOL" sources but during verification, the source files (from Etherscan i.e. without a metadata file) are provided with "Windows EOL".

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -341,7 +341,7 @@ describe('lib-sourcify tests', () => {
 
       expectMatch(match, 'perfect', childAddress);
     });
-    it('should verify a contract passing files with wrong newlines', async () => {
+    it('should verify a contract with LF eol sources passing sources with CRLF eol', async () => {
```
kuzdogan, line 344:
```suggestion
    it('should verify a contract which is originally compiled and deployed with Unix style End Of Line (EOL) source code, but being verified with Windows style (CRLF) EOL source code', async () => {
```

It's still a little vague to me. Can you write it like this? Sorry to be too intrusive, but let's lean on being more explicit and clear.

packages/lib-sourcify/src/lib/verification.ts
```diff
@@ -59,7 +59,31 @@ export async function verifyDeployed(
...
+        checkedContract.initSolcJsonInput(
+          checkedContractWithOriginalMetadata.metadata,
+          checkedContractWithOriginalMetadata.solidity
+        );
+        return matchWithOriginalMetadata;
+      } else {
+        return match;
+      }
+    }
+  }
```
kuzdogan, line 86: Another thing I'm noticing now, we are assuming the verification will work with `matchWithDeployedBytecode` only. 

We should check if the match status is partial after trying all matching methods. This means if a contract can't be matched with `matchWithDeployedBytecode` but `matchWithSimulation` we miss the chance to verify it perfectly. 
    - Although we think we are not going to need "simulation" and "creatorTx" verification methods, we still don't know if they are completely useless, especially "creatorTx".

If we choose to recurse by calling `verifyDeployed` again, we should be careful not to recurse infinitely

PR #977, opened 2023-04-05
Fix CORS error on req too large

kuzdogan:
Since the bodyParser, was set before the cors in server.ts, the response will have wrong CORS when body size limit is hit.
Move the cors settings above bodyParser.

Also add conditional CORS settings as session requires cookies and "*" CORS is not possible. Remove the individual cors() middleware on every session route in VerificationController.

TODO:
~~- [ ] Add informative message on the UI when HTTP status is `413`~~

### DIFF
--- a/src/server/controllers/VerificationController.ts
+++ b/src/server/controllers/VerificationController.ts
@@ -539,11 +539,6 @@ export default class VerificationController
   };
 
   registerRoutes = (): Router => {
-    const corsOpt = {
-      origin: config.corsAllowedOrigins,
-      credentials: true,
-    };
-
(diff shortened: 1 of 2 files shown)


kuzdogan:
I am unable to reproduce this to write the necessary update to the UI 😅 The UI already has a file size check and it shows a message 

https://github.com/ethereum/sourcify/blob/8bc26087d3bac131ac3f636d1105a7f43a357b5c/ui/src/pages/Verifier/index.tsx#L102-L111

image

Opening it for review

marcocastignoli (approved):
Everything looks good

PR #981, opened 2023-04-17
Change how addLibraryAddresses replaces the library addresses from the recompiled bytecode

marcocastignoli:
Instead of replacing the library using a regular expression, use split.

Thanks to @hellobloc and @samczsun for reporting the problem. I will add more information after we deploy.

### DIFF
--- a/packages/lib-sourcify/src/lib/verification.ts
+++ b/packages/lib-sourcify/src/lib/verification.ts
@@ -399,21 +399,21 @@ export function addLibraryAddresses(
   replaced: string;
   libraryMap: StringMap;
 } {
-  const PLACEHOLDER_START = '__$';
+  const PLACEHOLDER_START = '__';
   const PLACEHOLDER_LENGTH = 40;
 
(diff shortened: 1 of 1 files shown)


marcocastignoli:
This is the actual format:

`...35565b005b73__$a31769e258f964750f88ed70973ed7c298$__63874f33a160...`

Before using the keccak256 as the placeholder of the library the format was

`__{COMPLETE_PATH_FILE}:{CONTRACT_NAME}__`

But if "`{COMPLETE_PATH_FILE}:{CONTRACT_NAME}`" is longer than 40 bytes the string is stripped.

`...50565b600073__contracts/LibAAAAAAAAAAAAAAAAAAAAAAA__636d4ce63c60...`

The proposed solution is to not parse the placeholder in a RegExp, instead we can use a fixed length of 40 starting from `__`.

kuzdogan:
Can you add a test case for a pre-keccak256 placeholder version?

PR #986, opened 2023-04-19
Quick Release

marcocastignoli:
* remove verification through simulation


kuzdogan (approved):
Just a comment on comments :) Here we needed to put inline comments, but in general. let's use `//` line comments where possible. Makes it easier to see which lines are changed in PRs + easier to select and remove comments with hotkeys in the IDE

Issue #983, opened 2023-04-18
[repo] Case-insensitive address queries


volkyeth:
Currently the repo expects [Mixed-case checksum Addresses](https://eips.ethereum.org/EIPS/eip-55) on the queries, so if we use addresses with other cases it does not find the resource.

E.g: https://repo.sourcify.dev/contracts/full_match/100/0xC4c622862a8F548997699bE24EA4bc504e5cA865/metadata.json works https://repo.sourcify.dev/contracts/full_match/100/0xc4c622862a8f548997699be24ea4bc504e5ca865/metadata.json gives a 404

The dev experience could be improved by making the query work with addresses in any kind of case. As it stands, it's often necessary to add boilerplate to convert addresses to checksum mixed-casing in the code.

<sub>View in Huly <a href="https://app.huly.io/guest/w-kaanuzdogan-sourcify-65d30de0-75121c5c6c-fd615e?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NWU1N2U1MDhiNmVjZDA0ZDYyZDI1MDciLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6Incta2FhbnV6ZG9nYW4tc291cmNpZnktNjVkMzBkZTAtNzUxMjFjNWM2Yy1mZDYxNWUiLCJwcm9kdWN0SWQiOiIifQ._VZ4FVnH3xBSXy5D6AAd3_JZhEL-IcIVPCP-VLT9foE">HI-398</a></sub>

kuzdogan:
Thanks a lot for the report!

@marcocastignoli wasn't it already resolved with this? https://github.com/ethereum/sourcify/issues/588

marcocastignoli:
It used to work, I'll take a look at it. Thanks for the report

sealer3:
Related: if you pass a non-checksummed address in any of the verification tab inputs that accept addresses (Etherscan verification, contract/chain ID), it will give you a non-checksummed address "View in repository" link. I think the UI should always give a checksummed link and maybe checksum the address immediately upon entry.

Example:
![image](https://user-images.githubusercontent.com/125761775/233558881-de16fc6c-db5f-4fd7-b179-f8f8fdfe182a.png)
[...]

PR #990, opened 2023-04-21
Fix how the ui handles non-checksummed addresses

marcocastignoli:
See #936 for details.

Instead of verifying if an address is valid, tries to checksum it


kuzdogan (approved):

kuzdogan:
I guess you didn't mean to tag #936 but #983

marcocastignoli:
Uuups yes

Issue #996, opened 2023-04-24
Investigate: Library without immutables cannot be verified with `matchWithDeployedBytecode` because of placeholders in bytecode


kuzdogan:
I came across a contract (library) that could only be verified with the `creatorTxHash` as the recompiled deployed bytecode contains `0000000`s akin to immutables, but the library does not actually have immutables in the source code.

The contract is: https://testnet.snowtrace.io/address/0x28CA70cc5732D042381F9240bf45aB1b493aEFef#code

The part that differ is the bytes 1 to 21 in the bytecode, as below, which has the contract address `0x28CA70cc5732D042381F9240bf45aB1b493aEFef` value filled in during deployment.

Recompiled deployed bytecode:
```
[...]

PR #1000, opened 2023-04-26
Handle libraries with call protection

marcocastignoli:
See #996


marcocastignoli:
Since we discovered that this PR should handle all the cases, can we proceed? From my pov after writing a test we can merge, what do you think @kuzdogan?

kuzdogan:
Does this also cover the `via-IR` case? If so yes, let's add 
1. legacy compilation placeholder test case
2.`via-IR` test case 

and merge 👍

marcocastignoli:
> Does this also cover the via-IR case? If so yes, let's add

I suppose the actual function `replaceImmutableReferences` already does this.

marcocastignoli:
I added the tests

packages/lib-sourcify/src/lib/verification.ts
```diff
@@ -506,6 +513,20 @@ export function addLibraryAddresses(
...
+  const push20CodeOp = '73';
+  const callProtection = `0x${push20CodeOp}${'00'.repeat(20)}`;
+
+  if (template.startsWith(callProtection)) {
+    const replacedCallProtection = real.slice(0, 0 + callProtection.length);
+    return replacedCallProtection + template.substring(callProtection.length);
+  }
+  return template;
+}
+
```
kuzdogan, line 529: Doesn't this code only check the beginning of the bytecode with `.startsWith()`? 

How is the `viaIR` case passing? Does the `viaIR` case here have `0x73` at the beginning of the deployed bytecode? We should handle the case when the call protection is found in the middle of the bytecode.

kuzdogan (changes requested):

packages/lib-sourcify/src/lib/verification.ts
marcocastignoli: no, it handles it via the `immutableReferences`.

kuzdogan: Ahh right

kuzdogan (approved):

PR #1001, opened 2023-04-26
Implement a complete Metadata type

marcocastignoli:
* The Metadata type reflects the metadata object from https://docs.soliditylang.org/en/latest/metadata.html
* Used [`abitype`](https://github.com/wagmi-dev/abitype) for Abi related types

### DIFF
--- a/packages/lib-sourcify/src/lib/types.ts
+++ b/packages/lib-sourcify/src/lib/types.ts
@@ -1,3 +1,5 @@
+import { Abi } from 'abitype';
+
 export interface PathBuffer {
   path: string;
   buffer: Buffer;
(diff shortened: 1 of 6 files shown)


marcocastignoli:
@kuzdogan are any additional tests necessary for this PR?

src/server/controllers/VerificationController-util.ts
```diff
@@ -173,6 +174,10 @@ export type ContractWrapper = ContractMeta & {
   contract: CheckedContract;
 };
 
+type Mutable<Type> = {
+  -readonly [Key in keyof Type]: Type[Key];
+};
+
```
kuzdogan, line 180: What's this for?

packages/lib-sourcify/src/lib/types.ts
```diff
@@ -36,29 +38,99 @@ export interface MetadataSources {
...
+      [index: string]: {
+        details?: string;
+        params?: any;
+        returns?: any;
+      };
+    };
+    stateVariables?: any;
+    title?: string;
+    version?: number;
+  };
```
kuzdogan, line 68: Let's refactor this and `userdoc` out into exportable `Devdoc` and `Userdoc` types. I guess there are no NatSpec types being maintained, so we should maintain it ourselves and keep it reusable.

src/server/controllers/VerificationController-util.ts
marcocastignoli: This is used to remove the readonly keyword from the type, needed because we need to update that field in `constructorArgumentsArray`

kuzdogan (approved):

Issue #1002, opened 2023-04-26
Check if the Grafana-Loki is working


marcocastignoli:

marcocastignoli:
All the listeners were disabled after the refactoring

PR #1003, opened 2023-04-26
Re-enable SourcifyEventManager's listeners

marcocastignoli:
See #1002

In particular:
* `matchStored`: a listener for the `Verification.MatchStored` event
* `logger`: a generic listener that pushes all logs to loki


marcocastignoli:
I think that I need to implement the EventManager tests including these two listeners.

marcocastignoli:
Actually the existing tests already cover the Event Manager pretty well. The problem is related to `loggerLoki.ts`. But I think we can skip the coverage test for that branch: Loki is not enabled while running the tests on circleci. @kuzdogan what do you think?

Screenshot 2023-04-27 alle 15 47 17

kuzdogan:
Yes no need for tests there 👍

PR #1008, opened 2023-05-02
Add chain 296 - Hedera Testnet

svienot:
Adds support for chain 296 - Hedera testnet


svienot:
@kuzdogan we actually may have an issue with this one because our Hedera Testnet is reset every quarter (most data including contracts is wiped-out at each reset). Is there a way to completely wipe out all verification data for a given chain in the Sourcify repository? Without that ability we probably won't be able to integrate Testnet because there would be a risk that a newly created contract gets the address of a previously verified (but now deleted) contract...

svienot:
Closing this PR since the quarterly reset of Hedera testnet data may conflict with the contract verification data remaining permanently in  Sourcify repository.

PR #1022, opened 2023-05-09
Throw before if compiled bytecode empty instead of later in creation bytecode

kuzdogan:
I was looking at some other contracts and noticed that for abstract contracts the compiler returns empty bytecode.

I believe we should immediately throw if the user is trying to verify an abstract contract or for some reason, the compiler returns an empty bytecode.

### DIFF
--- a/packages/lib-sourcify/src/lib/verification.ts
+++ b/packages/lib-sourcify/src/lib/verification.ts
@@ -47,6 +47,15 @@ export async function verifyDeployed(
   };
   const recompiled = await checkedContract.recompile();
 
+  if (
+    recompiled.deployedBytecode === '0x' ||
+    recompiled.creationBytecode === '0x'
+  ) {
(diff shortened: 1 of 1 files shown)


marcocastignoli (approved):
from my pov is ok, but is better to thow or setting a null status to the match?

packages/lib-sourcify/src/lib/verification.ts
```diff
@@ -364,11 +373,6 @@ export async function matchWithCreationTx(
   creatorTxHash: string,
   recompiledMetadata: Metadata
 ) {
-  if (recompiledCreationBytecode === '0x') {
```
marcocastignoli, line 367: lets keep this

marcocastignoli (changes requested):

marcocastignoli (approved):

PR #1039, opened 2023-05-23
add-chain-2000 (dogechain)

oliviachef:


src/sourcify-chains.ts
```diff
@@ -790,6 +790,14 @@ const sourcifyChainsExtensions: SourcifyChainsExtensionsObject = {
       "https://explorer.plexfinance.us/" + BLOCKSCOUT_SUFFIX,
     txRegex: getBlockscoutRegex(),
   },
+  "2000": {
+    // DogeChain Mainnet
+    supported: true,
+    monitored: true,
```
marcocastignoli, line 796: Dogechain must not be monitored

```suggestion
    monitored: false,
```

marcocastignoli (changes requested):

PR #1049, opened 2023-06-04
Add Cronos Chain 25

crmax-it:


crmax-it:
After checking it can be removed @marcocastignoli

PR #1057, opened 2023-06-09
Update_Chain_1149

symplexialabs:
Due some internal problems, the Chain 1149 had to be reset. Thus, we had to deploy new contracts-tests.


kuzdogan:
For a short time there was a problem with lerna on our CI jobs. Can you please merge or rebase?

symplexialabs:
I give up. It's not worth the effort. It seems that everything is done only in order to complicate things. This is definitely not the way to produce winning solutions. Thanks.

kuzdogan:
Since this is not a new chain, the tests don't run on the CI but checked the chain test locally.

kuzdogan:
The RPCs have been unstable lately and the tests are failing. If not fixed we are going to have to remove support for this chain

manuelwedler:
@symplexialabs I'm disabling this chain as it has been constantly failing in our CI. Please reach out in case you still want to be supported by Sourcify. https://github.com/argotorg/sourcify/pull/2557

PR #1070, opened 2023-06-19
fix build - add optional title type

iainnash:
Currently the repo attemps to read the title type which should be an optional string on chain, but is omitted in the type definitions for chain in this repo.

This diff fixes that issue and can be confirmed with this link to the original schema:

https://github.com/ethereum-lists/chains/blob/master/tools/schema/chainSchema.json


kuzdogan:
Good catch! Thanks. However it seems we found this around the same time and I am seeing this just now, sorry. https://github.com/ethereum/sourcify/commit/243a40a2f0537e8ac9d84b4d2e6e87a99217d5e6

PR #1072, opened 2023-06-20
Update logging

kuzdogan:


src/common/loggerLoki.ts
```diff
@@ -23,16 +23,27 @@ if (process.env.GRAFANA_LOKI_URL) {
...
+    message: string;
+    timestamp?: string;
+    labels?: { event: string; level: string };
+  }) => {
+    return `${info.timestamp} [${info.level}]: ${
+      info.labels?.event
+        ? `[${info.labels?.event}] - ${JSON.stringify(info.message)}`
+        : info.message
+    }`;
+  }
```
kuzdogan, line 38: This will print a line instead of a JSON

marcocastignoli: ok

kuzdogan:
> Looks good to me, the only doubt that I have is about using directly the logger just for the debug. Couldn't we create a specific event for debugging? Just for the purpose of always passing through the EventManager?

I'm still not completely sure about using the EventManager. I tried for instance using the events in the monitor tests for example with `SourcifyEventManager.on("Monitor.Verified")` but couldn't make it work here. Maybe I missed something.
https://github.com/ethereum/sourcify/blob/302adaeee5c0cb6a6f45aa9eb13e9c04b7241fdc/test/monitor.js#L193-L197 

I don't have a full grasp of Events in node but felt like the modules themselves (monitor etc.) emitting events is more sensible.

Still we don't have to decide on this now and just get the job done for now.

PR #1082, opened 2023-06-28
Implement logging for LibSourcify

marcocastignoli:
* Created the simplest logging library inside LibSourcify (with no deps)
* change the default log level with `setLibSourcifyLoggerLevel`
* override the whole logging function with `setLibSourcifyLogger` implementing `ILibSourcifyLogger`

use the logger inside lib sourcify with `logError`, `logWarn`, `logInfo`, `logDebug`

@kuzdogan we need to decide what to log inside lib-sourcify

### DIFF
--- a/packages/lib-sourcify/README.md
+++ b/packages/lib-sourcify/README.md
@@ -112,8 +112,8 @@ console.log(match.status); // 'perfect'
 You can specify the log level using the `setLibSourcifyLoggerLevel(level)` where:
 
 - `0` is nothing
-- `1` is errors _[default]_
-- `2` is warnings
+- `1` is errors
+- `2` is warnings _[default]_
 - `3` is infos
 - `4` is debug
 
--- a/packages/lib-sourcify/README.md
+++ b/packages/lib-sourcify/README.md
@@ -104,3 +104,49 @@ const match = await verifyCreate2(
 console.log(match.chainId); // '0'. create2 matches return 0 as chainId
 console.log(match.status); // 'perfect'
 ```
+
+## Logging
+
+`lib-sourcify` has a basic logging system.
+
+You can specify the log level using the `setLibSourcifyLoggerLevel(level)` where:
+
+- `0` is nothing
+- `1` is errors _[default]_
+- `2` is warnings
+- `3` is infos
+- `4` is debug
+
+You can override the logger by calling `setLogger(logger: ILibSourcifyLogger)`. This is an example:
+
+```javascript
+const winston = require('winston');
+const logger = winston.createLogger({
+  // ...
+});
+
+setLibSourcifyLogger({
+  logLevel: 4,
+  setLevel(level: number) {
+    this.logLevel = level;
+  },
+  log(level, msg) {
+    if (level <= this.logLevel) {
+      switch (level) {
+        case 1:
+          logger.error(msg);
+          break;
+        case 2:
+          logger.warn(msg);
+          break;
+        case 3:
+          logger.info(msg);
+          break;
+        case 4:
+          logger.debug(msg);
+          break;
+      }
+    }
+  },
+});
+```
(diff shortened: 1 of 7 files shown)


packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -338,29 +340,29 @@ export async function performFetch(
   hash?: string,
   fileName?: string
 ): Promise<string | null> {
-  console.log(`Fetching the file ${fileName} from ${url}...`);
+  logInfo(`Fetching the file ${fileName} from ${url}...`);
```
marcocastignoli, line 343: This is an example of how to use the logger inside lib-sourcify

src/server/server.ts
```diff
@@ -28,11 +28,39 @@ import { resolveRefs } from "json-refs";
...
+          logger.info(msg);
+          break;
+        case 4:
+          logger.debug(msg);
+          break;
+      }
+    }
+  },
+});
+
```
marcocastignoli, line 63: This is how you override the default lib-sourcify's logger

packages/lib-sourcify/src/lib/logger.ts
```diff
@@ -0,0 +1,58 @@
+export interface ILogger {
+  logLevel: number;
+  log: (level: number, message: string) => void;
+  setLevel: (level: number) => void;
+}
+
+// Default logger behavior
+export const DefaultLogger: ILogger = {
+  logLevel: 1,
```
kuzdogan, line 9: Isn't default 3?

kuzdogan:
- [ ] Add how to override the logger in `lib-sourcify`'s `README`

packages/lib-sourcify/src/lib/logger.ts
marcocastignoli: From a quick search most libraries implement the warn level

src/server/server.ts
```diff
@@ -28,11 +28,39 @@ import { resolveRefs } from "json-refs";
...
+import { setLibSourcifyLogger } from "@ethereum-sourcify/lib-sourcify";
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const fileUpload = require("express-fileupload");
 
 const MemoryStore = createMemoryStore(session);
 
+// here we override the standard LibSourcify's Logger with a custom one
+setLibSourcifyLogger({
+  // No need to set again the logger level because it's set here
+  logLevel: 4,
```
kuzdogan, line 40:
```suggestion
  logLevel: process.env.NODE_ENV === "production" ? 3 : 4,
```

packages/lib-sourcify/src/lib/logger.ts
```diff
@@ -0,0 +1,58 @@
+export interface ILogger {
+  logLevel: number;
+  log: (level: number, message: string) => void;
+  setLevel: (level: number) => void;
+}
+
+// Default logger behavior
+export const DefaultLogger: ILogger = {
+  logLevel: 2,
```
kuzdogan, line 9: We should log info level, no?
```suggestion
  logLevel: 3,
```

kuzdogan (changes requested):
Nice! 

There needs to be more work on what to log and on which level to log, but we can work on that later as we see the needs. 

Just asking for small changes and clarifications.

packages/lib-sourcify/src/lib/logger.ts
marcocastignoli: I don't like library that by default they log info, I see most of them simply log warn and err

packages/lib-sourcify/README.md
```diff
@@ -104,3 +104,49 @@ const match = await verifyCreate2(
...
 ```
+
+## Logging
+
+`lib-sourcify` has a basic logging system.
+
+You can specify the log level using the `setLibSourcifyLoggerLevel(level)` where:
+
+- `0` is nothing
+- `1` is errors _[default]_
```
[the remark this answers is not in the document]
marcocastignoli: I forgot to update it here

PR #1083, opened 2023-06-28
Remove Web3 in favor of ethersjs

kuzdogan:
Draft PR

To keep consistency we should remove all Web3.js dependencies and instead use ethers.js.


kuzdogan:
Is codecov failing because of insufficient coverage? @marcocastignoli

src/server/controllers/verification/verification.common.ts
```diff
@@ -17,13 +17,13 @@ import {
...
 } from "@ethereum-sourcify/lib-sourcify";
 import { Session } from "express-session";
-import Web3 from "web3";
 import { AbiConstructor, AbiParameter } from "abitype";
 import QueryString from "qs";
 import fetch from "node-fetch";
 import { IVerificationService } from "../../services/VerificationService";
 import { IRepositoryService } from "../../services/RepositoryService";
 import { ContractMeta, ContractWrapper } from "../../common";
+import { id as keccak256str } from "ethers";
```
marcocastignoli, line 26: I see sometimes you are using id and other times keccak256, they are the same thing as far as I know (see the example here: https://docs.soliditylang.org/en/latest/path-resolution.html#id12) it would be more consistent to always use the same. But it's not bad

src/common/SourcifyEventManager/SourcifyEventManager.ts
```diff
@@ -4,48 +4,48 @@ import { EventManager, GenericEvents } from "../EventManager";
 interface Events extends GenericEvents {
   "*": (event: string, argument: any) => void;
   "Verification.MatchStored": (match: Match) => void;
-  "Monitor.Error.CantStart": (e: { chainId: string; message: string }) => void;
+  "Monitor.Error.CantStart": (e: { chainId: number; message: string }) => void;
   "Monitor.Started": (obj: {
-    chainId: string;
-    web3url: string;
+    chainId: number;
```
marcocastignoli, line 9: I think here it would be better to keep string and do a chainId.toString() when passing it. Just for the sake of always handling the chainId as a string

src/server/services/RepositoryService.ts
```diff
@@ -151,7 +151,7 @@ export class RepositoryService implements IRepositoryService {
   ): Array<FileObject> {
     const fullPath: string =
       this.repositoryPath +
-      `/contracts/${match}/${chain}/${toChecksumAddress(address)}/`;
+      `/contracts/${match}/${chain}/${getAddress(address)}/`;
```
marcocastignoli, line 154: this comment is valid for all the `getAddress`: there is a weird behavior, if the address has both uppercase and lowercase, it is considered formatted, but if it is wrongly formatted it throws an error (see the exampl here: https://docs.ethers.org/v5/api/utils/address/#utils-getAddress). IDK if we want to handle this

marcocastignoli (commented):
Did you run the monitor test? Just a few comments, my only real concern is about the `getAddress` that changes the behavior of the previous function.

src/server/controllers/verification/verification.common.ts
kuzdogan: Nope they are different. 

Ethersjs has 
1) `keccak256` that takes a `BytesLike` https://docs.ethers.org/v6/api/crypto/#keccak256
2) `id` that takes a string (UTF-8) https://docs.ethers.org/v6/api/hashing/#id

kuzdogan: Is the link you provided correct? Looks out of the context

marcocastignoli: sorry, wrong link because I thought it copied it :D the real link was from the ethers docs (but actually version 5). in the example it was stated that they are the same thing after converting the string into bytes. https://docs.ethers.org/v5/api/utils/hashing/#utils-id


```
// If needed, convert strings to bytes first:
utils.keccak256(utils.toUtf8Bytes("hello world"))
// '0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad'

// Or equivalently use the identity function:
utils.id("hello world")
// '0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad'
```

src/common/SourcifyEventManager/SourcifyEventManager.ts
kuzdogan: Hmm yeah the only place the `chainId` is a number is at the `Chain` type definition because `chains.json` has it as a number. In the rest of our code it's assumed a string mostly.

https://github.com/ethereum/sourcify/blob/staging/packages/lib-sourcify/src/lib/types.ts#L197

Let's do as you say.

src/server/services/RepositoryService.ts
kuzdogan: Hmm isn't that actually a good behavior? I don't expect any mixed cased addresses, and in fact, such an address is invalid. 

Also bear in mind to check v6 docs and not v5 https://docs.ethers.org/v6/api/address/#getAddress

marcocastignoli: I'm ok with using the ethers behavior, I'm just saying it will be different from before. For me we can keep it like you implemented it.

kuzdogan:
The `node-16` tests also cover monitor tests no? @marcocastignoli

marcocastignoli:
> The node-16 tests also cover monitor tests no? @marcocastignoli

oh yes, I totally forgot it 😆

src/server/controllers/verification/verification.common.ts
kuzdogan: Ok so we are talking about the same thing? What's your concern here? Since `id` itself is not really descriptive I'm using `keccak256str`. It is different than `keccak256`

marcocastignoli: Easy, I was just confused of having two different keccak functions. It is good like this

kuzdogan:
Merging this and moving the `node-v16` fix to another branch i.e. the fix for `should run getCreatorTx with regex`

marcocastignoli (approved):

PR #1092, opened 2023-07-05
Blockscout scraping

kuzdogan:
Recently noticed one of the tests started failing where we were scaping the transaction hash from the block explorer. This is because the chain's (Gnosis) Blockscout has been updated with the new Blockscout UI. This caused both the regex to not work anymore, as well as not being able to load the page contents with a fetch because the new UI is client-side rendered.

Update the scraping to try multiple regex'es, both the one for the old and the new UI.

Uses `puppeteer` to load the page to be scraped. This required to install the puppeteer dependencies on the containers that are running Sourcify and CI containers running tests: https://app.circleci.com/pipelines/github/ethereum/sourcify/4300/workflows/49a22229-32c3-408a-8459-622ecd863762/jobs/23775

Additionally we started getting `cov_{id} is not defined` errros in tests: https://app.circleci.com/pipelines/github/ethereum/sourcify/4309/workflows/35ad9ef7-524e-43b1-8ae5-4cf8d0455c65/jobs/23803
This was caused by the `nyc` library which hasn't been updated for some time. Move to `c8` instead for coverage which is more up-to-date. See https://github.com/istanbuljs/nyc/issues/1327


marcocastignoli (approved):
💯

PR #1101, opened 2023-07-11
Change local rpc

kuzdogan:
PR to change the primary Ethereum nodes used by Sourcify (Mainnet, Goerli, Sepolia). Instead of the LAN ones we use the ethpandaops' nodes.

Since the authentication is done through Headers, we had to change the `rpc` field in the `sourcify-chains.ts` into an ethers.js `FetchRequest`.

With that I also abstracted away the blockchain connections (`getTx`, `getBytecode` etc.) into a class called `SourcifyChain` inside `lib-sourcify`, such that we don't need to go through redundantly checking the `rpcUrl` + creating the `JsonRpc` or `Websocket` providers + sequentially trying the providers with the same sequential logic spread out to different places. With this, we have everything wrapped in `SourcifyChain` methods, and there's no need to create a new `Provider` every single time.


marcocastignoli (approved):
It is now much more readable having the SoucifyChain class. I read the edits multiple times but I cannot find any problem, 💯  again

PR #1105, opened 2023-07-12
Catch auxdata decode error when checking for perfect match

sealer3:
This helps when verifying contracts that don't have the CBOR metadata block at the end; currently, an error is thrown rather than making it a partial match.


marcocastignoli (approved):
You are right, lets just wait for @kuzdogan's approval, but for me we can merge this

kuzdogan:
Thank you!

PR #1166, opened 2023-08-29
Standalone monitor

kuzdogan:
PR for the complete writeup of the new sourcify-monitor as a standalone service.

Couple reminders:
- Removes the `monitored: true` field from `sourcify-chains.ts` and all other references to it.
- Renames `INFURA_ID` and `ALCHEMY_ID` to `xxx_API_KEY` to keep consistent internally and with chainid.network/chain.json format
- The npm workspaces does not include `package-lock.json` files in the submodules but in the root dir.
- Adds explicit typing to the `bytecode-utils`' `DecodedObject` type.
- Divides the old `MetadataSources` into `MetadataSource` (single source) and `MetadataSourceMap` as the `MetadataSourceMap` type is used in the `PendingContract` class' `pendingSources` and `fetchedSources`
- 

TODO:
- [x] ~~Update the `createsContract` function to check factory contracts.~~ (postponed see: https://github.com/ethereum/sourcify/tree/standalone-monitor-tests)
- [x] ~~Add testing with coverage (inside`services/monitor` and remove monitor tests from `test/`)~~ (postponed see: https://github.com/ethereum/sourcify/tree/standalone-monitor-tests)
- [x] Need to update `INFURA_ID` and `ALCHEMY_ID` to `INFURA_API_KEY` and `ALCHEMY_API_KEY` in https://github.com/sourcifyeth/infra secrets 
- [x] Fix logging level on production


marcocastignoli:
You said that we will have only one package-lock.json for the whole project, but I see many. Does it support different versions of a library for each package?

packages/lib-sourcify/src/lib/types.ts
```diff
@@ -30,13 +30,15 @@ export interface MissingSources {
...
-    license?: string;
-  };
+export interface MetadataSource {
+  keccak256: string;
+  content?: string;
+  urls?: string[];
+  license?: string;
+}
+
+export interface MetadataSourceMap {
```
marcocastignoli, line 40: This name was a bit confusing, but I see the fault is of `MetadataSource`, I honestly don't know a better name. Just leaving this comment to share this thought, but it's not blocking the PR.

services/monitor/src/ChainMonitor.ts
```diff
@@ -0,0 +1,281 @@
...
+
+      this.chainLogger.info(`Starting monitor on block ${startBlock}`);
+
+      // Start polling
+      this.pollBlocks(startBlock);
+
+      // Listen to new blocks
+      this.on(NEW_BLOCK_EVENT, this.processBlockListener);
+    } catch (err: any) {
+      this.chainLogger.error(`Error starting monitor: ${err.message}`);
```
marcocastignoli, line 83: Here we can add some information about the `ChainMonitor` (chainId?) so one knows which monitor failed directly from the error

services/monitor/src/ChainMonitor.ts
```diff
@@ -0,0 +1,281 @@
...
+   * Stops the monitor after executing all pending requests.
+   */
+  stop = (): void => {
+    this.chainLogger.info(`Stopping monitor`);
+    this.off(NEW_BLOCK_EVENT, this.processBlockListener);
+  };
+
+  // Tries to get the next block by polling in variable intervals.
+  // If the block is fetched, emits a creation event, decreases the pause between blocks, and goes to next block
+  // If the block isn't there yet, it increases the pause between blocks.
```
marcocastignoli, line 97: I would add an explanation of why increasing the pause between blocks is needed

services/monitor/src/ChainMonitor.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    pollNextBlock();
+  };
+
+  // ListenerFunction
+  private processBlockListener = async (block: Block) => {
+    this.chainLogger.info(`Found block ${block.number}. Now processing`);
+
+    for (const tx of block.prefetchedTransactions) {
+      // TODO: Check factory contracts with traces
```
marcocastignoli, line 138: Don't we want to include this?

services/monitor/src/ChainMonitor.ts
```diff
@@ -0,0 +1,281 @@
...
+      if (createsContract(tx)) {
+        const address = getCreateAddress(tx);
+        this.chainLogger.info(
+          `Found new contract in block ${block.number} with address ${address} created by tx ${tx.hash}`
+        );
+        this.processNewContract(tx.hash, address);
```
marcocastignoli, line 144: Here I see that we process all contracts concurrently. Could we risk to overload our server? Maybe awaiting would be enought until we implement something to handle many compilations?  Was it without the await also before?

Jonestly I prefer it like this (without the await), maybe we can try leaving it like this and see if it causes many problems.

services/monitor/src/PendingContract.ts
```diff
@@ -0,0 +1,165 @@
...
+      .fetch(this.metadataHash)
+      .catch((err) => {
+        throw new Error(
+          `Can't fetch metadata for ${this.address} ${this.metadataHash.hash}: ${err}`
+        );
+      });
+    logger.info(
+      `Fetched metadata for ${this.address} on chain ${this.chainId} from ${this.metadataHash.origin}`
+    );
+    // TODO: check if metadata hash matches this.metadataHash.hash
```
marcocastignoli, line 52: Why do we need this check? Because a potential source (ipfs, swarm) could not return the right file requested? If that's the case, I think we can skip the check and trust the source.

services/monitor/src/PendingContract.ts
```diff
@@ -0,0 +1,165 @@
...
+    this.metadata = JSON.parse(metadataStr) as Metadata;
+    this.pendingSources = { ...this.metadata.sources }; // Copy, don't mutate original.
```
marcocastignoli, line 54: Be aware that { ...object} clones only the first level properties, but deep properties are "cloned" as references to the first object. The only way to safely deep clone an object is to use `structuredClone`

services/monitor/src/DecentralizedStorageFetcher.ts
```diff
@@ -0,0 +1,113 @@
...
+      this.on(`${fileHash.hash} fetch failed`, failListener);
+
+      // Need the file for the first time.
+      if (!this.uniqueFiles[fileHash.hash]) {
+        this.uniqueFiles[fileHash.hash] = true;
+        this.subcriberCounter++;
+        this.uniqueFilesCounter++;
+
+        // TODO: Handle multiple gateways.
+        const gatewayFetcher = this.gatewayFetchers[0];
```
marcocastignoli, line 80: If I understood correctly we can pass multiple IPFS gateways, but right now we are using only the first one, right?

I think that it makes sense to have at least one IPFS fallback in the first release, we know IPFS gateways are not stable.

We could implement it so that it calls different gateways concurrently and the first one that replies makes the other stop. If no one replies we can safely say that the file is not available on IPFS. For the future, this also makes sense when using a local ipfs node, we could also optimize by, for example, adding as peers all the nodes that successfully respond while the local node cannot reply.

services/monitor/src/DecentralizedStorageFetcher.ts
```diff
@@ -0,0 +1,113 @@
...
+      // Subscribe to fetching of the file.
+      this.on(`${fileHash.hash} fetched`, successListener);
+      // Subscribe to failure of fetching the file.
+      this.on(`${fileHash.hash} fetch failed`, failListener);
+
+      // Need the file for the first time.
+      if (!this.uniqueFiles[fileHash.hash]) {
+        this.uniqueFiles[fileHash.hash] = true;
+        this.subcriberCounter++;
+        this.uniqueFilesCounter++;
```
marcocastignoli, line 77: These are not very clear to me, can you explain? :)

marcocastignoli (commented):
there is no really serious error, I leave it up to you to choose what to fix and what not for this release.

kuzdogan:
> You said that we will have only one package-lock.json for the whole project, but I see many. Also, does it support different versions of a library for each package?

I haven't removed all of them. Should I?

marcocastignoli:
> I haven't removed all of them. Should I?
> 
> Not sure how it handles each version but I think it does. Otherwise it wouldn't make sense to remove the `-lock`s from each package.

I think that if the "standard" is to not have package-lock.json in any package, let's remove them. But first we need to make sure that there are no problems regarding the libraries versions.

services/monitor/src/PendingContract.ts
kuzdogan: Good to know thanks

services/monitor/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    : path.join(__dirname, filePath);
+
+  if (fs.existsSync(absolutePath)) {
+    const jsonData = fs.readFileSync(absolutePath, "utf8");
+    if (!jsonData) {
+      if (throws) throw new Error(`File ${absolutePath} exists but is empty.`);
+      console.warn(`File ${absolutePath} exists but is empty.`);
+      return undefined;
+    }
+    return JSON.parse(jsonData);
```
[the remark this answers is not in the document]
kuzdogan: Like this?

https://github.com/ethereum/sourcify/blob/fa1f431908fc7d498eb86e9183cef3fd19307c15/services/monitor/src/index.ts#L37-L42

packages/lib-sourcify/src/lib/types.ts
marcocastignoli: yeah yeah it's clear, I just didn't like the name. This is not a big deal, just do as you prefer.

services/monitor/src/ChainMonitor.ts
marcocastignoli: If we will include this functionality before merging this PR

services/monitor/src/PendingContract.ts
marcocastignoli: Oh I see, for sources we can use the keccak but for the metadata file we must encode the metadata to ipfs or bzzr address right?

kuzdogan: Yes. let me add it

kuzdogan:
> > I haven't removed all of them. Should I?
> > Not sure how it handles each version but I think it does. Otherwise it wouldn't make sense to remove the `-lock`s from each package.
> 
> I think that if the "standard" is to not have package-lock.json in any package, let's remove them. But first we need to make sure that there are no problems regarding the libraries versions.

Removed other `-lock`s. 

It seems to correctly handle the dependencies in the parent `-lock`. I tried for example updating the `solc` version in `lib-sourcify` and corresponding versions are also updated in the parent `package-lock`. You can check the `-lock` diff: https://github.com/ethereum/sourcify/pull/1166/commits/3b04dfb506b5c87121a6c709c0273ca8d4fe81f9

PR #1167, opened 2023-08-30
Release

marcocastignoli:


nx.json
```diff
@@ -0,0 +1,20 @@
+{
```
kuzdogan, line 1: What's this?

marcocastignoli: This is needed in order to handle the cache for the tests. So it runs the tests only if there are differences in the build folder

.circleci/continue_config.yml
```diff
@@ -218,16 +220,16 @@ jobs:
       - checkout
       - run:
           name: install dependencies
-          command: npm install && npx lerna bootstrap
+          command: npm install
```
kuzdogan, line 223: Asking because idk, so now there's no need for `npx lerna bootstrap ` but we do `npm install`. Does the package linking handled in that case?

Also we need to update the docs in this case

marcocastignoli: yes it handles everything by itself. I needed to add the `workspace` key in the main `package.json`.

lerna.json
```diff
@@ -7,8 +7,11 @@
     },
     "bootstrap": {
       "npmClientArgs": ["--no-package-lock"]
+    },
+    "version": {
+      "allowBranch": "master"
```
kuzdogan, line 12: Does this force only running version on `master`?

lerna.json
```diff
@@ -7,8 +7,11 @@
...
     }
   },
   "packages": [".", "ui", "packages/*"],
-  "version": "0.1.5"
+  "version": "independent"
```
kuzdogan, line 16: Also an explainer here please :) I don't understand what this does

lerna.json
marcocastignoli: yes

kuzdogan:
- [ ] Publish the docs on release

PR #1170, opened 2023-09-07
Integrate Alliance database

marcocastignoli:
* save new verified contracts into the alliance database
* track transformations into Matches
* better naming for runtime vs creation, recompiled vs deployed bytecode
* always verify both runtime and creation bytecode

- [ ] Implement tests


packages/lib-sourcify/src/lib/types.ts
```diff
@@ -159,10 +159,58 @@ export interface RecompilationResult {
   immutableReferences: ImmutableReferences;
 }
 
+export type Transformation = {
+  type: 'insert' | 'replace';
+  reason: 'constructor' | 'library' | 'immutable' | 'auxdata';
+  offset: number;
+  id?: string;
```
kuzdogan, line 166: What's the id for?

packages/lib-sourcify/src/lib/verification.ts
```diff
@@ -273,50 +258,72 @@ export async function verifyCreate2(
...
       match.libraryMap = libraryMap;
       match.immutableReferences = immutableReferences;
-      match.status = 'partial';
+      match.runtimeMatch = 'partial';
+      match.runtimeTransformations?.push(
+        AuxdataTransformation(trimmedCompiledRuntimeBytecode.length, '0')
+      );
+      match.runtimeValues.cborAuxdata = { '0': auxdata };
     }
   }
```
kuzdogan, line 328: I thought we are using the auxdata extraction algorithm here to match contracts, when there are multiple auxdata

packages/lib-sourcify/src/lib/SourcifyChain.ts
```diff
@@ -239,4 +323,45 @@ export default class SourcifyChain {
         this.chainId
     );
   };
+
+  getContractCreationBytecode = async (
+    address: string,
+    transactionHash: string
+  ): Promise<string> => {
+    const txRecipt = await this.getTxRecipt(transactionHash);
+    const tx = await this.getTx(transactionHash);
```
kuzdogan, line 332: What's the difference between the tx and the reciept?

packages/lib-sourcify/src/lib/SourcifyChain.ts
```diff
@@ -239,4 +323,45 @@ export default class SourcifyChain {
         this.chainId
     );
   };
+
+  getContractCreationBytecode = async (
+    address: string,
+    transactionHash: string
+  ): Promise<string> => {
+    const txRecipt = await this.getTxRecipt(transactionHash);
```
kuzdogan, line 331: Typo: Receipt

packages/lib-sourcify/src/lib/SourcifyChain.ts
```diff
@@ -239,4 +323,45 @@ export default class SourcifyChain {
...
+    const tx = await this.getTx(transactionHash);
+    let creationBytecode = '';
+
+    if (txRecipt.contractAddress !== null) {
```
kuzdogan, line 335: Is this the address of the contract created? Does all RPCs return this? 

I thought this value wasn't always returned. That's why we double check the contract address to be created in `matchWithCreaetionTx` using EOA address + nonce

src/server/services/RepositoryService.ts
```diff
@@ -648,4 +667,312 @@ export class RepositoryService implements IRepositoryService {
...
+      match.deployedCreationBytecode === undefined
+    ) {
+      // Can only store contracts with both runtimeBytecode and creationBytecode
+      return;
+    }
+
+    if (match.creatorTxHash === undefined) {
+      // Can only store matches with creatorTxHash
+      return;
+    }
```
kuzdogan, line 702: Uh oh. I wasn't aware of these requirements. It means we just can't send contracts most of the times

kuzdogan (changes requested):
Ok well this is quite a lot.. Let's do one round of review. Then maybe we'll do a live session together.

packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -549,3 +668,127 @@ function reorderAlphabetically(obj: any): any {
 
   return ordered;
 }
+
+type Bytecode = string;
```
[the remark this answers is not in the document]
marcocastignoli: I think in general it is good practice, but using it only here it doesn't make sense. We can replace it everywhere or remove it from here

packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -233,6 +243,99 @@ export class CheckedContract {
...
+    let editedBytecode = '';
+    switch (bytecodeType) {
+      case 'creation': {
+        editedBytecode = `0x${contract?.evm.bytecode.object}`;
+        break;
+      }
+      case 'runtime': {
+        editedBytecode = `0x${contract?.evm?.deployedBytecode?.object}`;
+      }
+    }
```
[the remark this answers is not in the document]
marcocastignoli: Right, I was doing more stuff into that function and I forgot to refactor it after I removed them

packages/lib-sourcify/src/lib/CheckedContract.ts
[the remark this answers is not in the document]
marcocastignoli: I removed it

packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -549,3 +668,127 @@ function reorderAlphabetically(obj: any): any {
...
+) {
+  let auxdatasDiffs = [];
+  for (let i = 0; i < knownAuxdatas.length; i++) {
+    const diffPositions = getDiffPositions(knownAuxdatas[i], editedAuxdatas[i]);
+    auxdatasDiffs.push({
+      offsetStart: diffPositions[0],
+      offsetEnd:
+        knownAuxdatas[i].length - diffPositions[diffPositions.length - 1] - 1,
+      real: knownAuxdatas[i],
+      diff: knownAuxdatas[i].substring(
```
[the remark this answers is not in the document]
marcocastignoli: `knownAuxdatas` is correct, but `originalAuxdatas` is a better name

packages/lib-sourcify/src/lib/types.ts
```diff
@@ -320,3 +374,96 @@ export interface JsonInput {
...
+  };
+}
+export interface CompilerOutput {
+  errors?: CompilerOutputError[];
+  sources?: CompilerOutputSources;
+  contracts: CompilerOutputContracts;
+}
+
+export interface CompiledContractArtifactsCborAuxdata {
+  [index: string]: {
```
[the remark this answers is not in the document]
marcocastignoli: Because of the way the solidity compiler outputs every index. It is part of the spec of the Alliance Database

packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -549,3 +668,127 @@ function reorderAlphabetically(obj: any): any {
...
+): CompiledContractArtifactsCborAuxdata {
+  const auxdataDiffs = getAuxdatasDiff(originalAuxdatas, editedAuxdatas);
+
+  const diffPositionsBytecodes = getDiffPositions(
+    originalBytecode,
+    editedBytecode
+  );
+  const auxdataPositions: CompiledContractArtifactsCborAuxdata = {};
+
+  for (const offsetOfDiffInByteocde of diffPositionsBytecodes) {
```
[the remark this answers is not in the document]
marcocastignoli: This is an improvement but it wouldn't make the function more optimized, the real optimization here is to skip all the diffs at offset < than the index of the last character of the found auxdata.

packages/lib-sourcify/src/lib/verification.ts
```diff
@@ -273,50 +258,72 @@ export async function verifyCreate2(
...
+
   // Check if is a library with call protection
   // See https://docs.soliditylang.org/en/v0.8.19/contracts.html#call-protection-for-libraries
-  recompiledDeployedBytecode = checkCallProtectionAndReplaceAddress(
-    recompiledDeployedBytecode,
-    deployedBytecode
+  recompiledRuntimeBytecode = checkCallProtectionAndReplaceAddress(
+    recompiledRuntimeBytecode,
+    deployedRuntimeBytecode
   );
```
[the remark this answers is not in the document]
marcocastignoli: good catch, we need to discuss it with the Alliance

packages/lib-sourcify/src/lib/types.ts
marcocastignoli: valid for the case of auxdata transformation, it is the index of the auxdata (to handle multiple auxdata), in theory we could use the name of the contract instead of an number.

packages/lib-sourcify/src/lib/verification.ts
marcocastignoli: We are not, for now we are using this only for the alliance database, I asked you and we agreed to skip this for now. After we merge this we can work to support it in sourcify.

packages/lib-sourcify/src/lib/SourcifyChain.ts
marcocastignoli: from the ethers docs: 

* A **TransactionReceipt** includes additional information about a transaction that is only available after it has been includfed in the block.
* A **TransactionResponse** includes all properties about a transaction that was sent to the network, which may or may not be included in a block.

packages/lib-sourcify/src/lib/SourcifyChain.ts
marcocastignoli: exactly, read the lines later:

```typescript
if (txReceipt.contractAddress !== null) {
      // EOA created
} else {
      // Factory created
```

src/server/services/RepositoryService.ts
marcocastignoli: I managed to get every time all the requirements. I forced to execute both deployed and creation verification every time.

marcocastignoli:
I merge staging into this branch, for some reason we don't have any conflict, it is weird 🤔

PR #1178, opened 2023-09-16
Add chain 1433

iamakshaykumar:
# Add New Chain <chainId>

Thanks for your pull request to add a new support in Sourcify.

If you haven't done so, please follow the instructions on [how to request chain support](https://docs.sourcify.dev/docs/chain-support/) in docs.

Please check the following items before submitting your pull request.

## Checklist

- [x] I haven't modified the [chains.json](../../src/chains.json) file directly.
- [x] In [sourcify-chains.ts](../../src/sourcify-chains.ts) file
  - [x] I've set `monitored: false`.
- [x] `test-new-chain` test in Circle CI is passing.


kuzdogan:
The test-new-chain is failing because of the Puppeteer but that's another issue on our side.

Normally it should not even reach there. This is because the bytecodes are not matching. Are you sure the contract you provided matches `"shared/1_Storage.sol"`? The bytecodes are different. Here's the diff:

image

Please deploy a new contract [following the docs](https://docs.sourcify.dev/docs/chain-support/) or provide a matching contract.

iamakshaykumar:
bytecode changed

kuzdogan:
You are changing the bytecode of our template contract. Please don't do that.

Instead you need to deploy a matching contract as laid out in the docs and provide that correct contract's address

PR #1193, opened 2023-10-17
Standalone monitor tests

kuzdogan:
Adds basic tests for the new standalone monitor


services/monitor/src/ChainMonitor.ts
```diff
@@ -31,6 +31,7 @@ export class ChainMonitor extends EventEmitter {
   private blockIntervalLowerLimit: number;
   private bytecodeInterval: number;
   private bytecodeNumberOfTries: number;
+  private running = false;
```
marcocastignoli, line 34: I see that this variable is set at start and stop, is it needed to set it to false also when there is an error?

marcocastignoli (approved):
Only this comment, it took me a while to understand the flow behind the fake sourcify URL. Maybe it would be better to have a util function that returns the fake url + the nodeInterceptor instance, so it's clear where the url is coming from.

services/monitor/src/ChainMonitor.ts
kuzdogan: Not sure but I don't think so.

kuzdogan:
> Only this comment, it took me a while to understand the flow behind the fake sourcify URL. Maybe it would be better to have a util function that returns the fake url + the nodeInterceptor instance, so it's clear where the url is coming from.

Was the concept (nock) unclear? Would it be clear do you think if we'd add some explanation on FAKE_SORUCIFY and other places why they are there?

marcocastignoli:
> Was the concept (nock) unclear? Would it be clear do you think if we'd add some explanation on FAKE_SORUCIFY and other places why they are there?

It wasn't clear the variable name FAKE_SORUCIFY. It would have been better something like MOCK_URL, because at the beginning I thought it was some sort of test to test that the monitor would handle wrong urls

marcocastignoli (approved):

PR #1197, opened 2023-10-23
Fix monitor alchemy rpc by supporting specific chain's key

marcocastignoli:


marcocastignoli:
I don't know if we need additional env in the infra repo in order to make this PR work.

kuzdogan:
The fix will be cherry-picked to master

PR #1199, opened 2023-10-24
Remove solc as a lib-sourcify dependency

marcocastignoli:
- Implement an Interface that needs to be passed to lib-sourcify's function to be used as compiler
- remove everything regarding the compiler from lib-sourcify (keep it for now in its dependencies)
- potentially replace the solc in the test with a mock compiler


services/server/src/server/controllers/verification/verification.common.ts
```diff
@@ -29,6 +33,29 @@ import config from "../../../config";
...
+
+class Solc implements ISolidityCompiler {
+  async compile(
+    version: string,
+    solcJsonInput: JsonInput,
+    forceEmscripten: boolean = false
+  ): Promise<CompilerOutput> {
+    return await useCompiler(version, solcJsonInput, forceEmscripten);
+  }
+}
```
kuzdogan, line 47: Why not import the type from lib-sourcify?

kuzdogan (commented):
Looks OK. 

Can you state again what was the reason for these changes? 

- [ ] Also need to document the changes in the lib-sourcify README

marcocastignoli:
> Can you state again what was the reason for these changes?

1. The compiler is a big dependency and depending on the environment developers need a very specific version of it
2. We are going to use solc from an external service, so we need a way for lib-sourcify to call our external service for compilation
3. This is part of the browserification of lib-sourcify, solcjs has dependencies that would have forced us to follow specific rules to browserify lib-sourcify, without solcjs we are free to implement lib-sourcify for the browser as we think it's better

kuzdogan:
> > Can you state again what was the reason for these changes?
> 
> 1. The compiler is a big dependency and depending on the environment developers need a very specific version of it
> 2. We are going to use solc from an external service, so we need a way for lib-sourcify to call our external service for compilation
> 3. This is part of the browserification of lib-sourcify, solcjs has dependencies that would have forced us to follow specific rules to browserify lib-sourcify, without solcjs we are free to implement lib-sourcify for the browser as we think it's better

Right thanks. 

Just a small thing I think in 1) ~~"developers need a very specific version of it"~~ "developers need a specific platform binary "

PR #1203, opened 2023-10-26
Refactor the code that fetches the contract creation transaction

marcocastignoli:
I implemented a standard method to deal with contraction creation transaction fetching.

Basically a chain can provide a `ContractCreationFetcher` that is an object containing 

```ts
{
  url: "https://eth.blockscout.com/api/v2/addresses/${ADDRESS}", // the url with ${ADDRESS} that is going to be replaced
  responseParser: (response: any) => response?.creation_tx_hash // the function to parse the fetch response
}
```

One of the main objective of this PR is to remove puppeteer as a dependency, but I first want to double check with @kuzdogan if it makes sense since blockscout only supports API for 15 chains (without puppeteer we are going to lose support to about 80% of previously supported blockscout chains via puppeteer)


kuzdogan:
Why are we going to lose support to 80% of Blockscout? Aren't we just doing a `fetch()` and using the regex to extract the creator tx? I don't see the regex being used here. The reason we were using puppeteer is that the new Blockscouts are client-side rendered. The old ones should still work.

kuzdogan:
Also I know what you've done is in the same direction as what we've had previously but I think the way this is structured (i.e. also how we did before) is not ideal, because I suspect people won't be able to add the blockscout spec, etherscan spec both to the `sourcify-chains.ts` and then go to `config.ts` and add there correctly. IMO we should keep everything about the chain spec at one place: `sourcify-chains.ts`. 

Therefore I feel it makes more sense to have something like:
```js
  "534": {
    supported: false,
    blockscout: "https://exampleblockscout.com/api/v2/",
    etherscan: 
    etherscanAPI:...
    // 
  },

**Review by Gentleman1974 — APPROVED** [review]

File: services/server/src/server/services/utils/contract-creation-util.ts [hunk 1]
```diff
@@ -1,74 +1,285 @@
...
+}
+
+/**
+ * Finds the transaction that created the contract by either scraping a block explorer or querying a provided API.
+ *
+ * @param sourcifyChain
+ * @param address
+ * @returns
+ */
+export const getCreatorTx = async (
```
kuzdogan, line 192: I think in this function we should prioritize the APIs, and move the scrape methods to the bottom

kuzdogan (approved):
Fixed myself the blockscan naming to BlocsScan (that's how it's named https://blocksscan.io/) and removed the fetcher checks. Fetcher is never null.

TBH I find it overly complicated with a function to create a fetcher, then again check the type of the fetcher etc. Essentially there's only scraping and an API. The responseParser is a quite elegant solution. We could've just passed the URL + regex to a scraper function and URL + responseParser to a api fetcher function for all chains. 

But it's ok.

marcocastignoli:
> TBH I find it overly complicated with a function to create a fetcher, then again check the type of the fetcher etc. Essentially there's only scraping and an API. The responseParser is a quite elegant solution. We could've just passed the URL + regex to a scraper function and URL + responseParser to a api fetcher function for all chains.

yeah I see your point, I did that initially for the previous version. Maybe we can keep it like this, then if we notice that it doesn't bring any benefit, simplify :)

PR #1208, opened 2023-11-13
Add chain 2047,2048

BoThe1K:
# Add New Chain 2047,2048
## Checklist

- [x] I haven't modified the [chains.json](../../src/chains.json) file directly.
- [x] In [sourcify-chains.ts](../../src/sourcify-chains.ts) file
- [ ] `test-new-chain` test in Circle CI is passing.


BoThe1K:
@kuzdogan I guess this
```
      const metadataPath = path.join(
        "test",
        "chains",
        "sources",
        relativeMetadataPath
      );
```

should be 
```

      const metadataPath = path.join(
        "services",
        "server",
        "test",
        "chains",
        "sources",
        relativeMetadataPath
      );
```
as project structure changed, also look pipeline does not work and does not recognize `test:chains` cmd.

PR #1218, opened 2023-11-21
Compiler-lambda

marcocastignoli:
This change allows the configuration of the preferred Solidity compiler,
either lambda or local, by selecting the appropriate implementation
based on the value of the environment variable `SOLIDITY_COMPILER`.
This commit implements`SolcLambda` and `SolcLocal` classes to
represent the Lambda and local compiler implementations respectively.
The `ISolidityCompiler` interface is used to define the common methods
for both implementations.


marcocastignoli:
I'm going to merge this PR even if it's not completely ready since it's a priority to fix all the down-times we are having in the last weeks

PR #1232, opened 2023-12-13
Server config by passing config files, remove create2 verification

kuzdogan:
This PR allows users to modify the default Sourcify server config by passing config files under `services/server/config`. The default config is set at `default.js`. Production and staging configs `production.js` and `staging.js` respectively (must set up `NODE_ENV=production` or `staging`). These only override the given fields in this config. 

Similar to `.env`, it is possible to create a `config/local.js` file, which is ignored, to set up a local development environment.

Environment variables are still passed via `.env`, for which a template is given at `.env.dev`. However this PR removes the env vars that can be public config values. This means most of the public env vars are now configs, with the exception of the `IPFS_GATEWAY` which is used at the `lib-sourcify` to assemble a contract. This, in the future can be a config passed to the `lib-sourcify` constructor and a LibSourcify instance created by the server. For now this is left, and also the `IPFS_API` variable is kept to keep IPFS vars together.

### `.env` vars

Moves to config:
```bash
REPOSITORY_PATH=
REPOSITORY_SERVER_URL=
SOLC_REPO=
SOLJSON_REPO=
SERVER_PORT=
```

Removes:
```bash
AUTH0_AUDIENCE=
AUTH0_ISSUERBASEURL=
AUTH0_TOKENSIGNINGALG=
AUTH0_CLIENTID=

## These are not in sourcify-chains-default-json
NODE_URL_MAINNET=
NODE_URL_GOERLI=
NODE_URL_SEPOLIA=

TESTING=
TAG=
DOCKER_HOST_SERVER_PORT=
DOCKER_HOST_SOLC_REPO=
DOCKER_HOST_SOLJSON_REPO=
DOCKER_HOST_REPOSITORY_PATH=
```

### Create2

We wanted to remove the CREATE2 verification feature as it's not being used, spam prone (can spam by giving arbitrary salt), and the spam prevention compicates the code. While cleaning up `.env` and removing the `AUTH0` I also decided to remove the CREATE2 related code. 

### Misc

- Removes the docker-compose files for the server. The docker itself is pretty straightforward
- Remove `ForbiddenError` and `UnauthorizedError` as these were used in create2 endpoint authenticaiton
- 

TODO:
- [ ] Update the env vars in sourcifyeth/infra
- [ ] Update the docs


services/server/src/server/services/services.ts
```diff
@@ -1,14 +1,14 @@
...
+import config from "config";
 
 export const services = {
   verification: new VerificationService(supportedChainsMap),
   storage: new StorageService({
     ipfsRepositoryServiceOptions: {
       ipfsApi: process.env.IPFS_API as string,
-      repositoryPath: config.repository.path,
+      repositoryPath: config.get("repository.path"),
       repositoryServerUrl: process.env.REPOSITORY_SERVER_URL as string,
```
marcocastignoli, line 12: I think here you forgot to use config.get("repository.url")

services/server/README.md
```diff
@@ -86,35 +96,15 @@ npm start
...
 
-Then run:
+Also set up the environment variables in the `.env` file. You can see the list of required environment variables in the `.env.dev` file. Pass it with the `--env-file` flag or use the `--env` flag to pass individual environment variables.
 
 ```bash
-docker-compose up
+docker run \
+  -p 5555:5555 \
+  -v path/to/custom/sourcify-chains.json:/home/app/services/server/dist/sourcify-chains.json \
+  -v path/to/custom/config.js:/home/app/services/server/dist/config.js \
```
marcocastignoli, line 107: `config.js` or `local.js`?

marcocastignoli (changes requested):
I spotted just these two minor issues, I think next step is to completely remove process.env from lib-sourcify and pass them as options (just like we did it with the logger for lib-sourcify)

kuzdogan:
Addressed the issues, thanks. 

> I spotted just these two minor issues, I think next step is to completely remove process.env from lib-sourcify and pass them as options (just like we did it with the logger for lib-sourcify)

Do you mean as "next step" later or in this PR? Is it ready to go?

marcocastignoli (approved):
For me it's ready to go, we'll work on the lib-sourcify optimization in another PR 👍

PR #1277, opened 2024-02-22
Allow passing Etherscan API key as parameter in import from Etherscon

marcocastignoli:
Supported both in stateless, session + UI

### DIFF
--- a/services/server/src/server/controllers/verification/etherscan/etherscan.common.ts
+++ b/services/server/src/server/controllers/verification/etherscan/etherscan.common.ts
@@ -75,7 +75,8 @@ export const getSolcJsonInputFromEtherscanResult = (
 
 export const processRequestFromEtherscan = async (
   sourcifyChain: SourcifyChain,
-  address: string
+  address: string,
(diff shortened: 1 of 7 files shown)


kuzdogan (changes requested):
Two things:

1. In the UI if you enter the Address and then choose the chain, it immediately sends the request without waiting for the API key. I think the API key must come first in the form for better UX.

2. The `apiKey` is not added to the **session** OpenAPI yaml file but it still allows sending the API key in the request. Doesn't the OpenAPI request validator need to throw saying this field in the request is unexpected and has to throw?

marcocastignoli:
>  I think the API key must come first in the form for better UX.

Nice idea

> Doesn't the OpenAPI request validator need to throw saying this field in the request is unexpected and has to throw?

I'll also check this

marcocastignoli:
@kuzdogan I think it's better to open another PR to fix unexpected fields in requests because it's more impactful than this. https://cdimascio.github.io/express-openapi-validator-documentation/usage-validate-requests/

PR #1298, opened 2024-03-13
Improve logging

kuzdogan:
<sub>View in Huly <a href="https://huly.app/guest/w-kaanuzdogan-sourcify-65d30de0-75121c5c6c-fd615e?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NWYxZTZkMmExMzdmZWEzMDhkNDZiNmMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6Incta2FhbnV6ZG9nYW4tc291cmNpZnktNjVkMzBkZTAtNzUxMjFjNWM2Yy1mZDYxNWUiLCJwcm9kdWN0SWQiOiIifQ.y9KOlZ3uIs3Ts04p485b3cBbKT8QaEi0wSI5BAXw9xc">HI-1261</a></sub>

PR #1310, opened 2024-03-26
Enable database in tests and fix updating matches

marcocastignoli:
- Enable database check in the tests: assert that the match exists in sourcify_matches with address,chain,status
- There was a problem while updating a match from partial to perfect, it didn't update the row in sourcify_matches, this PR solves this problem
- I had to make `assertVerification` and `assertVerificationSession` async. So then I had to update all the calls in the tests

### TODO

- [x] After we merge #1298 I'm going to update this with new logging


marcocastignoli:
For some reason, tests are failing on CircleCI while they are successfully running  locally.

marcocastignoli:
I thought that the env variables in `services/server/test/server.js` were enough in order to start Sourcify with specific env variables. It turns out that is not the case, I fixed this by passing the environments in circleci's `npx lerna run build && npx lerna run lerna-test` configuration

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -144,13 +144,25 @@ export class SourcifyDatabaseService
         creationMatch: match.creationMatch,
       });
     } else if (type === "update") {
-      if (verifiedContractId) {
-        await Database.updateSourcifyMatch(this.databasePool, {
+      if (!verifiedContractId) {
+        throw new Error(
```
marcocastignoli, line 148: @kuzdogan do we need to add a logger.error() before throwing?

kuzdogan: Anyting we want to be alerted about needs to be `logger.error` logged. If that's the case yes.

marcocastignoli:
@kuzdogan All the errors thrown into storage services are logged inside each catch

e.g.
https://github.com/ethereum/sourcify/blob/a6c54f6f72e310402e5e87e0b57b45c448722578/services/server/src/server/services/StorageService.ts#L166-L174

So I don't think additional error-related logs are necessary, lmk what you think.

kuzdogan:
@marcocastignoli If that is the case, ok this way is the way to go

kuzdogan (approved):

PR #1327, opened 2024-04-05
Fix etherscan verification tests

manuelwedler:


services/server/test/etherscan/etherscanInstances.js
```diff
@@ -33,7 +33,10 @@ describe("Test each Etherscan instance", function () {
 
   let testedChains = [];
   for (const chainId in testContracts) {
-    if (process.env.TEST_CHAIN && process.env.TEST_CHAIN !== chainId) continue;
+    if (process.env.TEST_CHAIN && process.env.TEST_CHAIN !== chainId ||
+      !sourcifyChainsMap[chainId].supported) {
```
kuzdogan, line 37: Let's we throw with a message when the chain is not supported, instead of continuing? The idea is to recognize those and remove them, like we would have noticed we haven't removed Goerli. With this, it goes unnoticed.

kuzdogan (changes requested):

services/server/test/etherscan/etherscanInstances.js
manuelwedler: Sure, makes more sense like this.

Issue #1321, opened 2024-04-03
Swapping session `MemoryStore` with a Postgres store


kuzdogan:
We are using `MemoryStore` in our `express-session` ([here](https://github.com/ethereum/sourcify/blob/a6c54f6f72e310402e5e87e0b57b45c448722578/services/server/src/server/server.ts#L329-L331)]). According to [express-session](https://github.com/expressjs/session) "MemoryStore, is purposely not designed for a production environment". So we should switch to another store. Since we now have the PostgresDB, we can use a Postgres store. 

I'd say instantiate the DatabaseStore only if the SourcifyDatabaseService is present. I guess we can check this by looking at `StorageService.sourcifyDatabase`. Otherwise just use MemoryStore but also log a warning.

Potential fix for #1289 and #1270 

Some notes:
- [ ] Check how the sessions are deleted from the DB. In `MemoryStore` has a `checkPeriod` parameter that will kill and delete the session from memory after some time but we need to check if this is done automatically with the pg library we are using or we need to clean up manually.

PR #1335, opened 2024-04-08
Implement postgresql based session

marcocastignoli:
See #1321

### TODO

- [x] Implement tests for postgresql based session
- [x] exclude session in database backup
- [x] check database session pruning

### DIFF
--- a/services/server/src/config/default.js
+++ b/services/server/src/config/default.js
@@ -18,7 +18,7 @@ module.exports = {
     secret: process.env.SESSION_SECRET || "CHANGE_ME",
     maxAge: 12 * 60 * 60 * 1000, // 12 hrs in millis
     secure: false, // Set Secure in the Set-Cookie header i.e. require https
-    storeType: "memory",
+    storeType: "memory", // Where to save the session info. "memory" is only good for testing and local development. Don't use it in production!
   },
   // It is possible to outsource the compilation to a lambda function instead of running locally. Turned on in production.
   // Requires env vars AWS_REGION, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
--- a/services/server/src/config/default.js
+++ b/services/server/src/config/default.js
@@ -18,6 +18,7 @@ module.exports = {
     secret: process.env.SESSION_SECRET || "CHANGE_ME",
     maxAge: 12 * 60 * 60 * 1000, // 12 hrs in millis
     secure: false, // Set Secure in the Set-Cookie header i.e. require https
+    storeType: "memory",
   },
   // It is possible to outsource the compilation to a lambda function instead of running locally. Turned on in production.
   // Requires env vars AWS_REGION, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
--- a/services/server/src/server/session.ts
+++ b/services/server/src/server/session.ts
@@ -10,6 +10,8 @@ import { NextFunction, Request, Response } from "express";
 
 function initMemoryStore() {
   const MemoryStore = createMemoryStore(expressSession);
+
+  logger.warn("Using memory based session");
   return new MemoryStore({
     checkPeriod: config.get("session.maxAge"),
   });
@@ -31,8 +33,11 @@ function initDatabaseStore() {
 
   const PostgresqlStore = genFunc(expressSession);
 
+  logger.warn("Using database based session");
   return new PostgresqlStore({
     pool: pool,
+    // Pruning expired sessions every 15 minutes
+    pruneSessionInterval: 900,
   });
 }
 
--- a/services/server/src/server/session.ts
+++ b/services/server/src/server/session.ts
@@ -11,7 +11,9 @@ import { NextFunction, Request, Response } from "express";
 function initMemoryStore() {
   const MemoryStore = createMemoryStore(expressSession);
 
-  logger.warn("Using memory based session. Don't use memory session in production!");
+  logger.warn(
+    "Using memory based session. Don't use memory session in production!"
+  );
   return new MemoryStore({
     checkPeriod: config.get("session.maxAge"),
   });
@@ -27,8 +29,11 @@ function initDatabaseStore() {
   });
 
   // This listener is necessary otherwise the sourcify process crashes if the database is closed
-  pool.prependListener("error", () => {
-    logger.error("Database connection lost for session pool");
+  pool.prependListener("error", (e) => {
+    logger.error("Database connection lost for session pool", {
+      error: e,
+    });
+    throw new Error("Database connection lost for session pool");
   });
 
   const PostgresqlStore = genFunc(expressSession);
@@ -36,8 +41,8 @@ function initDatabaseStore() {
   logger.info("Using database based session");
   return new PostgresqlStore({
     pool: pool,
-    // Pruning expired sessions every 15 minutes
-    pruneSessionInterval: 900,
+    // Pruning expired sessions every 12 hours
+    pruneSessionInterval: 12 * 60 * 60,
   });
 }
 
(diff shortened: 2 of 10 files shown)


services/server/src/server/session.ts
```diff
@@ -0,0 +1,90 @@
...
+    },
+    store: getSessionStore(),
+  };
+}
+
+export default function getSessionMiddleware() {
+  const sessionMiddleware = session(getSessionOptions());
+  // We need to wrap the sessionMiddleware in a middleware to prevent it from returning all the postgresql errors in the 500 request
+  return (req: Request, res: Response, next: NextFunction) => {
+    sessionMiddleware(req, res, (error) => {
```
marcocastignoli, line 79: I have the doubt that in this way we are going to catch also all the previous middleware that execute `next(err)`

marcocastignoli:
Right now we don't have a way to "fine tune" the configuration ( `process.env` and `config`) to init the `Server` with specific settings for each test. I think we can improve the tests for this feature once we refactor them.

services/server/src/server/session.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (error) {
+        logger.error("Cannot store session in database", {
```
manuelwedler, line 101: Wouldn't this also be run if the memory session is used? The error messages would be a bit misleading in this case.

services/server/src/server/session.ts
manuelwedler: I feel so too, but I am not sure. Did you manage to test it in some way?

manuelwedler: Maybe it would be possible to add a test for this case?

manuelwedler:
I didn't come across the server.js test file before. Now I see what you was talking about 😅

services/server/src/config/default.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
     secret: process.env.SESSION_SECRET || "CHANGE_ME",
     maxAge: 12 * 60 * 60 * 1000, // 12 hrs in millis
     secure: false, // Set Secure in the Set-Cookie header i.e. require https
+    storeType: "memory",
```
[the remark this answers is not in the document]
marcocastignoli: Exactly, the `migration.js` file is the configuration for the `sourcify-server-migration` instance, that is the one used for the data migration from currenty_repositoryV1 to Database. Only one API is used there and it's not session based.

services/server/src/server/session.ts
marcocastignoli: > Maybe it would be possible to add a test for this case?

That's actually a good idea! But is it possible to make the memory session crash on purpose?

services/server/src/config/default.js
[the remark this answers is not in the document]
manuelwedler: Thanks for clarifying!

services/server/src/server/session.ts
manuelwedler: I would say with mocking it should be possible, but with the current state of the server tests it's fine like this.

What I was actually referring to is if we can make a previous middleware fail. So that we test that this middleware does not catch all errors.

marcocastignoli: Ah okok, I'll do it manually and check what happens 👍

marcocastignoli: Ok the previous middleware errors are not catched by this middleware 👍

manuelwedler (approved):
lgtm

services/server/src/server/session.ts
```diff
@@ -0,0 +1,115 @@
...
+function initDatabaseStore() {
+  const pool = new Pool({
+    host: process.env.SOURCIFY_POSTGRES_HOST,
+    database: process.env.SOURCIFY_POSTGRES_DB,
+    user: process.env.SOURCIFY_POSTGRES_USER,
+    password: process.env.SOURCIFY_POSTGRES_PASSWORD,
+    port: parseInt(process.env.SOURCIFY_POSTGRES_PORT || "5432"),
+  });
+
+  // This listener is necessary otherwise the sourcify process crashes if the database is closed
```
kuzdogan, line 29: Isn't this something expected? If the database is closed, that means we can't write anything, and it should really crash?

services/server/src/server/session.ts
```diff
@@ -0,0 +1,115 @@
...
+    logger.error("Database connection lost for session pool");
+  });
+
+  const PostgresqlStore = genFunc(expressSession);
+
+  logger.info("Using database based session");
+  return new PostgresqlStore({
+    pool: pool,
+    // Pruning expired sessions every 15 minutes
+    pruneSessionInterval: 900,
```
kuzdogan, line 40: We should keep this longer like 12 hours.

Also use notation 12 * 60 * 60

services/server/src/server/session.ts
```diff
@@ -0,0 +1,115 @@
...
+        process.env.SOURCIFY_POSTGRES_HOST &&
+        process.env.SOURCIFY_POSTGRES_DB &&
+        process.env.SOURCIFY_POSTGRES_USER &&
+        process.env.SOURCIFY_POSTGRES_PASSWORD
+      ) {
+        return initDatabaseStore();
+      } else {
+        // Log the error but continue with the memory session
+        // This is needed because tests all use the same local-test.js file but not all the tests have the database enabled
+        logger.error(
```
kuzdogan, line 59: Throw here

services/server/src/config/default.js
kuzdogan, line 21: This should be explained to other config files too. 

Also whenever we change anything in configs it must be documented in docs

```suggestion
    storeType: "memory", // Where to save the session info. "memory" is only good for testing and local development. Don't use it in production!
```

services/server/src/server/session.ts
```diff
@@ -0,0 +1,115 @@
...
+import config from "config";
+import session from "express-session";
+
+import logger from "../common/logger";
+import { NextFunction, Request, Response } from "express";
+
+function initMemoryStore() {
+  const MemoryStore = createMemoryStore(expressSession);
+
+  logger.info("Using memory based session");
```
kuzdogan, line 14:
```suggestion
  logger.warn("Using memory based session. Don't use memory session in production!");
```

kuzdogan (changes requested):
Few points. Otherwise LGTM

services/server/src/server/session.ts
marcocastignoli: Crashes with a bad, not understandable error, we can throw inside this listener a better error if that's what we want

services/server/src/server/session.ts
marcocastignoli: I think this is necessary because not all tests run with the database enabled but the `local-test.js` file contains `memory: "database"`. I don't know how to deal with this, ideas?

kuzdogan: My point is, if the user chose "database" and they haven't given the necessary values, it must throw. Souricy can't run with missing info.

For the tests, don't all they have the env vars set?

services/server/src/server/session.ts
kuzdogan: It's ok to add a human readable error message, alongside what we receive from Postgres but 

1) The postgres message should be available. Isn't completely useless?
2) It should indeed crash. My understanding from the error listener below is that Sourcify will not crash. Is it going to crash? (It should crash)

kuzdogan (commented):
Good to go? 

THe only thing I don't see is we didn't update the server README for the new config.

marcocastignoli:
Draft README

```

## Session

By default the server saves session in memory, but it's not a suggested method for the production environment. You can set session
store method in the configuration overriding the property `session.storeType`. The only two supported methods are:

- `memory`

kuzdogan:
> Draft README
> 
> ```
> 
> ## Session
> 
> By default the server saves session in memory, but it's not a suggested method for the production environment. You can set session
> store method in the configuration overriding the property `session.storeType`. The only two supported methods are:
> 
> - `memory`
> ```

Ok my bad. I thought we had the full server config JSON in Readme and we needed to update it but we don't have that but the chains config.json. So this isn't needed

kuzdogan:
Is this done? 

- [ ] exclude session in database backup

Otherwise good to go

marcocastignoli:
It's not done yet, @skylenet is working on it

Issue #1326, opened 2024-04-05
Compilation response exceeds the Lambda function response limits


kuzdogan:
We had a contract case reported that returns with an uninformative "Compiler error". We first thought this was a session issue as i n #1289 but turns out it's something else.

Method: Import from Etherscan
Address: 0xec713053c5c07A2A57185f6ED60FEf98094d65c9
Chain ID: 1 (Mainnet)

The contract was verifying fine locally but failing on staging and prod. so the AWS Lambda is to blame. 

I tried switching the compilation method to lambda locally and sure enough, I could reproduce it. Interestingly the first compilation (that is used for generating the metadata files from Etherscan reqs) works fine. The second one (for verifying) fails. This is because the compiler output is exceeding the [hard limit of AWS Lambda responses (6 MB)](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html). The first compilation only [outputs the metadata](https://github.com/ethereum/sourcify/blob/accc7f023a151735cafccda24bb47c2dfd052462/services/server/src/server/controllers/verification/etherscan/etherscan.common.ts#L57-L62) whereas the second compilation [is a full output](https://github.com/ethereum/sourcify/blob/accc7f023a151735cafccda24bb47c2dfd052462/packages/lib-sourcify/src/lib/CheckedContract.ts#L700-L715). 

Lambda logs: https://eu-north-1.console.aws.amazon.com/cloudwatch/home?region=eu-north-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fcompile/log-events/2024$252F04$252F05$252F$255B$2524LATEST$255Dc658e48fd88c415194681199b19ce7e4
```
2024-04-05T12:34:08.874+02:00	START RequestId: e4ea372e-c8cd-4158-87b7-fd035e90ee0d Version: $LATEST
2024-04-05T12:34:08.888+02:00	2024-04-05T10:34:08.888Z e4ea372e-c8cd-4158-87b7-fd035e90ee0d DEBUG Found solc 0.8.18+commit.87f61d96 with platform linux-amd64 at /tmp/solc-repo/solc-linux-amd64-v0.8.18+commit.87f61d96
[...]

PR #1340, opened 2024-04-10
Stream lambda function compiler responses

manuelwedler:
See #1326

I didn't implement any chunking myself. Instead I used the `pipeline` function because the docs recommend to do so. AWS's `responseStream` will just take care of chunking it.

Note that that streamed responses also have a limit of 20 MB (usual payload response was 6 MB). Maybe it could make sense to fallback to local compilation if an error is received from the lambda function?

TODO:
- [x] Add a dashboard to Grafana when fallback is used

### DIFF
--- a/serverless/compiler-lambda/index.js
+++ b/serverless/compiler-lambda/index.js
@@ -5,6 +5,8 @@ const { exec, spawnSync } = require("child_process");
 const solc = require("solc");
 const fetch = (...args) =>
   import("node-fetch").then(({ default: fetch }) => fetch(...args));
+const pipeline = require("util").promisify(require("stream").pipeline);
+const { Blob } = require('buffer');
 
 async function fetchWithTimeout(resource, options = {}) {
   const { timeout = 30000 } = options;
@@ -298,23 +300,22 @@ function asyncExecSolc(inputStringified, solcPath) {
   });
 }
 
-exports.handler = async (event) => {
+exports.handler = awslambda.streamifyResponse(async (event, responseStream, _context) => {
+  let output;
   try {
-    return {
-      success: true,
-      body: await useCompiler(
-        event.version,
-        event.solcJsonInput,
-        event.forceEmscripten
-      ),
-    };
+    output = await useCompiler(
+      event.version,
+      event.solcJsonInput,
+      event.forceEmscripten
+    );
   } catch (e) {
-    return {
-      success: true,
-      body: e.message,
-    };
+    output = { error: e.message };
   }
-};
+  console.debug("Compilation output: ", output);
+
+  const outputBlob = new Blob([JSON.stringify(output)]);
+  await pipeline(outputBlob.stream(), responseStream);
+});
 
 /* exports
   .handler({
--- a/serverless/compiler-lambda/index.js
+++ b/serverless/compiler-lambda/index.js
@@ -6,7 +6,7 @@ const solc = require("solc");
 const fetch = (...args) =>
   import("node-fetch").then(({ default: fetch }) => fetch(...args));
 const pipeline = require("util").promisify(require("stream").pipeline);
-const { Blob } = require('buffer');
+const { Blob } = require("buffer");
 
 async function fetchWithTimeout(resource, options = {}) {
   const { timeout = 30000 } = options;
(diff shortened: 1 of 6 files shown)


marcocastignoli:
> Note that that streamed responses also have a limit of 20 MB (usual payload response was 6 MB). Maybe it could make sense to fallback to local compilation if an error is received from the lambda function?

I like this idea!

services/server/src/server/services/compiler/lambda/SolcLambda.ts
```diff
@@ -44,45 +44,56 @@ export class SolcLambda implements ISolidityCompiler {
...
+        );
+      } else if (event.PayloadChunk?.Payload) {
+        streamResult += Buffer.from(event.PayloadChunk.Payload).toString(
+          "utf8"
+        );
+      }
+    }
+    logger.silly("Received stream response", { streamResult });
+
+    const output = JSON.parse(streamResult);
```
marcocastignoli, line 86: Maybe it could make sense to try catch this and in the case of an error `logger.error` the `streamResult`?

marcocastignoli (commented):
Just a small doubt regarding handling a JSON.parse potential error

services/server/src/server/services/compiler/lambda-with-fallback/SolcLambdaWithLocalFallback.ts
```diff
@@ -0,0 +1,38 @@
...
+  ): Promise<CompilerOutput> {
+    let compilerOutput: CompilerOutput;
+    try {
+      compilerOutput = await this.solcLambda.compile(
+        version,
+        solcJsonInput,
+        forceEmscripten
+      );
+    } catch (e) {
+      logger.error(
```
marcocastignoli, line 27: I think that here we need to switch on a specific error not all the errors

manuelwedler: (I pushed the new commit before I saw your comment update)

The problem with this is that I cannot really find out which error code is given by AWS in the case the response is too big. I was checking the docs and they were not very helpful (https://docs.aws.amazon.com/lambda/latest/api/API_InvokeWithResponseStreamCompleteEvent.html). 

We could also try to find a contract whose compiler output is bigger than the response stream limit of 20 MB. I'm just not sure how to go about that.

marcocastignoli: I'm thinking... Maybe on the lambda side you can find the output size and if it's more than 20mb send a specific response. Could this work?

serverless/compiler-lambda/index.js
```diff
@@ -298,23 +300,22 @@ function asyncExecSolc(inputStringified, solcPath) {
...
-    return {
-      success: true,
-      body: e.message,
-    };
+    output = { error: e.message };
   }
-};
+  console.debug("Compilation output: ", output);
+
+  const outputBlob = new Blob([JSON.stringify(output)]);
```
marcocastignoli, line 316: Continuing this comment: https://github.com/ethereum/sourcify/pull/1340#discussion_r1560488122

like this:
```
if (weight(outputBlob) > 20MB ) {
  output = { error: "specific error for this case" }
  return or throw
}
```

manuelwedler: Ah nice idea! That should work out! Thanks for the hint

services/server/src/server/services/compiler/lambda-with-fallback/SolcLambdaWithLocalFallback.ts
```diff
@@ -0,0 +1,42 @@
...
+    let compilerOutput: CompilerOutput;
+    try {
+      compilerOutput = await this.solcLambda.compile(
+        version,
+        solcJsonInput,
+        forceEmscripten
+      );
+    } catch (e) {
+      if (e instanceof LambdaResponseLimitExceeded) {
+        logger.error(
```
kuzdogan, line 28: warning, not an error

manuelwedler:
I created a dashboard for visualizing the number of times the fallback to local compilation is invoked (the dashboard is empty since the changes are not released yet). I expect that this fallback will be barely used, but let's see in Grafana.

marcocastignoli (approved):
👍👍

Issue #1318, opened 2024-04-02
Research and plan: Refactoring the verification flow for inconsistencies


kuzdogan:
Looking at the logs I noticed a contract returning multiple errors in all StorageServices: `Error storing to RepositoryV1` `...RepositoryV2` `...AllianceDatabase` `...SourcifyDatabase`. 

The contract in question is `0xbf1B6429214BcbDFF065697aFC6A4EC2ba392078` on chain `10`. This was an Import from Etherscan request with id `8419c051fafabc40af058b9b8d34967a`.

I wasn't able to reproduce the same logs initially with the session API locally. Upon trying the non-session API I was able to reproduce the same logs. 

Turns out this is because the verification through the session and non-session take different flows. Namely the session flow goes to `verifyContractsInSession` function which checks the `runtimeMatch` and `creationMatch` and only stores the contracts if one of them exists: https://github.com/ethereum/sourcify/blob/5295fa3c1336a6901856a060a8122646fa5be055/services/server/src/server/controllers/verification/verification.common.ts#L457-L459

In contrast the stateless flow first calls `services.verification.verifyDeployed` and then directly calls `storeMatch` without checking the `runtime` and `creationMatch` https://github.com/ethereum/sourcify/blob/5295fa3c1336a6901856a060a8122646fa5be055/services/server/src/server/controllers/verification/etherscan/stateless/etherscan.stateless.handlers.ts#L42-L43

IMO the storing should only be done in the verification service and not dispersed throughout the codebase. In principal, all the verification related things should be done in the Verif. Service. A similar thing with all the `extra-file-input-bug` related code which as well spread across the codebase. 

Example:
- When running verification for the contract `0xF873413AA072BFcdD97e81f74B190FFFB9110f42` on Base Sepolia (84532) via Import from Etherscan, I could observe the contract is compiled 3 times. My guess: one for etherscan to metadata conversion, one for verification, and final one for CBORAuxdata position finding.

Other Issues:
- It seems the creationTx is fetched multiple times throughout a contract verification. This should be enough once
```[tasklist]
### Tasks
- [ ] Make sure the `dryRun` param is accepted in all endpoints
```

marcocastignoli:
> IMO the storing should only be done in the verification service and not dispersed throughout the codebase. In principal, all the verification related things should be done in the Verif. Service. A similar thing with all the extra-file-input-bug related code which as well spread across the codebase.

[...]

PR #1343, opened 2024-04-11
Add `dryrun` query param to enable verification preview

### LINKED ISSUE #1319: Support for verification preview
Our verification UI flow would need to call the `/session/verify-checked` and `/session/input-files` API endpoints without actually persisting the result of the verification in the repository. This is required in order to obtain an exact preview of what the result of the verification is going to be for a given contract (from https://github.com/hashgraph/hedera-sourcify/issues/90)

Here's our implementation by introducing a `dryrun` query parameter
- https://github.com/hashgraph/hedera-sourcify/pull/92

Would you consider including this feature?

acuarica:
This PR adds a new query parameters to enable verification preview.

Fixes #1319.

### DIFF
--- a/services/server/src/server/controllers/verification/verification.common.ts
+++ b/services/server/src/server/controllers/verification/verification.common.ts
@@ -345,7 +345,7 @@ export const verifyContractsInSession = async (
   session: Session,
   verificationService: IVerificationService,
   storageService: StorageService,
-  dryRun: boolean = false,
+  dryRun: boolean = false
 ): Promise<void> => {
   logger.debug("verifyContractsInSession", {
     sessionId: session.id,
@@ -456,7 +456,13 @@ export const verifyContractsInSession = async (
     contractWrapper.status = getMatchStatus(match) || "error";
     contractWrapper.statusMessage = match.message;
     contractWrapper.storageTimestamp = match.storageTimestamp;
-    if ((match.runtimeMatch || match.creationMatch) && !dryRun) {
+    if (dryRun) {
+      logger.info("dryRun verification", {
+        sessionId: session.id,
+      });
+      return;
+    }
+    if (match.runtimeMatch || match.creationMatch) {
       await storageService.storeMatch(checkedContract, match);
     }
   }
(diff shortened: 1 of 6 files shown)


acuarica:
Hi @kuzdogan @marcocastignoli, I need to yet include tests for this feature. What is the test flow to make this ready for review?

Running `npm run test-local` fails with a bunch of errors, is that expected?

marcocastignoli:
> Hi @kuzdogan @marcocastignoli, I need to yet include tests for this feature. What is the test flow to make this ready for review?
> 
> Running `npm run test-local` fails with a bunch of errors, is that expected?

I'll take car of it, thanks for the PR 👍

marcocastignoli:
- [ ] Add logger.info for dryRun

acuarica:
> I'll take car of it, thanks for the PR

Great, thanks! Let me know if there is something I could do to merge this. Happy to contribute here!

services/server/src/server/controllers/verification/verification.common.ts
```diff
@@ -455,6 +456,12 @@ export const verifyContractsInSession = async (
     contractWrapper.status = getMatchStatus(match) || "error";
     contractWrapper.statusMessage = match.message;
     contractWrapper.storageTimestamp = match.storageTimestamp;
+    if (dryRun) {
+      logger.info("dryRun verification", {
+        sessionId: session.id,
```
marcocastignoli, line 461: @kuzdogan for session cases, is it enough to log only the sessionId? All the other information are available in previous logs, right?

kuzdogan:
@acuarica Do you only need this in the session endpoint? Because your changes only touch that enpoint and also does not change the UI

services/server/src/server/controllers/verification/verification.common.ts
kuzdogan: Hmm it doesn't seem to be the case. You can see by running it once. But generally just log more things than less

acuarica:
> Do you only need this in the session endpoint? Because your changes only touch that enpoint _[...]_

Yes, only here, thanks.

> and also does not change the UI

Indeed. For Hedera verification, we decided to directly consume `server` and `repository` images. Given that `ui` changes might be too Hedera specific, we decided to keep the `ui` fork with our custom modifications.

In any case, this `dryrun` feature is used the contract verification within https://hashscan.io.

kuzdogan (approved):
LGTM for this specific case. 

We should add `dryRun` to  all other endpoints when doing #1318

PR #1374, opened 2024-04-26
Typescript support for server tests

### LINKED ISSUE #1373: Add TypeScript to server tests
I think we should strictly enforce the usage of TypeScript everywhere in the project. Once added to the server tests, I think TypeScript makes writing tests easier and more maintainable.

manuelwedler:
Closes #1373 

My approach was to just fix all the errors the compiler gave me. There are still a lot of implicitly `any`-typed values. So I think this is just a start. We can further improve the typing while work on the test refactoring.

I also removed some unused code btw.

To make the CI run successfully, I fixed the tests in the contract-call-decoder package.


services/server/test/chains/chain-tests.spec.ts
```diff
@@ -67,18 +76,6 @@ describe("Test Supported Chains", function () {
...
-    const errorBody = currentResponse && currentResponse.body;
-    if (this.currentTest.state === "failed" && errorBody) {
-      console.log(
-        "Server response of failed test " + this.currentTest.title + ":"
-      );
-      console.log(errorBody);
-    }
-    currentResponse = null;
-  });
-
```
kuzdogan, line 81: Why are we removing this? This was helpful in seeing the server's error response. Or did we replace it? Otherwise mocha only gives `Expected status: 200 Actual: 500`

services/server/test/chains/deployContracts.ts
kuzdogan, line 1: I think it's better to leave this as simple JS. This was just a CLI tool to deploy the standard Sourcify test contract for new chains.

services/server/test/server.spec.ts
```diff
@@ -163,18 +174,6 @@ describe("Server", function () {
...
-    const errorBody = currentResponse && currentResponse.body;
-    if (this.currentTest.state === "failed" && errorBody) {
-      console.log(
-        "Server response of failed test " + this.currentTest.title + ":"
-      );
-      console.log(errorBody);
-    }
-    currentResponse = null;
-  });
-
```
kuzdogan, line 177: Same. See the comment on chain-tests.spec.ts

services/server/test/server.spec.ts
```diff
@@ -2283,83 +2322,69 @@ describe("Server", function () {
...
     it(verifierAllianceTestLibrariesManuallyLinked._comment, async () => {
       await verifierAllianceTest(
         server,
-        chai,
         storageService,
         localSigner,
         defaultContractChain,
         verifierAllianceTestLibrariesManuallyLinked
       );
     });
```
kuzdogan, line 2334: Not necessarily in this PR but the part below here can be nicely refactored. A lot of repetitive code here. Just import everyting and loop over it.

kuzdogan (changes requested):
Couple points. Otherwise looks good

services/server/test/server.spec.ts
kuzdogan: Yes. You can always do `--grep` in mocha to run a particular test or skip manually. It's ugly and really cumbersome IMO to write the long variable names and the whole test function for each test.

services/server/test/chains/chain-tests.spec.ts
manuelwedler: I removed it because the `currentResponse` variable was never set. It was commented in all places. I can bring it back but then we should also uncomment where the variable is assigned.

services/server/test/chains/deployContracts.ts
manuelwedler: The issue is that it imports a function from the helper module which is TS now. Shouldn't it be easy to just run it with `ts-node`?

services/server/test/server.spec.ts
manuelwedler: In my opinion, it is fine to have some repetitive code in tests (at least to some degree). The reason is that it makes the tests self-contained. And self-contained tests are more understandable and maintainable. I think creating tests in loops contradicts this principle.

In this specific case, I would rather write the content of `_comment` directly into the `it` call to achieve this. The parameters of the `verifierAllianceTest` should be reduceable with a proper server configuration setup for the tests, because they are always the same (only the testcase parameter differs).

Anyway, I would rather leave refactoring this part (regardless which way we go) in another PR.

kuzdogan: Closing this and I'm fine either way, not a strong opinion.

services/server/test/chains/chain-tests.spec.ts
kuzdogan: Hmm ok just remove it then. Now that we have proper logging, this is probably not needed

kuzdogan (approved):
All good after the discussion. Just please update the docs for `deployContract.ts`

services/server/test/chains/deployContracts.ts
manuelwedler: Alright, updated the docs:
https://docs.sourcify.dev/docs/chain-support/

PR #1390, opened 2024-05-08
Use Sourcify database as default StorageService

### LINKED ISSUE #1380: Replace the Server endpoints with the DB

marcocastignoli:
This PR closes #1380 

With this PR, all server endpoints that previously fetched information from RepositoryV1 will now retrieve data from the database and RepositoryV2.

This PR also sets the database as the source of truth when checking the existence of a match.

## Todo
- [x] Limit `/files/contracts` for 200 contracts, implement new paginated APIs
- [x] Fix CI tests (locally they are successful 😮‍💨)
- [x] https://github.com/ethereum/sourcify/pull/1390/files#r1594247371
- [x] Create tests for paginated APIs


services/server/src/server/controllers/repository/repository.routes.ts
```diff
@@ -60,6 +51,14 @@ router
   .route("/check-all-by-addresses")
   .get(safeHandler(checkAllByChainAndAddressEndpoint));
 
+router
```
marcocastignoli, line 54: This is the new alternative to the removed repository static files route https://github.com/ethereum/sourcify/pull/1390/files#diff-c71290f3b0c3ad7664c772a606bb8ec512d83d3f0d41bb73e792cfcbcac569e6L269-L274

services/server/src/server/services/utils/database-util.ts
```diff
@@ -142,9 +142,15 @@ export async function getSourcifyMatchByChainAddress(
   return await pool.query(
     `
       SELECT
-        sourcify_matches.*
+        sourcify_matches.*,
+        sourcify_matches.creation_match as creation_match_status,
+        sourcify_matches.runtime_match as runtime_match_status,
+        verified_contracts.*,
+        contract_deployments.*,
+        compiled_contracts.*
```
marcocastignoli, line 150: TODO: select only the needed columns

services/server/src/server/services/utils/repository-util.ts
marcocastignoli, line 1: In this file we had only types and interfaces that I moved to `services/server/src/server/types.ts`

services/server/test/server.spec.ts
```diff
@@ -2054,15 +2054,18 @@ describe("Server", async function () {
...
         // The path from the server response must be translated
         const translatedContractObject = fetchedContractFiles.find(
-          (obj: any) =>
-            obj.path ===
-            path.join(
-              contractSavedPath,
-              "sources",
-              pathTranslationJSON[originalPath]
-            )
+          (obj: any) => obj.path === relativeFilePath
```
marcocastignoli, line 2068: I had to change this test because now a relative path is expected, not an absolute path

marcocastignoli: I don't know what to do with this test. Both Database and RepositoryV2 don't store path translations, so testing it like this really doesn't make sense. Should I implement a new test to check specifically the files stored on RepositoryV1 or I simply get rid of it?

From my pov RepositoryV1 is going to be removed soon, we just need to find a way to run repo.sourcify.dev with RepositoryV2 or Database. After we remove it, testing path translations is going to be useless.

@kuzdogan @manuelwedler feedback?

services/server/test/server.spec.ts
```diff
@@ -2149,6 +2151,44 @@ describe("Server", async function () {
...
+        .request(server.app)
+        .get(`/files/contracts/any/${defaultContractChain}?limit=5`);
+      chai.expect(res1.body.pagination.currentPage).to.equal(0);
+      chai.expect(res1.body.pagination.hasNextPage).to.equal(false);
+      chai.expect(res1.body.pagination.hasPreviousPage).to.equal(false);
+      chai.expect(res1.body.pagination.resultsCurrentPage).to.equal(5);
+      chai.expect(res1.body.pagination.resultsPerPage).to.equal(5);
+      chai.expect(res1.body.pagination.totalPages).to.equal(1);
+      chai.expect(res1.body.pagination.totalResults).to.equal(5);
+    });
```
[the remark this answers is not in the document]
manuelwedler: Instead of calling `to.equal` for every property, I think you can call  [`deepEqual`](https://www.chaijs.com/api/assert/#method_deepequal) for `res1.body.pagination`.

services/server/test/server.spec.ts
manuelwedler: I'd say, in this case, don't make the extra effort for testing something that is going to be removed :)

marcocastignoli: Actually looking at this test again... If we want to keep path sanitization, then it makes sense to keep this test. But do we want to keep path sanitization for the database?

services/server/src/server/controllers/repository/get-contract-addresses-all.stateless.paths.yaml
```diff
@@ -3,8 +3,8 @@ openapi: "3.0.0"
 paths:
   /files/contracts/{chain}:
     get:
-      summary: Get all contract addresses verified on a chain (full or partial match)
-      description: Returns all verified contracts from the repository for the desired chain. Searches for full and partial matches.
+      summary: Get the first 200 contract addresses verified on a chain (full or partial match)
```
kuzdogan, line 6:
```suggestion
      summary: (Deprecated) Get the first 200 contract addresses verified on a chain (full or partial match)
```

services/server/src/server/controllers/repository/get-contract-addresses-all.stateless.paths.yaml
```diff
@@ -3,8 +3,8 @@ openapi: "3.0.0"
...
+      description: Returns the first 200 verified contracts from the repository for the desired chain. Searches for full and partial matches.
```
kuzdogan, line 7:
```suggestion
      description: Returns the first 200 verified contracts from the repository for the desired chain. Searches for full and partial matches. This endpoint is deprecated and only retuns 200 addresses. Use `/files/contracts/any/` and `/files/contracts/full/` instead
```

services/server/src/server/controllers/repository/get-contract-addresses-all.stateless.paths.yaml
kuzdogan, line 1: Add the `500` (or something else?) when more than 200 addresses exist

services/server/src/server/services/utils/database-util.ts
```diff
@@ -567,3 +571,64 @@ export function normalizeRecompiledBytecodes(
...
+        SUM(CASE WHEN sourcify_matches.creation_match != 'perfect' AND sourcify_matches.runtime_match != 'perfect' THEN 1 ELSE 0 END) AS partial_total
+    FROM sourcify_matches
+    JOIN verified_contracts ON verified_contracts.id = sourcify_matches.verified_contract_id
+    JOIN contract_deployments ON 
+        contract_deployments.id = verified_contracts.deployment_id
+        AND contract_deployments.chain_id = $1
+    GROUP BY contract_deployments.chain_id;
+    `,
+    [chain]
+  );
```
kuzdogan, line 590: In this case if a user asks a chain's contracts that has couple million addresses, it sends the query. This would take a lot of time no? I thought we were going to set the 200 limit **inside the query** and if the response hits 200 contracts, then we respond on server an error.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -88,15 +103,295 @@ export class SourcifyDatabaseService
...
+
+    const fullTotal = parseInt(matchAddressesCountResult.rows[0].full_total);
+    const partialTotal = parseInt(
+      matchAddressesCountResult.rows[0].partial_total
+    );
+    if (
+      fullTotal > MAX_RETURNED_CONTRACTS_BY_GETCONTRACTS ||
+      partialTotal > MAX_RETURNED_CONTRACTS_BY_GETCONTRACTS
+    ) {
+      logger.error(
```
kuzdogan, line 140:
```suggestion
      logger.info(
```

services/server/test/server.spec.ts
kuzdogan: We can remove the path sanitization for the database but for the repoV2 this leaves me thinking if store the sources at `path/to/the/source/with/many/folders/0xkeccak.sol` or just `sources/0xkeccak.sol` because for the former there still can be weird characters in the path and we actually need to store in a standard path.

kuzdogan (changes requested):

services/server/src/server/services/utils/database-util.ts
marcocastignoli: But this is just for calculating the total, we need a way to let the user know the total of contracts in a chain. Maybe we can optimize this by caching the result

services/server/test/server.spec.ts
marcocastignoli: RepositoryV2 stores everything inside the first level of `/sources` 

e.g.

/sources/0x11111.sol
/sources/0x11112.sol
...

services/server/src/server/services/utils/database-util.ts
kuzdogan: Do we need to? Isn't the only important thing if there are more contracts than N? If the DB still needs to count 1M+ rows then it defeats the purpose

services/server/test/server.spec.ts
kuzdogan: Ok nice then it's not a problem and we can remove all the path sanitization related code + tests

kuzdogan (commented):
Just added a comment about the path sanitization

services/server/test/server.spec.ts
marcocastignoli: We cannot remove "all the path sanitization related code" because it's still used in RepositoryV1, and right now RepositoryV1 is used in repo.sourcify.dev

Should this PR also make repo.sourcify.dev point to RepositoryV2 and remove RepositoryV1 completely?

kuzdogan: Ok right... No let's not mix them both.

I've created a new issue to replace the repo.sourcify.dev completely and under that I've added the subtask to remove the path sanitization related code. https://github.com/ethereum/sourcify/issues/1398

With that we can dismiss this here and move on.

kuzdogan (approved):

PR #1396, opened 2024-05-15
Implement Sourcify to VerA script

marcocastignoli:
This script is needed to push all verified_contracts from Sourcify to VerA database. VerA is a live database in which other entities are writing while the script is running.

0. CURRENT_VERIFIED_CONTRACT is a variable stored in a permanent file, default 1.
1. Extract N verified_contracts starting from CURRENT_VERIFIED_CONTRACT
2. For each verified_contract:
   a. Update the CURRENT_VERIFIED_CONTRACT counter
   b. Get all information from verified_contracts
   c. Get all the FK information
   d. Insert all the verified_contracts dependencies (compiled_contracts,contract_deployments,contracts,code) and verified_contracts using the new ids as FKs


services/verifier-alliance/main.mjs
```diff
@@ -0,0 +1,240 @@
...
+      `
+            SELECT * FROM verified_contracts
+            WHERE id >= $1
+            ORDER BY id ASC
+            LIMIT $2
+        `,
+      [CURRENT_VERIFIED_CONTRACT, N]
+    );
+
+    for (const contract of verifiedContracts) {
```
manuelwedler, line 72: So this script only process N contracts when it is run. This means the script should be run repeatedly to transfer all missing contracts. When would you stop running it? Why not iterate within the script until a stopping condition is reached?

services/verifier-alliance/main.mjs
```diff
@@ -0,0 +1,240 @@
...
+      const {
+        id: verifiedContractId,
+        deployment_id,
+        compilation_id,
+      } = contract;
+
+      // Update the CURRENT_VERIFIED_CONTRACT counter
+      CURRENT_VERIFIED_CONTRACT = verifiedContractId;
+      fs.writeFileSync(
```
manuelwedler, line 81: Wouldn't it be safer to only update this counter once we are sure the contract was written to the VerA database? If something goes wrong later, we might skip a contract, no?

services/verifier-alliance/main.mjs
marcocastignoli: Uh, I forgot that part 🤣 I focused too much on the internal part that I forgot to conclude the loop

services/verifier-alliance/main.mjs
marcocastignoli: the code is 

```
            WHERE id >= $1
            ORDER BY id ASC
            LIMIT $2
        `,
      [CURRENT_VERIFIED_CONTRACT, N]
```

so it starts from the last CURRENT_VERIFIED_CONTRACT. If the scripts breaks while synching a verified_contract then it starts again from that contract.

manuelwedler: Ah, I misunderstood it. Looks fine then :)

manuelwedler (changes requested):
There is the missing loop, but other than that looks good to me

services/verifier-alliance/main.mjs
marcocastignoli: Ok done @manuelwedler

manuelwedler (approved):
Looks good!

Issue #1405, opened 2024-05-24
Update contract_deployments after storing a better match with creatorTx information


marcocastignoli:
In the database there are some matches with creation_match={partial|perfect} but without any creationTx information stored in contract_deployments. This happened because when we update a contract with a better match with creatorTx information we do not update contract_deployments.

Use the following query on production to see the problem:

```sql
  SELECT
        verified_contracts.*
      FROM verified_contracts
      JOIN contract_deployments ON contract_deployments.id = verified_contracts.deployment_id
      WHERE 1=1
        AND contract_deployments.chain_id = 11155111
        AND contract_deployments.address = decode('977097E1C0054252E1CE8263E0FBF673432D7322', 'hex')
```

The results are two `verified_contracts`. The first one has no creation match, while the second one does. The second contract is the new “best” match, created using creation transaction information. Both contracts point to the same `contract_deployments` that was initially created by the first contract (that didn't have creation transaction information).

PR #1408, opened 2024-05-28
Update contract_deployments table after storing a match that previously didn't have contract creation information

marcocastignoli:
see #1405 

### TODO
- [ ] Add test (wait for #1403)

### DIFF
--- a/services/server/src/server/services/utils/database-util.ts
+++ b/services/server/src/server/services/utils/database-util.ts
@@ -17,10 +17,12 @@ export namespace Tables {
     bytecode: Buffer;
   }
   export interface Contract {
+    id?: string;
     creation_bytecode_hash?: Hash;
     runtime_bytecode_hash: Hash;
   }
   export interface ContractDeployment {
+    id?: string;
     chain_id: string;
     address: Buffer;
     transaction_hash: Buffer;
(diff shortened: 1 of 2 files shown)


services/server/src/server/services/utils/database-util.ts
```diff
@@ -632,3 +636,38 @@ export async function getSourcifyMatchAddressesByChainAndMatch(
     [chain, page * paginationSize, paginationSize]
   );
 }
+
+export async function updateContract(
+  pool: Pool,
+  { id, creation_bytecode_hash, runtime_bytecode_hash }: Tables.Contract
+) {
```
manuelwedler, line 643: Maybe add a type annotation here for the returning value. Should be `Tables.Contract`.

manuelwedler: After a closer look, I saw this function is not used. Any reason for adding it here? Maybe for the fixing script?

services/server/src/server/services/utils/database-util.ts
```diff
@@ -632,3 +636,38 @@ export async function getSourcifyMatchAddressesByChainAndMatch(
...
+  pool: Pool,
+  {
+    id,
+    transaction_hash,
+    block_number,
+    txindex,
+    deployer,
+    contract_id,
+  }: Omit<Tables.ContractDeployment, "chain_id" | "address">
+) {
```
manuelwedler, line 660: Here, a return type would also be nice, since it cannot be inferred from the `query` function.

manuelwedler (commented):
Good to have this fixed :) I just have some minor comments, looks good other than that

services/server/src/server/services/utils/database-util.ts
marcocastignoli: Maybe we can fix all of the return types in another PR, because it is something that has to be done everywhere

manuelwedler: Okay that's fine with me. But we should create an issue then

manuelwedler (approved):

Issue #1415, opened 2024-05-31
Implementing `/files/contracts/partial` endpoint


kuzdogan:
Right now we only have enpoints for `/files/contract/` (full) and `/files/contract/any`. 

We are missing the "partial" only endpoint.

In the making we should also add:
- [ ] **To all of the endpoints** Add explicit ordering of the contracts in the order they are added (id's are incremental). The default should be earliest first so the pages remain consistent
- [ ] Add an option to retrieve latest first (default is earliest first)

PR #1416, opened 2024-06-03
Add `/files/contracts/partial/{chain}` endpoint and explicit ordering, resolve SQL NULL comparison bug

kuzdogan:
This PR was intended to add the `/files/contracts/partial/{chain}` endpoint that is missing to list all contracts under a certain chain.

However during the process, specifically when adding proper tests, I've noticed the `partial` and `any` endpoints not working as expected, and some other potential issues explained below for further investigation. 

TLDR;
- Adding the `/files/contracts/partial/{chain}` endpoint as specified in #1415 with an option to sort `asc` and `desc`
- Adding tests
- [SQL NULL comparison bug](#sql-null-comparison-bug)
- Potential issues:
  - [Deduplication not working in `compiled_contracts`](#deduplication-not-working-in-compiled_contracts)
  - [`matchType` not used in `database-utils.ts`](#matchtype-not-used-in-database-utils-ts)

## SQL NULL comparison bug
This was due to the queries being used in `countSourcifyMatchAddresses` and `getSourcifyMatchAddressesByChainAndMatch`. 

Both SQL queries count the number of perfect matches with `creation_match = 'perfect' OR runtime_match = 'perfect'` and the partial matches with the logical inverse `creation_match != 'perfect' AND runtime_match != 'perfect'`. However this does not work as expected in SQL because the `creation_match` field in our case can be `NULL` and any comparison with `NULL` resolves to `NULL`, and not to `true` as expected. 

This can also be seen by comparing the contract numbers in `stats.json` vs the API output:

`/stats.json` for Mainnet:
```js
  "1": {
    ...
    "full_match": 131482,
    "partial_match": 529877,
  },
  // total: 661,359
```

`/files/contracts/full/1/`:
```
"pagination": {
    ...
    "totalResults": 132056
  }
```

`/files/contracts/any/1/`:
```
"pagination": {
    ...
    "totalResults": 273320
  }
```

The difference (i.e. partial count) gives 141,264 whereas from stats.json we've got 529,877. 

The bug can be resolved by adding a `COALESCE(creation_match, '')` to make the `NULL` values `''` for comparisons. 

# Potential issues

## Deduplication not working in `compiled_contracts`

When debugging the tests I've noticed the identical test contracts are getting added to the `compiled_contracts` table multiple times. I'd have expected them to be deduplicated because the table has a unique key property in the schema definition:
https://github.com/ethereum/sourcify/blob/653952de199123406cea9c3ebda4eda88cc0652a/services/database/migrations/20231109160022-create-alliance-schema.js#L204

However during testing the contracts get added multiple times:
```SQL
SELECT compiler, "language", creation_code_hash, runtime_code_hash FROM public.compiled_contracts x
```
image

Am I missing something or is there something wrong?

### To reproduce

1) add a breakpoint at
https://github.com/ethereum/sourcify/blob/653952de199123406cea9c3ebda4eda88cc0652a/services/server/test/integration/repository-handlers/files.spec.ts#L56

2) Add an `.only` modifier to describe:
https://github.com/ethereum/sourcify/blob/653952de199123406cea9c3ebda4eda88cc0652a/services/server/test/integration/repository-handlers/files.spec.ts#L43

3) run the `Mocha - Server Integration`, deploy and verify multiple contracts, and connect to the test database at `localhost:5431`.

### `matchType` in `database_utils.ts` 

While looking at the `database_utils.ts` code I've noticed the `matchType` variable is not used in functions:
https://github.com/ethereum/sourcify/blob/eeec8244c25db361f2f94981268118b188c1f53a/services/server/src/server/services/utils/database-util.ts#L429

https://github.com/ethereum/sourcify/blob/eeec8244c25db361f2f94981268118b188c1f53a/services/server/src/server/services/utils/database-util.ts#L452

But it seems these functions themselves are not used at all. Can they be removed?

### DIFF
--- a/services/server/test/integration/repository-handlers/files.spec.ts
+++ b/services/server/test/integration/repository-handlers/files.spec.ts
@@ -40,7 +40,7 @@ describe("Verify repository endpoints", function () {
     chai.expect(res4.body.full).has.a.lengthOf(1);
   });
 
-  describe.only(`Pagination in /files/contracts/{full|any|partial}/${chainFixture.chainId}`, async function () {
+  describe(`Pagination in /files/contracts/{full|any|partial}/${chainFixture.chainId}`, async function () {
     const endpointMatchTypes = ["full", "any", "partial"];
     for (const endpointMatchType of endpointMatchTypes) {
       it(`should handle pagination in /files/contracts/${endpointMatchType}/${chainFixture.chainId}`, async function () {
--- a/services/server/test/integration/repository-handlers/files.spec.ts
+++ b/services/server/test/integration/repository-handlers/files.spec.ts
@@ -40,64 +40,84 @@ describe("Verify repository endpoints", function () {
     chai.expect(res4.body.full).has.a.lengthOf(1);
   });
 
-  it("should handle pagination in /files/contracts/any/{chain}", async function () {
-    // Deploy 5 contracts
-    const contractAddresses: string[] = [];
-    for (let i = 0; i < 5; i++) {
-      const address = await deployAndVerifyContract(
-        chai,
-        chainFixture,
-        serverFixture
-      );
-      contractAddresses.push(address);
+  describe.only(`Pagination in /files/contracts/{full|any|partial}/${chainFixture.chainId}`, async function () {
+    const endpointMatchTypes = ["full", "any", "partial"];
+    for (const endpointMatchType of endpointMatchTypes) {
+      it(`should handle pagination in /files/contracts/${endpointMatchType}/${chainFixture.chainId}`, async function () {
+        const contractAddresses: string[] = [];
+
+        // Deploy 5 contracts
+        for (let i = 0; i < 5; i++) {
+          // Deploy partial matching contract if endpoint is partial or choose randomly if endpointMachtype is any. 'any' endpoint results should be consistent regardless.
+          const shouldDeployPartial =
+            endpointMatchType === "partial" ||
+            (endpointMatchType === "any" && Math.random() > 0.5);
+
+          const address = await deployAndVerifyContract(
+            chai,
+            chainFixture,
+            serverFixture,
+            shouldDeployPartial
+          );
+          contractAddresses.push(address);
+        }
+
+        // Test pagination
+        const res0 = await chai
+          .request(serverFixture.server.app)
+          .get(
+            `/files/contracts/${endpointMatchType}/${chainFixture.chainId}?page=1&limit=2`
+          );
+        chai.expect(res0.body.pagination).to.deep.equal({
+          currentPage: 1,
+          hasNextPage: true,
+          hasPreviousPage: true,
+          resultsCurrentPage: 2,
+          resultsPerPage: 2,
+          totalPages: 3,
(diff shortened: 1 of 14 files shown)


marcocastignoli:
> Deduplication not working in compiled_contracts

I cannot understand why this happens. Maybe the unique constraint doesn't work with a NULL column?

> But it seems these functions themselves are not used at all. Can they be removed?

Right now they are not used. If we need them for the rebuilding the database we can implement them again, you can remove them

marcocastignoli:
> The bug can be resolved by adding a COALESCE(creation_match, '') to make the NULL values '' for comparisons.

I didn't know about it, good catch!

services/server/test/integration/repository-handlers/files.spec.ts
```diff
@@ -43,46 +40,84 @@ describe("Verify repository endpoints", function () {
...
+    for (const endpointMatchType of endpointMatchTypes) {
+      it(`should handle pagination in /files/contracts/${endpointMatchType}/${chainFixture.chainId}`, async function () {
+        const contractAddresses: string[] = [];
+
+        // Deploy 5 contracts
+        for (let i = 0; i < 5; i++) {
+          // Deploy partial matching contract if endpoint is partial or choose randomly if endpointMachtype is any. 'any' endpoint results should be consistent regardless.
+          const shouldDeployPartial =
+            endpointMatchType === "partial" ||
+            (endpointMatchType === "any" && Math.random() > 0.5);
```
marcocastignoli, line 54: What's the purpose of using random instead of fixing for example odd as partial and even as full?

marcocastignoli (approved):
Just a doubt on why random is used, but other than that LGTM 👍

services/server/test/integration/repository-handlers/files.spec.ts
kuzdogan: No particular reason, could also use odd etc. The only purpose is to mix up partial and perfect matches.

kuzdogan:
> > Deduplication not working in compiled_contracts
> 
> I cannot understand why this happens. Maybe the unique constraint doesn't work with a NULL column?
> 

Ok I'll spin out another issue for the deduplication issue in `compiled_contracts

> > But it seems these functions themselves are not used at all. Can they be removed?
> 
> Right now they are not used. If we need them for the rebuilding the database we can implement them again, you can remove them

Ok but I don't really understand the use of the `matchType` variable there. I can keep the functions by noting they are unused for now, but the `matchType` is not used inside that function at all.

marcocastignoli:
> Ok but I don't really understand the use of the `matchType` variable there. I can keep the functions by noting they are unused for now, but the `matchType` is not used inside that function at all.
> 
> https://github.com/ethereum/sourcify/blob/28c68274aff6317e7c868c41946cb01d08f51cb1/services/server/src/server/services/utils/database-util.ts#L429-L451

Do you remember that inside sourcify_sync we had some values with `full_match` and some with `perfect`. The reason is that this function wasn't doing what it supposed to be doing, instead of inserting match_type, it should have inserted matchType. It's a mistake. As I told you that function is not used anymore, so we can remove it.

Issue #1419, opened 2024-06-04
Add return types QueryResult<T> to database helper functions


marcocastignoli:
In `services/server/src/server/services/utils/database-util.ts` every function returns a pg query QueryResult<any>, we can improve the return type including the object that the pg Result is returning. E.g. `QueryResult<RowType>`

PR #1422, opened 2024-06-05
Add endpoints and tests for missing files

marcocastignoli:
Missing files:
- constructor-args.txt
- creator-tx-hash
- library-map.json
- immutable-references.json

In this PR I'm not including:
- #1419, I will fix also this PR's missing types in one PR to fix all missing types
- Adding metadata into `sourcify-matches`


kuzdogan (commented):
The code itself LGTM. Just two considerations:

1) The test case name `LibrariesLinkedByCompiler` should actually be `LibrariesLinkedManually`. `library-map.json` is created only when we find placeholders in the contract recompiled bytecode. If the libraries are linked by the compiler (which is the recommended way) they will be in the metadata.json under `libraries` and they will be placed in the recompiled bytecode and won't have placeholders. 

So the test is accurate but just the naming.

2) This PR fixes the requests when these files (`library-map.json` etc.) are requested directly but what happens when all files of a contract are requested via `/files/any` as in the initial report from Blockscout? This does not fix that problem and these files need to be returned inside the response from that endpoint. 

See e.g. the difference of these two:
- https://repo.staging.sourcify.dev/contracts/full_match/11155111/0x4E7095a3519A33dF3D25774c2F9D7a89eB99745D/
- https://staging.sourcify.dev/server/files/any/11155111/0x4E7095a3519A33dF3D25774c2F9D7a89eB99745D

services/server/test/integration/repository-handlers/files.spec.ts
```diff
@@ -30,24 +30,103 @@ describe("Verify repository endpoints", function () {
...
+      .get(
+        `/repository/contracts/full_match/${chainFixture.chainId}/${chainFixture.defaultContractAddress}/metadata.json`
+      );
+    chai
+      .expect(res.text)
+      .to.equal(chainFixture.defaultContractMetadata.toString());
   });
 
-  it("should fetch immutable-references.json of specific address", async function () {
+  it("should fetch immutable-references.json of specific address, and it should be available in /files/tree", async function () {
```
kuzdogan, line 129: I think we should check if the contructor-args.txt etc. are inside the `files/any` with the correct content instead of `files/tree`. That was the main problem

kuzdogan (changes requested):
Small thing about the tests. The implementation code is much cleaner now 👍

PR #1424, opened 2024-06-06
db migration: Make `NULLS NOT DISTINCT` in `compiled_contracts_pseudo_pkey` constraint

### LINKED ISSUE #1420: Investigate deduplication in `compiled_contracts` table
As explained in #1416 the deplucation of the rows seem to be not working in the `compiled_contracts` table. Copying the explanation here:

## Deduplication not working in `compiled_contracts`

When debugging the tests I've noticed the identical test contracts are getting added to the `compiled_contracts` table multiple times. I'd have expected them to be deduplicated because the table has a unique key property in the schema definition:
https://github.com/ethereum/sourcify/blob/653952de199123406cea9c3ebda4eda88cc0652a/services/database/migrations/20231109160022-create-alliance-schema.js#L204

However during testing the contracts get added multiple times:
```SQL
SELECT compiler, "language", creation_code_hash, runtime_code_hash FROM public.compiled_contracts x
```
image

Am I missing something or is there something wrong?

### To reproduce

1) add a breakpoint at
https://github.com/ethereum/sourcify/blob/653952de199123406cea9c3ebda4eda88cc0652a/services/server/test/integration/repository-handlers/files.spec.ts#L56

2) Add an `.only` modifier to describe:
https://github.com/ethereum/sourcify/blob/653952de199123406cea9c3ebda4eda88cc0652a/services/server/test/integration/repository-handlers/files.spec.ts#L43

3) run the `Mocha - Server Integration`, deploy and verify multiple contracts, and connect to the test database at `localhost:5431`.

manuelwedler:
Fixes #1420 

I already ran the commands on the staging v2 database.

Additionally, I added all missing down functions in the migration of the sourcify database.

### DIFF
--- a/services/database/migrations/20231109160023-sourcify.js
+++ b/services/database/migrations/20231109160023-sourcify.js
@@ -33,7 +33,9 @@ exports.up = function (db, callback) {
       db.runSql.bind(
         db,
         `ALTER TABLE compiled_contracts ALTER COLUMN creation_code_hash DROP NOT NULL;
-        ALTER TABLE compiled_contracts ALTER COLUMN creation_code_artifacts DROP NOT NULL;`
+        ALTER TABLE compiled_contracts ALTER COLUMN creation_code_artifacts DROP NOT NULL;
+        ALTER TABLE compiled_contracts DROP CONSTRAINT compiled_contracts_pseudo_pkey;
+        ALTER TABLE compiled_contracts ADD CONSTRAINT compiled_contracts_pseudo_pkey UNIQUE NULLS NOT DISTINCT (compiler, language, creation_code_hash, runtime_code_hash);`
       ),
       db.runSql.bind(
         db,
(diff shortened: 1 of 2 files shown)


kuzdogan:
Are the changes here based on the assumption of running the migrations for the first time (i.e. setting up the DB) or on modifying the current DB? Because in my understanding migrations should be run when you're setting up the DB for the first time and if we needed to make changes on the DB we should make it manually.

manuelwedler:
It's under the assumption of setting it up for the first time. The first line dropping the constraint is needed here, since we first apply the constraint from the VerA db migration. Only after dropping we can apply the correct constraint for sourcify db. 

Since the staging v2 db is already running, I applied the changes there manually.

The tests are failing btw since the file endpoint tests assumed the old constraint. Working on fixing them at the moment.

kuzdogan:
Got it, thanks, that was exactly my question. 

Then it looks good to me

kuzdogan (approved):

manuelwedler:
The failing the was actually due to a problem in our database queries. I found some problems where we don't properly compare `NULL` values.

fantastilicious88 (changes requested):
Not right

fantastilicious88 (commented):
Cool

fantastilicious88:
Cool

PR #1426, opened 2024-06-11
Replace Ganache with Hardhat Network in tests

### LINKED ISSUE #1382: Replace Ganache in server tests because of deprecation
Ganache has been deprecated for a while now. We should move to another test chain, like Hardhat Network or Foundry's Anvil.

manuelwedler:
Closes #1382 

Hardhat Network is created as a child process as they don't provide a way to programmatically create a network (as we did with Ganache).


kuzdogan (approved):
LGTM 👍

PR #1446, opened 2024-06-21
Use Node.js v22

### LINKED ISSUE #1428: Upgrade Node version to v22

manuelwedler:
Closes #1428


kuzdogan (commented):
LGTM in general. Two things:

1) We should check the  code formatting settings in `lib-sourcify`. I see. your formatter is adding a traliing comma to function args. @marcocastignoli Is your code formatter adding a trailing comma as well to say `verification.ts` on save?

2) The node version settings of the lambda compiler I think we need to set on the AWS console.  

Also the reminder to wait for #1441 before merging this

manuelwedler:
1. That's because I upgraded prettier. It seems they changed the rule from v2 to v3. If you like to have no trailing comma, I can add the related rule. I personally prefer trailing commas, because it makes it a bit easier to move lines around.
2. I already checked AWS. We already use Node v20 there, and that's also the highest version supported on Lambda at the moment.

kuzdogan:
Ok I've moved prettier to root level and it seems to be resolved with that. My VSCode Prettier extension was resolving to a v2 Prettier. It makes more sense to have a single version in the root instead of individual versions in each module.

One thing I noticed is we still use node v18 in the monitor (check `monitor/.nvmrc`) Is that intentional?

kuzdogan (approved):
lgtm

PR #1449, opened 2024-06-24
Store metadata in database

marcocastignoli:
- This PR adds a new column `sourcify_matches.metadata`. The contract's metadata is going to be stored in this new field. This new field is going to be used by SourcifyDatabaseService as metadata's source of truth instead of RepositoryV2
- This PR also add a new check in tests: keccak256(repository metadata) === keccak256(database metadata)
- Metadata is always stored in database even for cases in which contracts don't have a metadata field in the bytecode (just like we do on repository)

Doubts:
- ~~We are going to rebuild production's database from scratch so we are not going to have problems while applying `ALTER TABLE sourcify_matches ADD COLUMN metadata json NOT NULL`. How do we want to handle this case on staging?~~ We are going to rebuild also staging database from scratch


manuelwedler (approved):
Looks good to me :+1:  nothing to add here

PR #1451, opened 2024-06-25
Add queries types

marcocastignoli:
See #1419


services/server/src/server/services/utils/database-util.ts
```diff
@@ -140,18 +128,23 @@ export async function getVerifiedContractByChainAndAddress(
   );
 }
 
+export type GetSourcifyMatchByChainAddressResult = Tables.SourcifyMatch &
+  Pick<Tables.VerifiedContract, "creation_values" | "runtime_values"> &
```
manuelwedler, line 132: Nice how you used Pick to create a type for a joined query!

services/server/src/server/services/utils/database-util.ts
```diff
@@ -607,7 +575,7 @@ export async function getSourcifyMatchAddressesByChainAndMatch(
   page: number,
   paginationSize: number,
   descending: boolean = false,
-) {
+): Promise<QueryResult<{ address: string }>> {
   let queryWhere = "";
   switch (match) {
     case "full_match": {
```
manuelwedler, line 581: (not related to this line, just couldn't pick the right one)

 The `updateContractDeployment` function would also need a return type.

manuelwedler (commented):
Thanks for adding these! just a few small improvements that could be made

services/server/src/server/services/utils/database-util.ts
marcocastignoli: it actually doesn't need a specific return type, I removed the `RETURN *`

manuelwedler: Okay great :+1:

manuelwedler (approved):

Issue #1453, opened 2024-06-27
Update Sourcify database schema to match Verifier Alliance changes


marcocastignoli:
See more information here: https://github.com/verifier-alliance/database-specs/pull/4

PR #1454, opened 2024-06-27
Change code_hash type to sha256  

marcocastignoli:
See #1453 

This PR:
- Create new types for stricter type checking for Bytes fields in the database
- Change code_hash type to sha256  
- Adds a new field code_hash_keccak
- Validate keccak256 and sha256 fields in Database tests

## Summary by CodeRabbit

- **New Features**
  - Introduced a new synchronization process for contracts in Sourcify.
  - Added new types for enhanced type safety (Bytes, BytesSha, BytesKeccak).

- **Bug Fixes**
  - Improved handling of undefined values in database operations.
  - Ensured integrity of code hashes using sha3-256.

- **Documentation**
  - Clarified database migration instructions for production environment setup.

- **Refactor**
  - Updated bytecode hash types for better clarity and consistency.

- **Chores**
  - Added production configuration settings for PostgreSQL.

### DIFF
--- a/services/server/package.json
+++ b/services/server/package.json
@@ -67,6 +67,7 @@
     "express-rate-limit": "^6.7.0",
     "express-session": "^1.18.0",
     "http-status-codes": "^2.3.0",
+    "js-sha3": "^0.9.3",
     "json-refs": "^3.0.15",
     "memorystore": "^1.6.7",
     "pg": "^8.12.0",
--- a/services/server/package.json
+++ b/services/server/package.json
@@ -67,7 +67,6 @@
     "express-rate-limit": "^6.7.0",
     "express-session": "^1.18.0",
     "http-status-codes": "^2.3.0",
-    "js-sha3": "^0.9.3",
     "json-refs": "^3.0.15",
     "memorystore": "^1.6.7",
     "pg": "^8.12.0",
--- a/services/server/src/server/services/storageServices/AbstractDatabaseService.ts
+++ b/services/server/src/server/services/storageServices/AbstractDatabaseService.ts
@@ -9,11 +9,10 @@ import {
   bytesFromString,
   normalizeRecompiledBytecodes,
 } from "../utils/database-util";
-import { Pool } from "pg";
+import { Pool, QueryResult } from "pg";
 import { AuthTypes, Connector } from "@google-cloud/cloud-sql-connector";
 import logger from "../../../common/logger";
-import { sha3_256 } from "js-sha3";
-import { BytesKeccak, BytesSha } from "../../types";
+import { BytesKeccak } from "../../types";
 
 export interface DatabaseServiceOptions {
   googleCloudSql?: {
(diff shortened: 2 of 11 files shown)


services/server/src/server/services/utils/database-util.ts
```diff
@@ -171,11 +179,11 @@ ${
 
 export async function insertCode(
   pool: Pool,
-  { bytecode_hash, bytecode }: Tables.Code,
+  { bytecode_hash, bytecode_hash_keccak, bytecode }: Tables.Code,
 ) {
   await pool.query(
-    "INSERT INTO code (code_hash, code) VALUES ($1, $2) ON CONFLICT (code_hash) DO NOTHING",
-    [bytecode_hash, bytecode],
+    "INSERT INTO code (code_hash, code_hash_keccak, code) VALUES ($1, $2, $3) ON CONFLICT (code_hash) DO NOTHING",
```
kuzdogan, line 185: I don't think you can instert `code_hash`? That was the whole point of the changes, no? 

Also IMO that means, we should get the `code_hash` from the DB and not calculate ourselves on our server.

services/server/package.json
```diff
@@ -67,6 +67,7 @@
     "express-rate-limit": "^6.7.0",
     "express-session": "^1.18.0",
     "http-status-codes": "^2.3.0",
+    "js-sha3": "^0.9.3",
```
kuzdogan, line 70: Do we need this library? Doesn't Node v20+ have sha3-256 in its `crypto` lib?

services/server/test/integration/database.spec.ts
```diff
@@ -197,19 +203,31 @@ describe("Verifier Alliance database", function () {
     chai.expect(parseInt(row.block_number)).to.equal(blockNumber);
     chai.expect(parseInt(row.transaction_index)).to.equal(txIndex);
 
-    ///
+    // Check Keccak256 for code.code_hash_keccak
+    chai
+      .expect(`0x${toHexString(row.compiled_creation_code_hash_keccak)}`)
+      .to.equal(keccak256(bytesFromString(testCase.compiled_creation_code)));
```
kuzdogan, line 209: Actually... I think we should not be calculating values when asserting test cases. Imagine for example there's a bug in the keccak function. Or it could be that the output value is different when you input a `Buffer` vs `string` to the `keccak` function. 

Since we are also calculating the `code_hash_keccak` field in the same way as this line in our server, this assertion is actually self-affirming. I think the test cases should contain the actual hash values and not be calculated on the fly. 

This doesn't have to be done in this PR but we can branch out this to another issue.

Is it clear what I mean?

services/server/src/server/services/storageServices/AbstractDatabaseService.ts
```diff
@@ -155,6 +189,12 @@ export default abstract class AbstractDatabaseService {
     recompiledContract: CheckedContract,
     match: Match,
   ): Promise<Database.DatabaseColumns> {
+    const {
+      sha256OnchainCreationBytecode,
+      sha256OnchainRuntimeBytecode,
+      sha256RecompiledCreationBytecode,
+      sha256RecompiledRuntimeBytecode,
+    } = this.getSha256Bytecodes(recompiledContract, match);
     const {
```
kuzdogan, line 198: Shouldn't we be getting the sha256 value directly from the DB response instead of calculating ourselves? See the comment in `database-utils.ts`

kuzdogan (changes requested):

kuzdogan:
See the points. I haven't reviewed the full code as it depends on my comment. I'll do another review after our discussion.

(Couldn't write it in the review message)

services/server/test/integration/database.spec.ts
marcocastignoli: It is clear, from my pov we can postpone this to another PR and more analysis

marcocastignoli:
@kuzdogan Ok got your point, so now we don't calculate sha256 from javascript but we use directly postgres digest function

services/server/src/server/services/utils/database-util.ts
```diff
@@ -93,25 +93,42 @@ export namespace Tables {
...
-  compiledContract: Partial<Tables.CompiledContract>;
-  verifiedContract: Partial<Tables.VerifiedContract>;
+  compiledContract: Pick<
+    Tables.CompiledContract,
+    | "language"
+    | "fully_qualified_name"
+    | "compilation_artifacts"
+    | "creation_code_artifacts"
+    | "runtime_code_artifacts"
+  >;
```
kuzdogan, line 110: Why are you explicitly picking fields here and not the full Tables.CompiledContract type?

kuzdogan (approved):
LGTM. One small question. Also let's move the unresolved "test value assertion" discussion in a separate issue.

services/server/src/server/services/utils/database-util.ts
marcocastignoli: Because we need all fields but not id

marcocastignoli: Because we need all fields but not id

marcocastignoli: the alternative is 

```
  verifiedContract: Omit<
    Tables.VerifiedContract,
    "id" | "compilation_id" | "deployment_id"
  >;
```

kuzdogan: Wait do you mean

```ts
  compiledContract: Omit<  
    Tables.CompiledContract,  
    "id"
  >;  
```

I also think this is much more concise and understandable

kuzdogan (approved):

PR #1472, opened 2024-07-10
Add `metadata_match` fields and remove cborAuxdataPositions from artifacts

marcocastignoli:
- add `metadata_match`
- remove cborAuxdataPositions from artifacts
- update verifier-alliance test files accordingly
- update tests


services/database/migrations/20240710103541-add-metadata_match.js
```diff
@@ -0,0 +1,53 @@
...
+exports.up = function (db, callback) {
+  async.series(
+    [
+      db.runSql.bind(
+        db,
+        `ALTER TABLE public.verified_contracts ADD runtime_metadata_match bool NOT NULL;`,
+      ),
+      db.runSql.bind(
+        db,
+        `ALTER TABLE public.verified_contracts ADD creation_metadata_match bool NOT NULL;`,
```
manuelwedler, line 28: Shouldn't it be `ADD COLUMN`?

services/database/migrations/20240710103541-add-metadata_match.js
```diff
@@ -0,0 +1,53 @@
...
+exports.down = function (db, callback) {
+  async.series(
+    [
+      db.runSql.bind(
+        db,
+        `ALTER TABLE sourcify_matches DROP COLUMN runtime_metadata_match;`,
+      ),
+      db.runSql.bind(
+        db,
+        `ALTER TABLE sourcify_matches DROP COLUMN creation_metadata_match;`,
```
manuelwedler, line 44: These queries are altering the wrong table. Should be `public.verified_contracts`.

manuelwedler (changes requested):
The db migrations need some changes. The rest looks good :+1:

services/database/migrations/20240710103541-add-metadata_match.js
marcocastignoli: apparently it works in both ways :)

manuelwedler: ah I didn't know! Being a bit more verbose here makes the query at least more understandable :)

manuelwedler (approved):

PR #1475, opened 2024-07-15
Fix storage of metadata files

manuelwedler:
We found that the metadata file which the server stores is the one provided by the user. This can be a problem when a user uploads a wrong metadata file which still yields a verification match. In this case, we were storing the wrong metadata file. This PR fixes that by assigning the metadata of the recompilation to the CheckedContract. I added tests on lib-sourcify and on the server.


packages/lib-sourcify/src/lib/verification.ts
```diff
@@ -74,6 +74,12 @@ export async function verifyDeployed(
     );
   }
 
+  // Store the correct metadata in case the wrong metadata yields a match
+  checkedContract.initSolcJsonInput(
+    JSON.parse(recompiled.metadata),
+    checkedContract.solidity,
+  );
```
kuzdogan, line 81: Shouldn't this be done in `recompile()`? And not as `initSolcJsonInput` but just assigning the metadata and metadataRaw

services/server/test/integration/path-sanitization.spec.ts
kuzdogan, line 1: Are these formatting changes?

services/server/test/testcontracts/Storage/metadataModified.json
kuzdogan, line 1: Why are we touching this file?

services/server/test/verifier-alliance/constructor_arguments.json
kuzdogan, line 1: Why are we removing the `v` prefixes in all these verifier-alliance test files? Is this another change in this PR? I didn't see anything related in the description

kuzdogan (changes requested):
Asked for a change in the implementation and I see some changed that I couldn't understand why they are there

services/server/test/integration/path-sanitization.spec.ts
manuelwedler: I realized that when using the async import, the received object contains a `default` key under which the object is stored again. This is something about Node.js's module system.

I had failing tests because the imported metadata.json was not the exact same object as in the file, but also included this `default` key. For consistency, I just refactored all asynchronous JSON imports, such that we don't run into unexpected behavior in the future.

services/server/test/verifier-alliance/constructor_arguments.json
manuelwedler: The compiler outputs the Solidity version without the `v` prefix. As the database tests check the metadata from these json files against the stored metadata, I had failing tests here. So I needed to make sure the recompiled metadata matches the one defined here. 

I could have been a bit more descriptive about the test changes in the PR description. Sorry for that

services/server/test/testcontracts/Storage/metadataModified.json
manuelwedler: Because the recompiled metadata does not include the content. I needed to make the recompiled match with the metadata defined here, as in my other comment.

packages/lib-sourcify/src/lib/verification.ts
manuelwedler: After thinking about it again, I agree that it makes more sense to have this in `recompile()`.

I mainly used `initSolcJsonInput`, because we also use it in line 311 for replacing the stored metadata. I just realized that we use it there because we have another `CheckedContract` instance. So yes makes sense to just assign the metadata

kuzdogan (approved):
lgtm

PR #1476, opened 2024-07-16
Implement vera changes

marcocastignoli:
- https://github.com/verifier-alliance/database-specs/pull/5
- https://github.com/verifier-alliance/database-specs/pull/6


services/server/src/server/services/storageServices/AbstractDatabaseService.ts
```diff
@@ -442,7 +459,7 @@ export default abstract class AbstractDatabaseService {
       }
 
       if (
-        (hasCreationAuxdataTransformation &&
+        (!existingVerifiedContract.creation_metadata_match &&
```
marcocastignoli, line 462: This is not related to this PR, but I noticed that we were still using transformations to check if an existing match needs to be upgraded

kuzdogan (approved):
Just one comment

PR #1479, opened 2024-07-16
Nullable metadata_match

marcocastignoli:
- https://github.com/verifier-alliance/database-specs/pull/10


kuzdogan (approved):
All good. 

Just as a side note I'm getting worried about the number of migration files we have and wondering if it will make it more difficult to manage. Could be that it actually makes it easier. Just saying

PR #1489, opened 2024-07-24
Turn off DBs

kuzdogan:
Turning off writing to DBs to be able to do a release because the DB schema on prod. is not updated and the release would fail to write contracts. The repo writing logic must be unaffected by the latest changes. We are doing an earlier release to add new chains. The contracts after this change will get written to repoV1/2 and we will add them later during the migration.

And using repoV1 as the read source since writing to the DB is turned off.

Similarly the VerA DB schemas are still not ready. It is ok to have writing to VerA turned off on staging but it would be nice if we could get the VerA schema ready before the prod. migration so that we write everything simulatanously to Sourcify and VerA and don't have to move rows from Sourcify DB to VerA

- [ ] After merge, double check if indeed the staging instance stops writing to the DB

### DIFF
--- a/services/server/src/config/master.js
+++ b/services/server/src/config/master.js
@@ -1,6 +1,22 @@
+const {
+  RWStorageIdentifiers,
+  WStorageIdentifiers,
+} = require("../server/services/storageServices/identifiers");
+
 module.exports = {
   server: {
     port: 80,
+  },
+  storage: {
+    read: RWStorageIdentifiers.RepositoryV1,
+    writeOrWarn: [
+      // WStorageIdentifiers.AllianceDatabase,
+      RWStorageIdentifiers.RepositoryV1,
+    ],
+    writeOrErr: [
+      WStorageIdentifiers.RepositoryV2,
+      // RWStorageIdentifiers.SourcifyDatabase,
+    ],
   },
   repositoryV1: {
     path: "/home/app/repository",
--- a/services/server/src/config/staging.js
+++ b/services/server/src/config/staging.js
@@ -8,7 +8,7 @@ module.exports = {
     port: 80,
   },
   storage: {
-    read: RWStorageIdentifiers.SourcifyDatabase,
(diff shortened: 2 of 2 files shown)


services/server/src/config/master.js
```diff
@@ -1,7 +1,23 @@
+const {
+  RWStorageIdentifiers,
```
Chator1, line 2: 0xFD689e5f2d8d9Aec0aD328225Ae62FdBDdb30328

services/server/src/config/staging.js
```diff
@@ -1,7 +1,23 @@
+const {
+  RWStorageIdentifiers,
+  WStorageIdentifiers,
+} = require("../server/services/storageServices/identifiers");
+
 module.exports = {
   server: {
     port: 80,
   },
+  storage: {
```
Chator1, line 10: 0xFD689e5f2d8d9Aec0aD328225Ae62FdBDdb30328

Chator1 (approved):

manuelwedler (approved):
Looks good. However, I don't get why we don't use RepoV2 for reading. Could you clarify that?

Chator1:
Hello sir, I hope you are in the best condition. I want you to know that my
knowledge of this work is very limited or non-existent. I do not know
anything, but I am trying hard. Perhaps I can help my family with this. We
are going through very difficult circumstances. I hope you can help and
guide me. Many thanks and appreciation to you.

في الأربعاء، ٢٤ يوليو ٢٠٢٤ ٧:٣٥ م Manuel Wedler ***@***.***>
كتب:

kuzdogan:
> Looks good. However, I don't get why we don't use RepoV2 for reading. Could you clarify that?

Now that you ask I'm thinking we could, maybe @marcocastignoli remembers and can enlighten us when he's back. But I guess it was about us saving file names in keccak hashes?? I mean we can still work around that. Maybe we were lazy and thought repoV1 is ok for now.

PR #1493, opened 2024-07-29
add `compiled_contracts` artifacts constraints

marcocastignoli:
See https://github.com/verifier-alliance/database-specs/pull/12

### DIFF
--- a/services/database/migrations/20240729112258-compiled_contracts-constraints.js
+++ b/services/database/migrations/20240729112258-compiled_contracts-constraints.js
@@ -23,12 +23,20 @@ exports.up = function (db, callback) {
       db.runSql.bind(
         db,
         `
+        CREATE OR REPLACE FUNCTION is_object(obj jsonb)
+            RETURNS boolean AS
+        $$
+        BEGIN
+            RETURN
+                jsonb_typeof(obj) = 'object';
+        END;
+        $$ LANGUAGE plpgsql;
+
         CREATE OR REPLACE FUNCTION validate_json_object_keys(obj jsonb, mandatory_keys text[], optional_keys text[])
             RETURNS boolean AS
         $$
         BEGIN
             RETURN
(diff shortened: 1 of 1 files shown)


services/database/migrations/20240729112258-compiled_contracts-constraints.js
```diff
@@ -0,0 +1,109 @@
+"use strict";
```
kuzdogan, line 1:
```suggestion
// Constraints for the `compiled_contracts` table, according to the Verifier Alliance schema https://github.com/verifier-alliance/database-specs/pull/12
"use strict";
```

services/database/migrations/20240729112258-compiled_contracts-constraints.js
```diff
@@ -0,0 +1,109 @@
...
+exports.down = function (db, callback) {
+  async.series(
+    [
+      db.runSql.bind(
+        db,
+        `
+        ALTER TABLE compiled_contracts DROP CONSTRAINT compiled_contracts_pseudo_pkey;
+        
+        ALTER TABLE compiled_contracts
+        ADD CONSTRAINT compiled_contracts_pseudo_pkey UNIQUE (compiler, language, creation_code_hash, runtime_code_hash);
```
kuzdogan, line 93: Why do we add and remove the constraint here?

kuzdogan (commented):
I haven't looked in detail but I guess it's a copy paste of the VerA code?

In that case let's wait for that PR before merging this

services/database/migrations/20240729112258-compiled_contracts-constraints.js
marcocastignoli: down migration is used to restore the state to the previous one. So I'm removing the new constraint and adding back the old one.

kuzdogan (approved):

PR #1535, opened 2024-08-06
Implement gateways custom settings in Monitor

marcocastignoli:
Also add support for headers setting

See https://github.com/sourcifyeth/private-issues/issues/27


manuelwedler (approved):
looks good :)

PR #1549, opened 2024-08-12
Changes to the database sync script

kuzdogan:
Introduces changes to the DB sync script:
- Adds a `--deprecated` option to mark some chains deprecated and to send them to the `/verify-deprecated` endpoint instead.
- For more max concurrency, Removes the 1s wait between calls and reduces the `await` time inside the while loop
- Refactors `processContract` to remove the ineffective Promise. Prev. it was an `async` function returning a Promise, so it was a double Promise. Async functions should not return Promises.
- Conditionally enable `/verify-deprecated` in config
- Don't only send files with `.sol` and the `metadata.json` but all files. This is needed as some contract files don't have a file extension in our legacy repository.


marcocastignoli (approved):
lgtm

Issue #1548, opened 2024-08-12
Support Filebase IPFS gateway under authentication


marcocastignoli:
As we did for monitor (https://github.com/ethereum/sourcify/commit/d99067a94f0486f5e98d253ad9609398479cf3d4) we need to support custom headers in lib-sourcify's `CheckedContract.fetchMissing` for IPFS gateways so that we can send Filebase authentication.

marcocastignoli:
Closed by #1550

PR #1550, opened 2024-08-13
Add support for headers to IPFS Gateways

marcocastignoli:
See #1548 

On a side note: I didn’t implement a complex mechanism for handling IPFS gateways, but I think it will be needed in a future PR. We might be able to reuse code from the monitor to support multiple IPFS gateways.


packages/lib-sourcify/test/functions.spec.ts
```diff
@@ -137,18 +141,53 @@ describe('Checked contract', () => {
   it('Should return null after failed performFetch', async () => {
     expect(await performFetch('httpx://')).to.equal(null);
   });
+  it('Should call fetchWithBackoff with headers', async () => {
```
kuzdogan, line 144: Here I think we can check the actual headers sent with the request if we want to check the headers: https://chatgpt.com/share/e/1785b6e6-f18d-4f72-9820-62fc5b186cac

packages/lib-sourcify/src/lib/CheckedContract.ts
```diff
@@ -748,8 +756,20 @@ function createJsonInputFromMetadata(
...
+  if (process.env.IPFS_GATEWAY_HEADERS) {
+    try {
+      ipfsGatewaysHeaders = JSON.parse(process.env.IPFS_GATEWAY_HEADERS);
+    } catch (error) {
+      logError('Error while parsing IPFS_GATEWAY_HEADERS option', { error });
+      throw new Error('Error while parsing IPFS_GATEWAY_HEADERS option');
+    }
+  }
+  return {
+    url: process.env.IPFS_GATEWAY || 'https://ipfs.io/ipfs/',
```
kuzdogan, line 770: Let's make sure this url has a trailing slash if it doesn't. Later on we append IPFS CID's assuming there's a trailing slash.

kuzdogan (changes requested):
Small changes.

I see we add a change in session verification endpoints but shouldn't there be also one in stateless then?

We also need to add this new env var to the server .env.template.

packages/lib-sourcify/test/functions.spec.ts
kuzdogan: Chatgpt response 

In your test where you want to check the actual headers sent in a request, you can leverage the capabilities of `nock` to intercept and inspect the requests directly. Here's how you can enhance your test to validate that the headers sent in the request match the expected headers:

### Modifying the Test to Check Headers

First, ensure you have a way to capture the headers within the `nock` interceptor. You can do this by using a callback function in the `.reply()` method of `nock` to access the incoming request details. Here’s an example of how you can adjust your existing test to inspect and assert the headers:

```javascript
const nock = require('nock');
const expect = require('chai').expect;

describe('Fetch Tests', () => {
  it('Should call fetchWithBackoff with headers', async () => {
    const ipfsGateway = 'http://ipfs-gateway';

    // Create a variable to store the headers received in the request
    let actualHeaders;

    nock(ipfsGateway, {
      reqheaders: {
        test: 'test',
      },
    })
      .get('/')
      .reply(function (uri, requestBody) {
        // Store headers for inspection
        actualHeaders = this.req.headers;
        return [200];
      });

    const response = await fetchWithBackoff(ipfsGateway, {
      headers: {
        test: 'test',
      },
    });

    expect(response.status).to.equal(200);
    // Now assert the actual headers
    expect(actualHeaders.test).to.equal('test');
  });

  it('Should fail getIpfsGateway with unparsable headers', async () => {
    process.env.IPFS_GATEWAY_HEADERS = `{ "test": unparsable-value }`;
    try {
      getIpfsGateway();
    } catch (error: any) {
      expect(error.message).to.equal(
        'Error while parsing IPFS_GATEWAY_HEADERS option',
      );
    }
  });
});
```

### Key Changes:

1. **Storing Headers for Inspection:** I added a variable `actualHeaders` to store the headers from the intercepted request.
2. **Reply Callback Function:** Inside the `.reply()` method, I assigned the request headers to `actualHeaders`. This is accessed via `this.req.headers` within the callback.
3. **Header Assertion:** After the request is made, you can assert the values of `actualHeaders` to ensure they contain the expected header values.

Make sure that the header names match the actual header names sent by your `fetchWithBackoff` function and how headers are formatted in HTTP requests (they might be lowercased, for instance).

This approach gives you a direct method to inspect and assert the content of HTTP headers in requests sent by your application code during tests.

marcocastignoli: The request is accepted only if headers contains exactly `test: 'test',`, so it doesn't make sense to check twice

marcocastignoli:
@kuzdogan 
> I see we add a change in session verification endpoints but shouldn't there be also one in stateless then?

The change is at `/session/input-contract`, that doesn't have a equivalent stateless endpoint.

packages/lib-sourcify/test/functions.spec.ts
kuzdogan: Ah ok so 
```
    nock(ipfsGateway, {
      reqheaders: {
        test: 'test',
      },
    })
```

 would fail for example if you have sent 
```
await fetchWithBackoff(ipfsGateway, {
      test: 'testdasdasdsadadasdd',
    });
```

marcocastignoli:
> We also need to add this new env var to the server .env.template.

Added to `.env.dev`

packages/lib-sourcify/test/functions.spec.ts
marcocastignoli: Exactly

kuzdogan (approved):

PR #1563, opened 2024-08-15
Update sync script

kuzdogan:
- Adds a command `sync-single` to debug single contracts
- Throws inside the `processContract` instead of retunring `false`
- Also throws if status is not `200`. Previously the script assumed these to be verified (logs: Success) but does not update the sync table (correctly). This gave the impression those contracts were verified from script's output.


services/server/src/config/master.js
```diff
@@ -37,7 +37,7 @@ module.exports = {
     // credentials as env vars
   },
   rateLimit: {
-    enabled: true,
+    enabled: false,
```
marcocastignoli, line 40: Why are these changes here?

marcocastignoli: They are already on staging

marcocastignoli (approved):

PR #1564, opened 2024-08-15
Log according to GCP severity format if running on GCP

### LINKED ISSUE #1559: Adapt log levels for GCP
Sourcify by default outputs the logs of different levels in a field called `level` in the output JSON. This is the standard output of `winston` and as well as what Prometheus expects.

However GCP uses a different field to mark this info, namely `severity` which is an uppercase enum e.g. `WARNING` ([see docs](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity))

We should have a check in the server logger if server is running on GCP and modify the log field accordingly.

kuzdogan:
Fixes #1559

### DIFF
--- a/services/server/src/common/logger.ts
+++ b/services/server/src/common/logger.ts
@@ -83,19 +83,50 @@ const injectRequestId = format((info) => {
   return requestId ? { ...info, requestId } : info;
 });
 
+// Choose between the GCP and the standard JSON format.
+const chooseJSONFormat = () => {
+  const isOnGCP = process.env.K_SERVICE || process.env.GOOGLE_CLOUD_PROJECT;
+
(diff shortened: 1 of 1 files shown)


marcocastignoli (approved):
I just have a doubt about `silly: "DEBUG"`, but we can keep it for now and change it in the future if we realize that it's actually a problem

kuzdogan:
GCP has no "SILLY" unfortunately. But we barely use silly on deployments. This does not mean `logger.silly` entries will be printed when `NODE_LOG_LEVEL=debug`. Just `logger.silly` entries will appear "DEBUG" level when `NODE_LOG_LEVEL=silly`.

PR #1572, opened 2024-08-16
Store found creatorTxHash even if getContractCreationBytecodeAndReceipt fails

manuelwedler:
We found a problem where the server didn't store the creation tx hash for a contract even though it found it before. The problem was there that lib-sourcify does not add it to the match if it was not able to fetch the creation bytecode, for example because of an unresponsive RPC. This PR changes that behavior. The creatorTxHash is always added to the match if the server obtained it itself. If the user provided the creatorTxHash, we don't add it in this case, because it could be the wrong tx hash.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37421,7 +37421,7 @@
         "gh-pages": "6.1.1",
         "hardhat": "2.22.8",
         "mocha": "10.7.3",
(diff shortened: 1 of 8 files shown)


kuzdogan:
We decided not to store the creator tx hashes directly but instead re-verify those contracts. See #1589

PR #1582, opened 2024-08-20
Fix filename collusion

kuzdogan:
Found in the [failin CI chain test (6119)](https://app.circleci.com/pipelines/github/ethereum/sourcify/6708/workflows/691eabb9-fba5-4a94-a762-a9344bc32471/jobs/40656), where it was unexpectedly fetching files from IPFS when the files are already provided.

Turns out when there are files with the same name, the files get overwritten by the `readFilesRecursively` function. Fixed by using a relative path as the key in the `files` object.

Also adds a log line to debug easier missing files when checking contracts

### DIFF
--- a/packages/lib-sourcify/src/lib/validation.ts
+++ b/packages/lib-sourcify/src/lib/validation.ts
@@ -102,6 +102,11 @@ export async function checkFiles(
   metadataFiles.forEach((metadata) => {
     const { foundSources, missingSources, invalidSources, metadata2provided } =
       rearrangeSources(metadata, byHash);
+    logDebug(`Checking contract`, {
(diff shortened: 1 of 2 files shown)


marcocastignoli (commented):
Is it expected that once `baseDirectory` is set in the first call (`baseDirectory = baseDirectory || directoryPath`) then it never changes?

It's not clear why isn't enough to use `filePath` instead of `file` in `files[file] = fs.readFileSync(filePath).toString();`

kuzdogan:
Yes baseDirectory is only passed in recursive calls. 

The reason I used a relativePath was to make it nicer, as `filePath` is the full absolute path. That could be more failproof actually.

marcocastignoli (approved):

PR #1585, opened 2024-08-20
Add chain 1625

oyyblin:
This PR adds support for Gravity Alpha Mainnet - https://chainlist.org/chain/1625


kuzdogan:
Thanks for the PR, two things:
1. Please don't change the package-lock
2. Don't add an `rpc` field in the chains file. We will use the one from the chains.json which is the same as you've provided. If the chain is recently added let us know and we'll update the chains.json file.

kuzdogan:
Closing due to inactivity

PR #1587, opened 2024-08-21
Throw HTTP 409 when a partial match already exists on partial verification

kuzdogan:


marcocastignoli (changes requested):
I see `409` response is not included into `paths.yaml` files, the rest looks good 👍

marcocastignoli (approved):

PR #1591, opened 2024-08-26
Remove support for 9977, 192837465, and 842. Update chains.json

kuzdogan:


marcocastignoli:
Should we also disable tests for those chains?

kuzdogan:
> Should we also disable tests for those chains?

Yes thanks for the reminder

marcocastignoli (approved):

Issue #1562, opened 2024-08-15
Going through the unverified contracts during migration


kuzdogan:
When moving from the old infra to GCP, we had to re-verify all contracts on the GCP instance. This is because the old infra only had files (after the [incident](https://twitter.com/SourcifyEth/status/1799159285676687636), we had the backups) and to have the DB data, we need to compile them all and verify. 

In the end we verified +5M contracts, this is the final sync table, ordered by `not_synced`, totals on the last row:

<details>
<summary>Table</summary>

|chain_id|synced|not_synced|total_sync_status|
|--------|------|----------|-----------------|
|42|0|45231|45231|
|1|684901|35830|720731|
|5|151834|942|152776|
|69|10963|684|11647|
|42161|505612|240|505852|
|11155111|279989|124|280113|
|8453|688097|115|688212|
|5700|69|106|175|
|84531|49732|98|49830|
|56|10665|96|10761|
|4|124145|95|124240|
|80001|207251|71|207322|
|137|89574|56|89630|
|28|364|50|414|
|3|76503|47|76550|
|10243|16|47|63|
|10|1923456|36|1923492|
|1149|0|32|32|
|42220|3727|24|3751|
|42261|215|21|236|
|43113|16160|21|16181|
|100|8244|19|8263|
|82|818|19|837|
|122|467|17|484|
|44787|6304|17|6321|
|83|930|15|945|
|23294|142|15|157|
|71402|89|15|104|
|1101|139|14|153|
|71401|262|13|275|
|77|1140|12|1152|
|0|0|11|11|
|11155420|12004|11|12015|
|9001|283|10|293|
|7700|862|9|871|
|8217|19|9|28|
|202401|0|9|9|
|10200|257|8|265|
|534352|117|8|125|
|42262|77|7|84|
|43114|15937|7|15944|
|2222|1582|7|1589|
|7672|137|6|143|
|13337|321|6|327|
|106|115|6|121|
|369|130902|5|130907|
|97|12012|5|12017|
|1115|712|5|717|
|41|987|5|992|
|420|5235|4|5239|
|888|66|4|70|
|288|416|4|420|
|17000|11339|4|11343|
|5000|139|3|142|
|51|41|3|44|
|40|1444|3|1447|
|1313161555|166|2|168|
|9996|126|2|128|
|23295|189|2|191|
|314|74|2|76|
|62621|33|2|35|
|666666666|58|2|60|
|1116|3698|2|3700|
|7701|172|2|174|
[...]

PR #1594, opened 2024-08-27
Fix verification logic

kuzdogan:
This PR adds two changes:

1. https://github.com/ethereum/sourcify/commit/2bf7168edc30416d38b5576552902141c9015373 fixes the incorrect return when a different auxdata was found in the creation bytecode. This became apparent when debugging the "cannot generate artifacts" error in #1562. Here the contract has a different auxdata at the end of the creation code because it returns raw bytecode of the factory contract. The verification logic incorrectly stops when it finds different auxdata, whereas it should've continued to find the auxdatas by editing the contract and recompiling.
2. Fixes the wrong match check (`match.creationBytecode` instead of `match.runtimeBytecode`) in the verification logic.

Also adds small modifications to logs during fetching traces for clarity.

### DIFF
--- a/packages/lib-sourcify/src/lib/verification.ts
+++ b/packages/lib-sourcify/src/lib/verification.ts
@@ -188,7 +188,7 @@ export async function verifyDeployed(
         });
         match = await tryToFindPerfectMetadataAndMatch(
           checkedContract,
-          runtimeBytecode,
+          runtimeBytecode, // TODO: This is also weird we pass the runtime bytecode here
           match,
           async (match, recompiled) => {
             await matchWithCreationTx(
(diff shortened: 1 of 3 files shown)


packages/lib-sourcify/src/lib/verification.ts
```diff
@@ -179,11 +179,11 @@ export async function verifyDeployed(
...
-      if (match.runtimeMatch === 'partial') {
+      if (match.creationMatch === 'partial') {
         logDebug('Matched partial with creation tx', {
           chain: sourcifyChain.chainId,
           address,
-          runtimeMatch: match.runtimeMatch,
+          creationMatch: match.creationMatch,
           creatorTxHash,
         });
         match = await tryToFindPerfectMetadataAndMatch(
```
marcocastignoli, line 189: Just FYI, I thought that also the `runtimeBytecode` passed to `tryToFindPerfectMetadataAndMatch` was wrong but `tryToFindPerfectMetadataAndMatch` use it only to extract the decoded auxdata.

marcocastignoli (approved):

PR #1611, opened 2024-09-03
Remove chains 82 and 83 from Monitor

kuzdogan:


manuelwedler:
What's the reason for this?

kuzdogan:
Sorry I pushed this in hurry. The reason is RPCs of these chains are not responding in our monitor. They constantly result "None of the RPCs responded"... I guess we're being blocked

marcocastignoli (approved):

Issue #1614, opened 2024-09-04
Support for the new tables `source_codes` and `compiled_contracts_sources`


marcocastignoli:
We are in the process of changing VerA and Sourcify database by creating two new tables: `source_codes` and `compiled_contracts_sources`. The new tables will replace the `compiled_contracts.sources` column.

![Image](https://github.com/user-attachments/assets/572fc691-3094-4569-83a9-27f2166a16e4)

kuzdogan:
Can you paste the final schema image here @marcocastignoli ? The above image is old

marcocastignoli:
> Can you paste the final schema image in the first message @marcocastignoli ? The above image is old

done

PR #1615, opened 2024-09-04
Add support for the new tables `sources` and `compiled_contracts_sources`

marcocastignoli:
See #1614 

This PR also implement the execution of insert and update queries inside an SQL transaction. I included also two tests (one for insert, one for update) to check if the database doesn't change after a failed transaction. The insert/update of the `sourcify_matches` table is not included in the transaction because it is executed only if a `verified_contracts` is successfully created or updated.


manuelwedler (approved):
Good job!

PR #1616, opened 2024-09-06
Fix monitor rpc authentication

kuzdogan:
Previous changes #1603 does not include the necessary changes to parse and replace the API keys. The PR adds that.


manuelwedler (approved):
lgtm

PR #1619, opened 2024-09-09
Add tests for Monitor authenticateRpcs

kuzdogan:
I've noticed I forgot to push the commits to https://github.com/ethereum/sourcify/pull/1616 and merged without the tests.

Here are the tests for added coverage, written with AI


marcocastignoli (approved):
It's good, but let's remember for the future that we could implement something generic to pass custom http headers to rpc (like this: https://github.com/ethereum/sourcify/commit/88ef9c766351c63047db56b68c153ad8cb5ca147)

PR #1623, opened 2024-09-10
Fix matching with creation when there's constructor arguments

kuzdogan:
The current algorithm checks for equality between the normalizedRecompiledCreationBytecode vs the normalizedOnchain... one. This unfortunately always fails when there are constructor arg's in the onchain bytecode. Instead we should've checked with `startsWith` as we do in other places.

Found out when verifying the following contract. Prev. it was `creationMatch=null` and now it's `creationMatch=partial`.

0x379002701bf6f2862e3dfdd1f96d3c5e1bf450b6
Base Mainnet (8453)
Import from Etherscan


manuelwedler (commented):
Can we have a test case for this?

marcocastignoli (approved):

kuzdogan:
Added tests, please check again

PR #1625, opened 2024-09-10
Fix the check for already existing partial match

marcocastignoli:
Saving a new match with creationMatch better than the existing one doesn't work. 

There are two places in which we check for already existing partial match:

1. In `storeMatch` inside `StorageService.ts`
  https://github.com/ethereum/sourcify/blob/498821296528660df41eb7a4c7bbe83b15245491/services/server/src/server/services/StorageService.ts#L246-L258

2. In `updateExistingVerifiedContract` inside `AbstractDatabaseService.ts`
  https://github.com/ethereum/sourcify/blob/d7c3ac708ce8badc09c369edd3e74a0260379753/services/server/src/server/services/storageServices/AbstractDatabaseService.ts#L452-L481

The second case is working fine but the execution can't reach it because it exits when reaching the first case.

We cannot remove the first case because otherwise we could have different contracts stored for each storage services, we need a centralized check in `storeMatch` before continuing calling every enabled storage service.

So in `storeMatch`, instead of evaluating the combined match status, we can verify if either the runtime or creation match is better, and ensure that neither the runtime nor creation is worse than the previous match.

### TODO

- [ ] Add tests for updating contracts to better matches


services/server/src/server/services/StorageService.ts
```diff
@@ -230,6 +234,52 @@ export class StorageService {
...
+    const statusPriority = [null, "partial", "perfect"];
+    const status1Priority = Math.max(
+      0,
+      statusPriority.findIndex((status) => status === status1),
+    );
+    const status2Priority = Math.max(
+      0,
+      statusPriority.findIndex((status) => status === status2),
+    );
+    return status1Priority - status2Priority;
```
manuelwedler, line 252: I find this function overly complicated.
```suggestion
    const scores = {
       "partial": 1,
       "perfect": 2,
    }
    return scores[status1] - scores[status2]
```

manuelwedler (commented):
I think your approach is right, but I could also be missing something. I'd say write some tests to see if this is fine.

services/server/test/integration/verification-handlers/verify.stateless.spec.ts
```diff
@@ -241,6 +242,7 @@ describe("/", function () {
       null,
       chainFixture.defaultContractAddress,
       chainFixture.chainId,
+      "perfect",
```
marcocastignoli, line 245: It was already default, I think it's better for it to be explicit

services/server/src/server/controllers/repository/repository.routes.ts
```diff
@@ -142,7 +143,7 @@ router
 // This route covers constructor-args.txt, creator-tx-hash.txt, library-map.json, immutable-references.json files
 router
   .route("/repository/contracts/:match/:chain/:address/*")
-  .get(safeHandler(getFileEndpoint));
+  .get(validateAddress, getFileEndpoint);
```
manuelwedler, line 146: Why is this no `safeHandler` anymore?

services/server/test/integration/verification-handlers/verify.stateless.spec.ts
```diff
@@ -172,7 +174,47 @@ describe("/", function () {
...
+      serverFixture.sourcifyDatabase,
+      null,
+      res,
+      null,
+      chainFixture.defaultContractAddress,
+      chainFixture.chainId,
+      "perfect",
+    );
+    // If the contract is perfectly verified (both creation and runtime), the storageTimestamp is returned
+    chai.expect(res.body.result[0].storageTimestamp).to.exist;
```
manuelwedler, line 214: I am not sure if this is the right way to check if compilation was skipped. This test is based on the assumption that the storageTimestamp is given in such a situation. But I would rather explicitly check that the compiler is not called. You can do that with sinon.js (I added it in a PR which then was not merged). A spy is the right concept for this I think.

manuelwedler (commented):
Looks good overall. I like how some functionalities are moved to middlewares now :)
Just three minor comments to address.

services/server/src/server/services/utils/util.ts
```diff
@@ -53,3 +54,45 @@ export async function readFile(
...
+ * is better than the second one; 0 if they are the same; or a
+ * negative number if the first Status is worse than the second one
+ */
+export function getStatusDiff(status1: Status, status2: Status): number {
+  const scores = {
+    error: 0,
+    "extra-file-input-bug": 0,
+    partial: 1,
+    perfect: 2,
+  };
```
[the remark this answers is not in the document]
marcocastignoli: I get this ts error while trying to do that

image

services/server/src/server/controllers/repository/repository.routes.ts
marcocastignoli: Uuuuups forgot that, good catch 🙏

services/server/src/server/services/utils/util.ts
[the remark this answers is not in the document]
manuelwedler: Ah okay makes sense. You can do a `String(status)` to make typescript happy, but take the approach you prefer here.

Btw this is an example for why I don't like mixing types as with the `Status` here.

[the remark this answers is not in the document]
marcocastignoli: I agree with you regarding the mixing types!

manuelwedler (approved):
Looks good to me :)
Could you only please add sinon.js to the dev dependencies of the server before merging? This only works at the moment because the monitor has it installed

PR #1637, opened 2024-09-17
Reusable server module

RaoulSchaffranek:
### Overview

This PR aims to make the `sourcify-server` module available as a library that can be distributed via package registries like npm.

### Background

We are developing a Solidity debugger that utilizes Sourcify as a source-code repository. Our goal is to leverage Sourcify for debugging public transactions on mainnet and testnets, as well as on local testnets such as Anvil. Given the complexity of installing Sourcify, we plan to bundle a Sourcify server within our VSCode extension and manage the server process ourselves for local development.

### Changes Summary

To achieve this, the following changes have been made:

1. **Modularization of the Server Module:**
   - The server module has been divided into two parts to avoid side effects like reading configuration files upon import:
     - A declarative module that exports the server class (this is the original server module).
     - A CLI module that handles configuration file reading and server startup.

2. **Splitting the sourcify-chains Module:**
   - Similarly, the `sourcify-chains` module has been split into:
     - A declarative module named `sourcify-chain-repository`.
     - An imperative module called `sourcify-chains`.

3. **Decoupling from node-config:**
   - All cross-references to `node-config` have been removed. Configuration files are now read exclusively by the CLI module. All other modules explicitly pass down their configuration options.

This modular approach allows us to integrate Sourcify more seamlessly into our project. We tried to keep the changes at a minimum and barely modified the existing code. The biggest diffs stem from moving imperative code to the cli module.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `ChainRepository` class for managing blockchain networks and validating chain IDs.
	- Added a command-line interface for running the server.
- **Improvements**
	- Refactored session management for enhanced flexibility.
	- Updated server configuration to streamline options and improve maintainability.
	- Enhanced Solidity compiler interactions by passing instances directly to functions.
- **Bug Fixes**
	- Corrected storage configuration in the README for clarity.
- **Chores**
	- Removed unused imports and simplified code across various files.


services/server/src/server/server.ts
```diff
@@ -70,13 +73,23 @@ export class Server {
...
+    this.port = options.port;
     logger.info("Server port set", { port: this.port });
     this.app = express();
 
+    this.chainRepository = new ChainRepository(options.chains);
+
+    this.app.set('chainRepository', this.chainRepository);
+    this.app.set('solc', options.solc);
+    this.app.set('verifyDeprecated', options.verifyDeprecated);
+    this.app.set('sessionOptions', options.sessionOptions);
```
kuzdogan: We don't use this anywhere
```suggestion
```

services/server/src/server/cli.ts
```diff
@@ -0,0 +1,241 @@
...
+  process.env.AWS_SECRET_ACCESS_KEY === undefined
+) {
+  throw new Error(
+    "AWS credentials not set. Please set them to run the compiler on AWS Lambda.",
+  );
+}
+
+let selectedSolidityCompiler: ISolidityCompiler;
+if (config.get("lambdaCompiler.enabled")) {
+  logger.info("Using lambda solidity compiler with local fallback");
```
kuzdogan, line 55: This throwing when no credentials found should only be if `lambdaCompiler.enabled`, so move it inside if `lambdaCompiler.enabled`.

```suggestion
let selectedSolidityCompiler: ISolidityCompiler;
if (config.get("lambdaCompiler.enabled")) {
  logger.info("Using lambda solidity compiler with local fallback");

  if (
    process.env.AWS_REGION === undefined ||
    process.env.AWS_ACCESS_KEY_ID === undefined ||
    process.env.AWS_SECRET_ACCESS_KEY === undefined
  ) {
    throw new Error(
      "AWS credentials not set. Please set them to run the compiler on AWS Lambda.",
    );
  }
```

services/server/src/server/server.ts
```diff
@@ -324,74 +337,6 @@ export class Server {
   }
 }
```
kuzdogan, line 342: I imagine some people will assume running `dist/server.js`. For compatibility, should we add something like below? Would this cause any problems?

```ts
if (require.main === module) {
  import "./cli"
}
```

kuzdogan (changes requested):
Couple small things, otherwise lgtm! Thanks a lot. 

Tbh I didn't know you could do `app.get('var')` so this makes things a lot easier.

I made some changes directly myself mainly:
- getting rid of injecting `req.services` after learning `app.get('services')`
- Renaming `repoPath` to `solcRepoPath` and also adding missing `solcJsonRepoPath`s

Added couple comments for your review. I still need to look at the tests if they'll work fine, let's see.

services/server/src/server/server.ts
RaoulSchaffranek: Fixed in [ac173f1](https://github.com/ethereum/sourcify/pull/1637/commits/ac173f1bb615fd73c7982d75715860eb2b841529)

services/server/src/server/cli.ts
RaoulSchaffranek: Fixed in [6447916](https://github.com/ethereum/sourcify/pull/1637/commits/6447916591a4b914ce9ae43588a931bcea8b0d7a)

kuzdogan:
Tests are passing if you want to check the comments @RaoulSchaffranek 

I'll handle the conflicts and the rest

marcocastignoli:
@kuzdogan 

I see we added so even if we run `node/dist/server.js` everything works as expected.

https://github.com/RaoulSchaffranek/sourcify/blob/4d77f987b0506d5cf946845e8fbf36cc395dd5ca/services/server/src/server/server.ts#L341-L343
```ts
if (require.main === module) {
  import("./cli");
}
```

It's ok, but I prefer to have only one entry point to reduce confusion.

E.g. in this Dockerfile running `"./dist/server/server.js"` works, but then from the `package.json` we use `node ./dist/server/cli.js`. I think it's confusing for new users
https://github.com/ethereum/sourcify/blob/cfc95f3473895c4541046743076e54cee50f226e/services/server/Dockerfile.debug#L11

services/server/src/server/server.ts
marcocastignoli: I see we added so even if we run `node/dist/server.js` everything works as expected.

https://github.com/RaoulSchaffranek/sourcify/blob/4d77f987b0506d5cf946845e8fbf36cc395dd5ca/services/server/src/server/server.ts#L341-L343
```ts
if (require.main === module) {
  import("./cli");
}
```

It's ok, but I prefer to have only one entry point to reduce confusion.

E.g. in this Dockerfile running `"./dist/server/server.js"` works, but then from the `package.json` we use `node ./dist/server/cli.js`. I think it's confusing for new users
https://github.com/ethereum/sourcify/blob/cfc95f3473895c4541046743076e54cee50f226e/services/server/Dockerfile.debug#L11

services/server/src/server/services/compiler/local/solidityCompiler.ts
```diff
@@ -174,26 +176,28 @@ export async function useCompiler(
...
-    solcJson.settings = {
+  if (!solcJsonInput.settings) {
+    solcJsonInput.settings = {
       outputSelection: outputSelection,
     };
   }
-  solcJson.settings.outputSelection = outputSelection;
-  const compiled = await useCompiler(compilerVersion, solcJson);
+  solcJsonInput.settings.outputSelection = outputSelection;
+  const compiled = await solc.compile(compilerVersion, solcJsonInput);
```
marcocastignoli, line 200: Why here it's calling `solc.compile` instead of `useCompiler`?

marcocastignoli (changes requested):
Looks good! Thanks @RaoulSchaffranek for the big PR! I have only one doubt regarding the compiler.

Also, I moved `process.env.LOGGING_TOKEN`  and `hideIpInLogs` into Server's options.

kuzdogan:
@coderabbitai review

services/server/src/server/services/compiler/local/solidityCompiler.ts
kuzdogan: Good catch. Idk tbh. I think I saw there's an error there (`useCompiler` is missing the 2 new parameters) and I just went with this fix. Good catch, thanks

services/server/src/server/server.ts
kuzdogan: Hmm ok right. Let's just use `./cli.js`

services/server/src/server/services/compiler/local/solidityCompiler.ts
kuzdogan: Actually no, this function should work with any compiler, not only with the `useCompiler` function found in the same file i.e. the local solidityCompiler. So it makes sense to use `solc` there. 

The problem is, this function shouldn't have been in this file, but in the VerificaitonService or another common place. Because it can be used via the lambda compiler as well, and does not only work with the local compiler.

marcocastignoli: Right, `getAllMetadataAndSourcesFromSolcJson` shouldn't be in that file. 

That function exists only because `CheckedContract` needs metadata as input, just to convert it back into json input. I think we can also leave it there for now and implement allowing passing json input to `CheckedContract` soon. Is there any other reason why `getAllMetadataAndSourcesFromSolcJson` exists?

services/server/src/server/controllers/verification/solc-json/stateless/solc-json.stateless.handlers.ts
```diff
@@ -42,9 +52,9 @@ export async function verifySolcJsonEndpoint(req: Request, res: Response) {
     );
   }
 
-  const match = await req.services.verification.verifyDeployed(
+  const match = await services.verification.verifyDeployed(
     contractToVerify,
-    chain,
+    chainRepository.sourcifyChainMap[chain],
```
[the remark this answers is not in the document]
marcocastignoli: I think this is a valid point from someone new looking at Sourcify code. We should name already validated (by openapi validator) parameters in the right way. For example in this case instead of `chain` something like `validatedChain`. So that variable it is explicitly already validated.

@kuzdogan what do you think? We can do it in another PR.

services/server/src/sourcify-chains.ts
```diff
@@ -186,88 +185,4 @@ if (missingChains.length > 0) {
...
-}
-
-export {
-  sourcifyChainsMap,
-  sourcifyChainsArray,
-  supportedChainsMap,
-  supportedChainsArray,
-  LOCAL_CHAINS,
-};
+export { sourcifyChainsMap };
```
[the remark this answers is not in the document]
marcocastignoli: Good catch here! Finally corerabbit catched something 👍 

`services/server/test/chains/deployContracts.ts` has a missing import

services/server/src/server/server.ts
```diff
@@ -164,7 +163,7 @@ export class Server {
               // This is a placeholder token. In a real application, use a more secure method for managing and validating tokens.
               const token = authHeader && authHeader.split(" ")[1];
 
-              return token === process.env.SETLOGGING_TOKEN;
+              return token === options.loggingToken;
             },
           },
         },
```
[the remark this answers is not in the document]
marcocastignoli: @kuzdogan how do we want this to work? Because this was the previous behavior.

Should we force the user to always provide a loggingToken?

services/server/src/server/server.ts
RaoulSchaffranek: I don't have hard feelings for any option. Antoher option to keep backwards compatibility could be renaming `server.ts` to something like `app.ts` and then `cli.ts` to `server.ts`.

One additional comment about `require.main === module`: This [StackOverflow answer](https://stackoverflow.com/questions/45136831/node-js-require-main-module#answer-60309682) mentions some compatibility issue when the module is loaded as a EcmaScript module. Not sure, if this is relevant and how this would affect the execution.

RaoulSchaffranek:
> Tests are passing if you want to check the comments @RaoulSchaffranek

Thanks for pushing this forward. I gave this PR another pass, and I believe I addressed all the comments you had for me. Could you let me know if I missed something?

services/server/src/server/services/compiler/local/solidityCompiler.ts
kuzdogan: It's not there **just** to convert it back to json input but a CheckedContracts is based on a metadata. The metadata file nicely describes what sources a contract needs so we can "validate" if there are sources missing etc. Without that there's no way to find it out without compiling. With a given std-json this whole "one metadata-one CheckedContract" model fails because each *contract* (a file can have multiple contracts) is a contract to compile.

services/server/src/server/controllers/verification/solc-json/stateless/solc-json.stateless.handlers.ts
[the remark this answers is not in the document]
kuzdogan: I think this is part of the 'middleware' solution. Chain validation is required for almost every endpoint, so we can add a middleware to the endpoints using the SourcifyChain to validate the chain and inject it into `req`. I don't think we should do this now and we can just accept it now

services/server/src/server/server.ts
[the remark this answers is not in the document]
kuzdogan: I think this is the correct fix? I don't think it should be required, but only required if you want to modify the log levels.

kuzdogan:
@RaoulSchaffranek all good from your side, thanks. Just working out the final details and that everyone in the team becomes aware of the changes. Thankyou!

services/server/src/server/server.ts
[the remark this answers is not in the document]
marcocastignoli: Isn't the api to change logs level always enabled? (Even with an undefined `loggingToken`)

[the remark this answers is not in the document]
kuzdogan: Yes, but in this case it will always fail if you haven't set it up. IMO it's fair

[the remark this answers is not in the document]
marcocastignoli: Are we sure?

```
            BearerAuth: (req) => {
              const authHeader = req.headers["authorization"];
              // This is a placeholder token. In a real application, use a more secure method for managing and validating tokens.
              const token = authHeader && authHeader.split(" ")[1];

              return token === options.loggingToken;
            },
```

If authHeader is undefined then `token` is undefined, then `token === options.loggingToken` will be true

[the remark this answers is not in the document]
kuzdogan: Yes that's true before the fix, hence the fix above solves it:
```ts
if (!options.loggingToken) {
   return false;
 }
```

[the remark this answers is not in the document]
marcocastignoli: Aaaah you were referring to that fix, not mine. Okok then, let's add that

kuzdogan (approved):

kuzdogan:
@marcocastignoli need approval here

marcocastignoli (approved):

PR #1638, opened 2024-09-18
Merge Master back to staging

kuzdogan:


manuelwedler (approved):
For the future, I think it would make more sense to merge the changelog and version updates first into staging. That way we have another quick review before the release and it would not be needed to merge master back into staging. What do you think?

kuzdogan:
@manuelwedler One thing that might be a problem is we need to choose the version (major, minor...) already when writing the changelogs. Let me try to walk through an alternative process from your suggestion.

- Open a PR from staging to master
- **Locally go to staging,** choose package versions, write and commit changelogs
- push changelogs to origin/staging
- review the PR including the changelogs
- Approve the PR
- Merge staging to master locally
- switch to master
- choose the versions in lerna (this will update inter-dependencies and create tags)
- push the tags to origin
- push the commits to origin/master, this will merge the PR automatically
- create GH releases

I think this would work. 

Currently what we do:
- Open a PR from staging to master
- merge staging to master locally
- choose package versions, write and commit changelogs (on master)
- choose the versions in lerna
- push the tags to the origin
- push the commits to origin/master, this will merge the PR
- create GH releases
- Merge master back to staging (this now requires another PR)

manuelwedler:
> * Merge staging to master locally
> * switch to master
> * choose the versions in lerna (this will update inter-dependencies and create tags)

Maybe I'm missing how the Lerna update functionality works. Why do we need to run the lerna version update on local master?

kuzdogan:
Because we need the latest and to-be-published commit hash to be the commits of the released version tags.

PR #1652, opened 2024-09-23
Support custom schema in AbstractDatabaseService and database session

marcocastignoli:
With the new Verifier Alliance database schema naming system, the schemas are named `v0`, `v1`,...  We should add support to custom schemas in AbstractDatabaseService.

This PR adds the support to custom schemas also to the database session

Open questions: 
- Manually adding `"schema"."table"` before every table is the only way I found to implement this. I'm open to suggestions

### DIFF
--- a/services/server/src/server/services/storageServices/AbstractDatabaseService.ts
+++ b/services/server/src/server/services/storageServices/AbstractDatabaseService.ts
@@ -27,6 +27,7 @@ export interface DatabaseServiceOptions {
     user: string;
     password: string;
   };
+  schema?: string;
 }
 
 export default abstract class AbstractDatabaseService {
@@ -41,6 +42,7 @@ export default abstract class AbstractDatabaseService {
   postgresDatabase?: string;
   postgresUser?: string;
   postgresPassword?: string;
+  schema: string = "public";
 
   constructor(options: DatabaseServiceOptions) {
     this.googleCloudSqlInstanceName = options.googleCloudSql?.instanceName;
@@ -51,6 +53,9 @@ export default abstract class AbstractDatabaseService {
     this.postgresDatabase = options.postgres?.database;
     this.postgresUser = options.postgres?.user;
     this.postgresPassword = options.postgres?.password;
+    if (options.schema) {
+      this.schema = options.schema;
+    }
   }
 
   async init() {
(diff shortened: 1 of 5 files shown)


kuzdogan:
This way also seems fine but have you looked at the `search_path` option:?

https://stackoverflow.com/a/70047912

Looks like you can do something like this:
```sql
SET search_path TO myschema, public;
```

kuzdogan:
This structure is perfectly fine for now, but on a side note, if we're using the same function parameter over and over again (like here with `pool` and `schema`) IMO it's an indication that this should actually be a class. Then we can access these var with `this.pool`, `this.schema`.

marcocastignoli:
I didn't want to touch `search_path` because it is already the configuration value that we use to fix the "extensions installed globally".

Also I think it's better if the schema is set explicitly by an environment variable.

kuzdogan:
> I didn't want to touch `search_path` because it is already the configuration value that we use to fix the "extensions installed globally" problem.

Can you explain this?

services/server/src/server/services/storageServices/AbstractDatabaseService.ts
```diff
@@ -51,6 +53,9 @@ export default abstract class AbstractDatabaseService {
     this.postgresDatabase = options.postgres?.database;
     this.postgresUser = options.postgres?.user;
     this.postgresPassword = options.postgres?.password;
+    if (options.schema) {
+      this.schema = options.schema;
+    }
```
kuzdogan, line 58:
```suggestion
    } else {
      logger.warn(`Database schema name not set, using default "public"`, { identifier: this.IDENTIFIER, schema: options.schema })
    }
```

kuzdogan (approved):
lgtm. Just a small suggestion

marcocastignoli:
> Can you explain this?

When you install an extension in PostgreSQL, it is installed globally, but it’s only accessible to the schemas included in the `search_path`. To ensure a consistent `search_path` across all users on the Verifier Alliance database, I executed the following command:

```sql
ALTER DATABASE "verifier-alliance" SET search_path = "$user", v1, v0, pg_catalog;
```

Including `"$user"` and `pg_catalog` is essential because `"$user"` allows for user-specific schemas to be prioritized, which is important when different users might have their own schemas. `pg_catalog` is necessary because it provides access to system tables and functions, which are critical for the database's internal operations.

This is why I decided not to modify the `search_path` to address this issue.

services/server/src/server/services/storageServices/AbstractDatabaseService.ts
marcocastignoli: Abstract properties cannot be accessed in the constructor, I'll log the schema somewhere else

PR #1658, opened 2024-09-24
Change `s3-backup-check` to check not the contracts from the previous day, but from 2 days before

marcocastignoli:
`s3-backup-check` is failing too often because the s3 sync process is taking more than 24h to complete. As a workaround let's change the test to check not the contracts from the previous day, but from 2 days before.

This PR also include prettier fixes


manuelwedler (approved):
lgtm

PR #1663, opened 2024-09-25
routescanApi support for getCreatorTx

### LINKED ISSUE #1650: Fetching creatorTx’s from routescan
Similar to #1607 we can also use Routescan to fetch the creator tx hashes instead of Etherscan instances.

kuzdogan:
The PR adds support for fetching creation tx's from the routescan APIs.

Fixes #1650 

- Adds the `routescanApi` fields to the chains found in https://routescan.io/rpcs manually
- Adds the handler function for `routescanApi` case in `contract-creation-util.s`.
- Adds the test for the `routescanApi` method. In the process, I also consolidated `etherscanApi` and `routescanApi` tests and added the missing `blockscoutApi` and `avalancheApi` tests. Previously, the `etherscanApi` test was not actually getting executed because in reality the `blockscoutApi` takes precedence. Now we make sure there are no other methods in the `SourcifyChain` before we run the test.


kuzdogan:
I don't understand why the coverages are failing here. The related lines are running for sure as can be seen from the test logs:

```
Fetching Creator Tx - fetcher={"type":"api","url":"https://api.routescan.io/v2/network/mainnet/evm/1/etherscan?module=contract&action=getcontractcreation&contractaddresses=${ADDRESS}"}
...
```

Can I get a review and a short look to why the lines are not getting covered? I might be missing something @manuelwedler

manuelwedler:
Hm yes that's weird indeed. Is it possible that the output of `test:unit` is overwritten `test:integration` since the latter is run afterwards?

kuzdogan:
Yup exactly

Inside `services/server` the result of `npm run test:unit && npm run test-local:integration && npm run cov:html`:
image

vs. `npm run test:unit && npm run cov:html`:
image

Anything from the top of your mind? If not I'll think about it

manuelwedler:
I think we should merge the two test commands again. I don't see a way to run c8 in an "incremental" way. 

We can keep the `unit` and `integration` folders, but I think we have to make the command `npx c8 --reporter=none mocha --exit test/unit/** test/integration/**`. You can check again if there is another way but I don't see it at the moment

manuelwedler (approved):
There many indirect changes in codecov (https://app.codecov.io/gh/ethereum/sourcify/pull/1663/indirect-changes). I don't understand how they happened, but looks more like a configuration change. Still, I think this PR is good now

kuzdogan:
The project coverage is still failing but I think it's good to go for a review @manuelwedler

PR #1648, opened 2024-09-23
Remove .sol extension in repositoryV2

kuzdogan:
Following the discussion: https://discord.com/channels/1160912737539473410/1160912738374127689/1287445616972140565

I don't see any downsides of this. If you think of something lmk.

PR #1664, opened 2024-09-25
Add rename_repov2_sol_extension.sh script

manuelwedler:
Adding the script to remove the .sol extension from contracts in repoV2. I will run it now after merging https://github.com/ethereum/sourcify/pull/1648 on staging. On production, I will run it once we do a new release.


marcocastignoli (approved):
lgtm

PR #1669, opened 2024-09-27
Fixing Monitor crashing regularly, small additional fixes

kuzdogan:
The Monitors were crashing because we were throwing here

https://github.com/ethereum/sourcify/blob/33cc02ec655596b8f1001ec52c2b0c4ab987f91a/services/monitor/src/ChainMonitor.ts#L348-L351

- Removes the throw
- Fixes the Avalanche RPC URLs
- Removes the launch.json built and run Monitor

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -35,22 +35,6 @@
     {
       "type": "node",
       "request": "launch",
-      "name": "Monitor",
-      "cwd": "${workspaceFolder}/services/monitor",
-      "program": "dist/index.js",
-      "args": ["--chainsPath=chains-dev.json"],
-      "envFile": ".env",
-      "outFiles": [
-        "${workspaceFolder}/services/monitor/dist/**/*.js",
-        "${workspaceFolder}/packages/**/build/**/*.js"
-      ],
-      "smartStep": true,
-      "console": "integratedTerminal",
-      "outputCapture": "std"
-    },
-    {
-      "type": "node",
-      "request": "launch",
       "name": "Monitor - without build",
       "cwd": "${workspaceFolder}/services/monitor",
       "program": "dist/index.js",
--- a/services/monitor/monitorChains.json
+++ b/services/monitor/monitorChains.json
@@ -139,7 +139,7 @@
     "rpc": [
       {
         "type": "ApiKey",
-        "url": "https://ava-mainnet.blastapi.io/{API_KEY}",
+        "url": "https://ava-mainnet.blastapi.io/{API_KEY}/ext/bc/C/rpc",
         "apiKeyEnvName": "BLAST_API_KEY"
       }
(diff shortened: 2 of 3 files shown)


manuelwedler:
Why shouldn't we throw in that place?

kuzdogan:
@manuelwedler Why should we? The monitor was able to successfully assemble the contract and send a request to the Sourcify server. It's a problem either on the network (probably this in our case) or an error of the Sourcify server. I don't think this should stop the Monitor from running. 

We already log an error and actually we need to pay attention to that if there's an increase in errors. This wouldn't work now as we have plenty of these errors and need to fix it first.

marcocastignoli:
I agree with @kuzdogan that the monitor should not be stopped for that error. When I implemented it I was confused by the try-catch at the top of the callstack

https://github.com/ethereum/sourcify/blob/460fc8c2f61209f0725aca8b58c0f8245dc05a24/services/monitor/src/ChainMonitor.ts#L78-L106

Since this is an event being triggered, the error it's not catched. Instead of not throwing, I think it makes more sense to add another try-catch inside `processBlockListener`, so we have a central place in which all errors are catched inside the event.

manuelwedler:
> @manuelwedler Why should we? The monitor was able to successfully assemble the contract and send a request to the Sourcify server. It's a problem either on the network (probably this in our case) or an error of the Sourcify server. I don't think this should stop the Monitor from running.
> 
> We already log an error and actually we need to pay attention to that if there's an increase in errors. This wouldn't work now as we have plenty of these errors and need to fix it first.

Okay makes sense to me. I didn't understand that the monitor was crashing for it. Thanks for explaining.

kuzdogan:
Is this good to go? I need an approval

manuelwedler:
> Since this is an event being triggered, the error it's not catched. Instead of not throwing, I think it makes more sense to add another try-catch inside `processBlockListener`, so we have a central place in which all errors are catched inside the event.

I think it still makes sense to address this

kuzdogan:
Yeah I missed the lower part of that comment.

But I don't really understand the point there.

> Since this is an event being triggered, the error it's not catched. Instead of not throwing, I think it makes more sense to add another try-catch inside processBlockListener, so we have a central place in which all errors are catched inside the event.

Where's the event? The event mechanism is inside the `DecentralizedSouceFetcher`. `processBlockListener` is also a listener function being triggered on a new block. I don't see the point in throwing inside `processsNewContract` just to catch it outside it i.e. in `processBlockListener`, the only place it's being called.

manuelwedler (approved):
I had a quick look at the full code and think it is actually fine to remove the throw here to fix this issue. I don't really get how Marco's comment was meant to be implemented / how it would improve this PR. So I'm approving for moving forward.

Issue #1640, opened 2024-09-19
Update url handling in SourcifyDatabase to support SourcifyDB-only and SourcifyDB+repov2 configurations


marcocastignoli:
Currently, the [`getTree()` function](https://github.com/ethereum/sourcify/blob/6156037dde2374260830fd832b6f9831d16bf1fe/services/server/src/server/services/storageServices/SourcifyDatabaseService.ts#L416-L428) retrieves the file tree from `repositoryV1`, which forces you to run RepositoryV1Service (+ an http server) together with SourcifyDatabase. This prevents running sourcify with SourcifyDatabaseService alone or in combination with `RepositoryV2Service`.

In the current implementation, URLs are constructed using `repositoryV1.serverUrl`. We should directly use the `https://sourcify.dev/server/repository/contracts/{full_match | partial_match}/{chain}/{address}/{filePath}` API for URL construction.

kuzdogan:
Ah ok, I thought we were reading the literal file tree from `repositoryV1` like `tree -L 2 /Users/kuzdogan/sourcify/repositoryV1/contracts/full_match/1/0x025030BdAa159f281cAe63873E68313a703725A5` and that's why I preferred not working on this now.

If we are using that for the URLs only and it's a simple fix, we can do this after #1637

marcocastignoli:
Fixed with #1677

PR #1677, opened 2024-10-01
Use server public url in getTree instead of repositoryV1 url

marcocastignoli:
See #1640

This PR also adds a list of all the available Sourcify configs in the readme file.


manuelwedler (approved):
lgtm

Issue #1635, opened 2024-09-17
Plan migration from Sourcify DB v0 to v1: custom script vs direct SQL script


marcocastignoli:
With the new VerA schema (https://github.com/verifier-alliance/database-specs/pull/15), we need to migrate existing `compiled_contracts.sources` to the new tables `sources` and `compiled_contracts_sources`. Ensuring data integrity and minimizing downtime are critical for a smooth transition.

Evaluate whether to develop a custom js script or to execute a direct migration using an SQL script.

marcocastignoli:
```sql
-- Begin the transaction
BEGIN;

-- Step 1: Extract and insert unique sources into 'sources' table
WITH extracted_sources AS (
    SELECT
        cc.id AS compilation_id,
        src.key AS path,
        src.value #>> '{}' AS content, -- Extracts the text without quotes
        digest(src.value #>> '{}' , 'sha256') AS source_hash,
        ''::bytea AS source_hash_keccak -- Ensure keccak256 function exists
    FROM
        compiled_contracts cc
    CROSS JOIN LATERAL
        jsonb_each(cc.sources::jsonb) AS src(key, value)
    ORDER BY cc.id
    LIMIT 100
),
inserted_sources AS (
    INSERT INTO sources (
        source_hash,
        source_hash_keccak,
        content
    )
    SELECT DISTINCT
        es.source_hash,
        es.source_hash_keccak,
        es.content
    FROM
        extracted_sources es
    ON CONFLICT (source_hash) DO NOTHING
    RETURNING source_hash
),
inserted_links AS (
    INSERT INTO compiled_contracts_sources (
        compilation_id,
        source_hash,
        path
    )
    SELECT
        es.compilation_id,
        es.source_hash,
        es.path
    FROM
        extracted_sources es
    ON CONFLICT (compilation_id, path) DO NOTHING
    RETURNING id
)
-- Main statement to execute the CTEs and retrieve summary counts
SELECT
    (SELECT COUNT(*) FROM inserted_sources) AS sources_inserted,
    (SELECT COUNT(*) FROM inserted_links) AS links_inserted;

-- Commit the transaction
COMMIT;
```

marcocastignoli:
On my machine it took 5 minutes to migrate 383'160 `compiled_contracts` to 148'865 `sources` and 2'659'011 `compiled_contracts_sources`

marcocastignoli:
[...]

PR #1681, opened 2024-10-01
Add script to finalize v1 migration

marcocastignoli:
After running the script in #1635, `sources.source_hash_keccak` is empty. Since Postgresql doesn't support keccak256, we need to manually update the column. In this PR, a script is implemented to update each `sources.source_hash_keccak`

### DIFF
--- a/services/database/schema-updates/post-v0-to-v1-upgrade.mjs
+++ b/services/database/schema-updates/post-v0-to-v1-upgrade.mjs
@@ -1,0 +1,95 @@
+// This script is used to finish the upgrade of the database schema from v0 to v1
+// The sources table is filled but the source_hash_keccak column is empty
+// This script fills the source_hash_keccak column with the keccak256 hash of the source
+
+import pg from "pg";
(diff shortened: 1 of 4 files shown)


services/database/schema-updates/post-v0-to-v1-upgrade.mjs
```diff
@@ -0,0 +1,95 @@
+// This script is used to finish the upgrade of the database schema from v0 to v1
+// The sources table is filled but the source_hash_keccak column is empty
+// This script fills the source_hash_keccak column with the keccak256 hash of the source
+
+import pg from "pg";
+import { keccak256 } from "ethers";
```
manuelwedler, line 6: I think `ethers` should be defined in the package.json of the `database` folder if used here.

manuelwedler (changes requested):
Nice use of a `do ... while` loop! Only some smaller comments

manuelwedler (approved):

PR #1684, opened 2024-10-02
Fix missing source after match upgrade

marcocastignoli:
Inside the function `AbstractDatabaseService.updateExistingVerifiedContract` we didn't call https://github.com/ethereum/sourcify/blob/c14282cbac1e6d5e407e309c7d7dcf935b78adc6/services/server/src/server/services/storageServices/AbstractDatabaseService.ts#L444-L449

So all match upgrades were not containing sources.


manuelwedler (approved):
looks good

Issue #1653, opened 2024-09-23
Refactor `database-util.ts` file into a class


marcocastignoli:
We are passing `databasePool` and `schema` to every function inside `database-util.ts`. It's better if we move all the functions that query the database into a class and then use its properties to access `databasePool` and `schema`.

PR #1689, opened 2024-10-08
Refactor database utils into a class

marcocastignoli:
See #1653


kuzdogan (changes requested):
Couple things:

I've noticed the schema env var was not documented in https://github.com/ethereum/sourcify/pull/1652/files Please add documentation for that including the env var and that it defaults to public.

I see we are no longer passing `this.schema` but we still pass the `client` in the Database functions. Can't we also remove that? `node-postgres` says: "In fact, pool.query delegates directly to client.query internally": https://node-postgres.com/features/queries **unless we are using a transaction**: https://node-postgres.com/apis/pool#poolquery Maybe we can move the client acquisition and handling all inside the `Database` and remove this from the `DatabaseService`s. 

As a side note, turns out we could pass `this.schema` in the SQL query with `pg-format`: https://node-postgres.com/features/queries#parameterized-query The docs say it can be vulnerable for injection. In our case it should be fine because `this.schema` is not a user input, right? So we don't need to change that part of the code now.

Can you also check the "Prepared statements" docs if this is something we can take advantage of? https://node-postgres.com/features/queries#prepared-statements

marcocastignoli:
> I've noticed the schema env var was not documented in https://github.com/ethereum/sourcify/pull/1652/files Please add documentation for that including the env var and that it defaults to public.

Right

> unless we are using a transaction

That's exactly why we are using it... "Maybe we can move the client acquisition and handling all inside the Database and remove this from the DatabaseServices." we can do this.

> The docs say it can be vulnerable for injection. In our case it should be fine because this.schema is not a user input, right? So we don't need to change that part of the code now.

Yes, everything is passed via arguments except this.schema, I don't think we should care about this.schema being vulnerable because it's not passed by the users

> Can you also check the "Prepared statements" docs if this is something we can take advantage of? https://node-postgres.com/features/queries#prepared-statements

Good idea

marcocastignoli:
> Maybe we can move the client acquisition and handling all inside the Database and remove this from the DatabaseServices

Actually regarding this it's not that straightforward because multiple transactions can be initialized from a Pool. We cannot remove the `PoolClient` parameter from the insert/updates functions because all of them need to be executed within the `PoolClient` that starter the transaction. The only improvement that we can do is the following

```ts
    let client: PoolClient;
    try {
      client = await this.database.beginTransaction();
      onchainCreationCodeInsertResult = await this.database.insertCode(
          client,
          databaseColumns.onchainCreationCode,
        );
} catch(e) {
    await this.database.rollbackTransaction(client!);
} finally {
    this.database.releaseClient(client!)
}
```

Personally I don't think this is much better than before... @kuzdogan wdyt?

kuzdogan:
Ok yes, I see. Moving the client acquisition inside `Database.ts` doesn't seem possible.

marcocastignoli:
@kuzdogan I added `SOURCIFY_POSTGRES_SCHEMA` and `ALLIANCE_POSTGRES_SCHEMA` in `.env.dev`.

PR #1693, opened 2024-10-11
Rpcs with trace

kuzdogan:
A draft PR to add RPCs with trace support, and refactoring the code related to processing traces. It turned into a larger-than-expected PR as 

1. I had to modify `SourcifyChains` to be able to mark which RPCs have trace support and which type of traces (`trace_transaction` or `debug_traceTransaction`).
2. I've encountered couple critical bugs

The bugs:
1. a10bdce5b44f8360623f4e6499b1b3db8b24f0ba: Turns out we were not extracting the correct field from `trace_transaction` type traces. Previously we were getting `trace.result.code` which was the resulting runtime bytecode of the contract, whereas we had to extract `trace.action.init` for the creation bytecode. 
2. e26aafa00f05285e2e5b186e1df216623cea989d: We were doing a double check on the address that this tx should yield (`keccak(tx.from, tx.nonce)`). This is a correct check since we don't have an inherent way to know that this `creatorTxHash` is indeed for the given address. However this only works and only needed for the contracts created with an EOA, i.e. when we take the tx payload as the creation bytecode. For factory created contracts that are created with `create2` this was causing the creationMatch to fail silently. We already have the yielding address in the trace, so we can use that for double checking if this `creatorTx` is for this contract. 

In fact we didn't need this check since also the EOA created contracts have the created address in the tx receipt:
https://github.com/ethereum/sourcify/blob/a10bdce5b44f8360623f4e6499b1b3db8b24f0ba/packages/lib-sourcify/src/lib/SourcifyChain.ts#L458-L461

### DIFF
--- a/packages/lib-sourcify/src/lib/SourcifyChain.ts
+++ b/packages/lib-sourcify/src/lib/SourcifyChain.ts
@@ -10,6 +10,7 @@ import {
   Chain,
   FetchContractCreationTxMethods,
   SourcifyChainExtension,
+  TraceSupportedRPC,
 } from './types';
 import { logDebug, logError, logInfo, logWarn } from './logger';
 
@@ -27,6 +28,7 @@ interface JsonRpcProviderWithUrl extends JsonRpcProvider {
 export type SourcifyChainInstance = Omit<Chain, 'rpc'> &
   Omit<SourcifyChainExtension, 'rpc' | 'sourcifyName'> & {
     rpc: Array<string | FetchRequest>;
+    traceSupportedRPCs?: TraceSupportedRPC[];
   };
 
 class CreatorTransactionMismatchError extends Error {
@@ -40,6 +42,8 @@ export default class SourcifyChain {
   title?: string | undefined;
   chainId: number;
   rpc: Array<string | FetchRequest>;
+  traceSupport?: boolean;
+  traceSupportedRPCs?: TraceSupportedRPC[];
   supported: boolean;
   providers: JsonRpcProviderWithUrl[];
   fetchContractCreationTxUsing?: FetchContractCreationTxMethods;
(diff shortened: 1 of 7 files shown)


marcocastignoli (approved):
lgtm! I think the system to store `traceSupportedRPCs` in a separate array is probably an overkill but it doesn't make the code too complicated. Maybe a couple comments in the code would make its use clear.

kuzdogan:
> lgtm! I think the system to store `traceSupportedRPCs` in a separate array is probably an overkill but it doesn't make the code too complicated. Maybe a couple comments in the code would make its use clear.

The reason I had a separate array for that is that the field `SourcifyChain.rpc`expects `Array<string | FetchRequest>`. This is because the `SourcifyChain.provider: JsonRpcProviderWithUrl` which is an `ethers` type, is initialized with either a `string` or `FetchRequest`:

https://github.com/ethereum/sourcify/blob/c41089c5bca0e0e1219e2e313991512b17b1aeed/packages/lib-sourcify/src/lib/SourcifyChain.ts#L78-L103

That's why I couldn't mark the trace support directly inside the `rpc` array's elements, the way I did it in `sourcify-chains.json`. 

I agree that it's not the cleanest and if you have any ideas happy to hear.

kuzdogan:
Quicknode endpoints both have subdomains and API keys, so I needed to add this option in `sourcify-chains.json`. Please have a short look. Then it's ready to merge

marcocastignoli (approved):
> Quicknode endpoints both have subdomains and API keys, so I needed to add this option in `sourcify-chains.json`. Please have a short look. Then it's ready to merge

I thought maybe we can move back to specify the chains inside a `.ts` file so that environment variables are available and we can do more advanced things such as creating factories for RPCs. e.g. 

```
{
...
"rpc": [
    getEthPandaOpsRPC("https://rpc.sepolia.ethpandaops.io")
]
...
}


// the code above will add to the rpc array

{
        "type": "FetchRequest",
        "url": "https://rpc.sepolia.ethpandaops.io",
        "headers": [
          {
            "headerName": "CF-Access-Client-Id",
            "headerEnvName": "CF_ACCESS_CLIENT_ID"
          },
          {
            "headerName": "CF-Access-Client-Secret",
            "headerEnvName": "CF_ACCESS_CLIENT_SECRET"
          }
        ]
      }
```


But let's discuss this change separately and implement this in another PR. LGTM!

Issue #1702, opened 2024-10-15
Flare Mainnet RPC working now


Uttam-Singhh:
Hello there,
Flare Mainnet Support was turned off as RPCs were failing but it should work now!

https://github.com/ethereum/sourcify/commit/c43fd285b72f42e82616d200db86e464bf970976#diff-e0d5a0ef165fc606c0002001139dda3be139ddba09a465873b8d83b185a5b7fdL697

Would appreciate if this could be fixed, please let me know if you need anything else to fix this.

<img width="967" alt="img 2024-10-16 at 12 49 19 AM" src="https://github.com/user-attachments/assets/d0205b65-5939-4c32-88c5-7f7d4cd8baf5">

manuelwedler:
Hey there, feel free to open a PR for enabling support again

Uttam-Singhh:
Sure 
Thank you

> Hey there, feel free to open a PR for enabling support again

PR #1704, opened 2024-10-17
Enabling Flare Mainnet Chain-14 support

Uttam-Singhh:
Enabling Flare Mainnet Chain-14 support as advised here #1702 

Flare Mainnet Support was turned off as RPCs were failing but it should work now!


manuelwedler (approved):
Thanks!

Issue #1661, opened 2024-09-24
Update docs and start publishing sourcify-server following the Reusable Server Module


kuzdogan:
The docs need to be updated for #1637 

Besides, should we start publishing the server on npm? cc: @RaoulSchaffranek

RaoulSchaffranek:
Yes, I'm in full support of publishing the server to npm. It makes it easier for downstream tools like our debugger to manage the dependency on Sourcify and stay up-to-date. Another advantage is that npm packages don't need to be rebuilt every time they're installed - which means less messing with manual build steps.

kuzdogan:
As said in #1708 I skipped adding the docs because it's rather complicated to set up a new Server() with options. I tried to do the fixes and it takes a lot of time. I'd postpone it until someone requests it and also until https://github.com/ethereum/sourcify/issues/1665

PR #1708, opened 2024-10-21
Publish sourcify-server and only publish the package in its own CI run

kuzdogan:
The changes adds the sourcify-server package to be published to npm.

It also adds a check to only publish the packages in their own tagged CI runs. Previously e.g.  when there was a new tag for theese packages, both lib-sourcify and bytecode-utils's CI runs tried to publish both packages. This results in the first one succeeding and publishing both packages but the second one to fail.

Now we check to only publish the package in their corresponding tag CI builds.

Related to #1661. Still need to add the docs how to use the server package.

Edit: TBH I'd skip adding the docs because it's rather complicated to set up a `new Server()` with options. I tried to do the fixes and it takes a lot of time. I'd postpone it until someone requests it and also until #1665

### DIFF
--- a/.circleci/scripts/publish_to_npm.sh
+++ b/.circleci/scripts/publish_to_npm.sh
@@ -1,35 +1,56 @@
 #!/bin/bash
 
-BYTECODE_UTILS_LOCAL_VERSION=$(cat packages/bytecode-utils/package.json \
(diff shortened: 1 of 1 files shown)


manuelwedler (approved):
lgtm :)

PR #1711, opened 2024-10-22
Add API-level proxy detection and implementation resolution

### LINKED ISSUE #1655: API-level proxy detection and implementation resolution
As it has been a feature that was requested frequently, we decided to implement proxy detection and resolution of the implementation contract as an API-level functionality. This is made possible by the [whatsabi](https://github.com/shazow/whatsabi) library which detects major proxy styles by only looking at the bytecode. As a first proof of concept we implement this in API v1 before we incorporate it in the API v2 design.

### Spec

On a request on the `check-all-by-addresses` endpoint, the bytecode is loaded from the db as before. We use the bytecode to call `whatsabi.autoload`. This will detect if the contract is a proxy and which type of proxy. It works without the need of rpc calls. If the address is a proxy we resolve the implementation address. This will make use of an rpc call depending on the type of proxy. 

The response of the API should look like this then:

```
[
  {
    "address": "0x6F1D75a53a8805DcA5347aE5F3eDE035CAE3CBC1",
    "chainIds": [
      {
        "chainId": "5",
        "status": "perfect",
        "isProxy": true,
        "proxyType": "EIP1967Proxy",
        "implementations": [
            {
                address: "0x751D7C0Cf91a9b7704541b44E5fF7BeC3D2caA6F",
                name: "Logic contract"
            }
         ]
      }
    ]
  }
]
```

If a contract is not verified, we don't have the bytecode. In this case we can let whatsabi fetch the bytecode as a fallback, and show the proxy status even if a contract is not verified. This is necessary to also support Minimal Proxies (https://github.com/ethereum/sourcify/issues/1624).

### ToDos

- [x] Estimate costs in terms of additional RPC calls (see https://github.com/ethereum/sourcify/issues/1655#issuecomment-2382836439)
- [x] Wait for next release of whatsabi that should include https://github.com/shazow/whatsabi/pull/120
- [x] Investigate the source code of `FixedProxy` contracts in our DB
- [x] Investigate multi proxy contracts and what implementations they resolve to
- [x] Decide what to return as implementation address for proxies that have multiple implementations
- [ ] Implement PoC in API v1

manuelwedler:
Closes #1655 

This implements an API-level proxy detection as in #1655 by making use of the [whatsabi](https://github.com/shazow/whatsabi) library. A few explanations about this PR:

- I had to add two methods to the `SourcifyChain` class for making it compatible to the whatsabi library as a provider, and for being able to resolve the implementation addresses of `DiamondProxy` contracts.
- I added a `proxyResolutionError` to the API response for the case that the proxy resolution fails. This could be due to an unresponsive RPC for example. In such a case, I still want to respond successfully to the user but let them know why the proxy fields are not there.
- Related to https://github.com/shazow/whatsabi/issues/142:
   - A `FixedProxyResolver` was returned in a lot of cases where actually a library was called by a `DELEGATECALL`. I decided to only support EIP1167 proxies for this resolver. Any other `FixedProxy` would be a non-standardised proxy (or even a library). So in case a `FixedProxyResolver` is detected, we check explicitly for the opcodes standardised in EIP1167.
   - I saw a few cases where multiple resolvers were returned, and also some cases where a resolver resolves to `0x0`. The latter was for example the case for a factory contract which deploys EIP1967 contracts. Therefore, I iterate over the returned resolvers and return the first implementation address which is not `0x0`.
- This feature is only added to the `/check-all-by-addresses` endpoint and not the `check-by-addresses` endpoint, because in the latter the `chainIds` array does not contain objects.

### Remaining issue (non-blocking)

There is still a problem left from https://github.com/shazow/whatsabi/issues/142:
A contract that embeds a proxy (as for example a factory contract) could falsely be detected as a proxy.

A first proposal to solve this is to not consider contracts with `CREATE`/`CREATE2` opcodes as proxies. However, this will not catch all contracts that only embed proxy bytecodes, and also exclude factory contracts that could be at the same time a proxy.
  
Edit: We decided that this problem is not a blocker for this PR. IMO, it is good to get this out and test it. The proxy data won't be persisted in our database with this change, so before we persist it we might be able to improve the proxy detection.


services/server/src/server/controllers/repository/repository.handlers.ts
```diff
@@ -128,9 +136,71 @@ export async function checkAllByChainAndAddressEndpoint(
...
+          type Implementation = { address: string; name?: string };
+          let proxyStatus: {
+            isProxy?: boolean;
+            proxyType?: ProxyType | null;
+            implementations?: Implementation[];
+            proxyResolutionError?: string;
+          } = {};
+          if (
+            req.query.resolveProxies === "true" &&
+            found[0].onchainRuntimeBytecode
```
kuzdogan, line 149: Just to make sure, here we use the `onchainRuntimeBytecode` from the Database right? Not fresh from an RPC

services/server/src/server/controllers/repository/repository.handlers.ts
```diff
@@ -128,9 +136,71 @@ export async function checkAllByChainAndAddressEndpoint(
...
+                      await services.storage.performServiceOperation(
+                        "checkAllByChainAndAddress",
+                        [implementationAddress, chainId],
+                      );
+                    if (
+                      implementationFound.length > 0 &&
+                      implementationFound[0].contractName
+                    ) {
+                      implementation.name = implementationFound[0].contractName;
+                    }
```
kuzdogan, line 177: From what I understand we only return `contractName` only if the implementation is verified on Sourcify. Otherwise we just return the address. 

If so we should note this down in the documentation or maybe have a separate field for `isVerified`

services/server/src/server/controllers/repository/repository.handlers.ts
```diff
@@ -128,9 +136,71 @@ export async function checkAllByChainAndAddressEndpoint(
...
+          }
+
+          map.get(address).chainIds.push({
+            chainId,
+            status: found[0].runtimeMatch,
+            ...proxyStatus,
+          });
         }
       } catch (error) {
         // ignore
```
kuzdogan, line 206: Not introduced by this PR but not a good idea to ignore silently. We should log here

kuzdogan (commented):
I'll further review after rebase

kuzdogan:
> Done. I think this was due to rewriting history. IMO, we should never do this again.

Right, but not never, IMO it was necessary in this case, unfortunately.

services/server/src/server/services/utils/proxy-contract-util.ts
kuzdogan, line 1: I don't fully understand what the code here does in `detectAndResolveProxy`. Some explanation and comments would help.

services/server/src/server/services/utils/proxy-contract-util.ts
```diff
@@ -0,0 +1,112 @@
...
+    codeCache,
+  );
+
+  const detectionResult = await whatsabi.autoload(address, {
+    provider: cachedCodeProvider,
+    abiLoader: false,
+    signatureLookup: false,
+    followProxies: false,
+  });
+  const proxies = detectionResult.proxies;
```
kuzdogan, line 43: For me it's not clear what we are doing after this. We have the `proxies`. Now what? Why are we doing the steps below?

services/server/src/server/services/utils/proxy-contract-util.ts
```diff
@@ -0,0 +1,112 @@
...
+  const fixedProxy = proxies.find(
+    (proxy) => proxy instanceof whatsabi.proxies.FixedProxyResolver,
+  );
+  if (fixedProxy && isEIP1167Proxy(bytecode, fixedProxy.resolvedAddress)) {
+    return {
+      isProxy: true,
+      proxyType: "EIP1167Proxy",
+      implementations: [fixedProxy.resolvedAddress],
+    };
+  } // Ignore FixedProxy otherwise because it could be a false-positive
```
kuzdogan, line 55: Why? Didn't get it.

services/server/src/server/services/utils/proxy-contract-util.ts
```diff
@@ -0,0 +1,112 @@
...
+      };
+    } catch (error) {
+      // Falsely detected as a diamond proxy
+    }
+  }
+
+  const checkedProxyTypes: Set<ProxyType> = new Set([
+    "FixedProxy",
+    "DiamondProxy",
+  ]); // Needed because of potential duplicates
```
kuzdogan, line 85: Also didn't get the potential duplicates issue here

services/server/src/server/services/utils/proxy-contract-util.ts
```diff
@@ -0,0 +1,112 @@
...
+  for (const proxy of proxies) {
```
kuzdogan, line 86: Above I see we handle the EIP1167 and DiamondProxy cases. Is this here the rest of the cases for other ProxyTypes?

services/server/src/server/controllers/repository/repository.handlers.ts
```diff
@@ -128,9 +136,71 @@ export async function checkAllByChainAndAddressEndpoint(
...
+          ) {
+            try {
+              const sourcifyChain = chainRepository.supportedChainMap[chainId];
+              const proxyDetectionResult = await detectAndResolveProxy(
+                found[0].onchainRuntimeBytecode,
+                address,
+                sourcifyChain,
+              );
+
+              const implementations = await Promise.all(
```
kuzdogan, line 159:
```suggestion
              // Find contract names if the implementations are verified on Sourcify
              const implementations = await Promise.all(
```

kuzdogan (changes requested):
Some comments and asking for explanations on the proxy resolution. In general looks good, and not a too big PR 👍

services/server/src/server/controllers/repository/repository.handlers.ts
manuelwedler: yes exactly. That's why I added it to the database querying method.

services/server/src/server/services/utils/proxy-contract-util.ts
manuelwedler: I will rename the variable to `proxyResolvers`. The returned object is needed for resolving the implementation address.

manuelwedler:
Added more comments. Hope they are comprehensive enough

manuelwedler:
I also updated whatsabi to the latest version, and added a check for factory contracts which is enabled by the latest version.

manuelwedler:
> Just, proxy resolution is added to `checkAllByChainAndAddress` but not `checkByChainAndAddress`. I'm ok with not adding it to the latter endpoint but just wanted to bring it up.

Unfortunately, the responds of the `checkByChainAndAddress` endpoint is structered differently and does not allow to add the proxy detection in a backwards-compatible way. I mentioned this in the PR description.

services/server/src/server/services/utils/proxy-contract-util.ts
```diff
@@ -0,0 +1,132 @@
...
+    (proxy) => proxy instanceof whatsabi.proxies.FixedProxyResolver,
+  );
+  // Only return EIP1167Proxies because whatsabi can falsely detect non-proxy contracts as FixedProxies (e.g. libraries)
+  if (fixedProxy && isEIP1167Proxy(bytecode, fixedProxy.resolvedAddress)) {
+    return {
+      isProxy: true,
+      proxyType: "EIP1167Proxy",
+      implementations: [fixedProxy.resolvedAddress],
+    };
+  }
```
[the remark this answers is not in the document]
manuelwedler: The whatsabi returns a `FixedProxy` for anything called with a `DELEGATECALL`. I think the idea is to also catch non-standardized "minimal" proxies, but the problem with this approach is that it can also include libraries as seen in this issue: https://github.com/shazow/whatsabi/issues/142

Therefore, I decided to only support the standard EIP 1167.

PR #1726, opened 2024-10-29
Sync back to staging

kuzdogan:


manuelwedler (approved):
Thanks for fixing the release. I think we also need to move back to the old process. I cannot think of a way how the new one would work out

kuzdogan:
IMO we should move to a more robust process potentially giving up the requirement of doing the tags on master commits.

Alhaidra (commented):
Good

PR #1727, opened 2024-10-29
Fix wrong tag name for sourcify-server in the npm publish script

kuzdogan:
Publishing the `sourcify-server` on npm failed on CI because of the wrong package name: https://app.circleci.com/pipelines/github/ethereum/sourcify/7227/workflows/fc503e37-673d-42a6-b69e-c318145e95b3/jobs/43733

Fixes this by giving the correct package name


manuelwedler (commented):
Can we maybe rename the server package to `@ethereum-sourcify/server`? I don't like the inconsistency so much here

manuelwedler (approved):
Probably renaming would break the versioning. Just merge this if you think renaming is too much effort

kuzdogan:
I would think about it thoroughly later, esp. with the incoming "organizational" changes we are having

manuelwedler:
Good point. Maybe we want to rename all packages then

PR #1728, opened 2024-10-29
Implement a storage service to upload contracts on S3

marcocastignoli:
This PR implements an S3-based storage service. The service allows uploading verified contract files to an S3 bucket while maintaining the same directory structure as RepositoryV2.

- Added new `S3RepositoryService` class that implements the storage interface
- Added S3 configuration options in .env.dev
- Integrated AWS SDK for S3 operations
- Added S3 storage service initialization in the server setup

`S3RepositoryService` is an extension of `RepositoryV2`, in which the write/delete/get file functions are overridden: instead of editing a fs, it's calling the AWS SDK APIs.

TODO:
- [x] Clean code
- [x] Add tests


marcocastignoli:
> Haven't looked yet but IMO we should name it to `S3Repository` instead of `RepositoryS3` it looks too similar to`RepositoryV3`

I already called it like this in the code but forgot to update the description, I had the same thought yesterday :D

services/server/src/config/local-test.js
```diff
@@ -15,7 +15,10 @@ module.exports = {
   },
   storage: {
     read: RWStorageIdentifiers.SourcifyDatabase,
-    writeOrWarn: [RWStorageIdentifiers.RepositoryV1],
+    writeOrWarn: [
+      RWStorageIdentifiers.RepositoryV1,
+      WStorageIdentifiers.S3Repository,
```
manuelwedler, line 20: Can we enable this also in staging already?

services/server/package.json
```diff
@@ -56,6 +56,7 @@
     "@ethereum-sourcify/lib-sourcify": "^1.10.0",
     "@google-cloud/cloud-sql-connector": "1.4.0",
     "abitype": "1.0.6",
+    "aws-sdk": "^2.1691.0",
```
manuelwedler, line 59: I think you should use `@aws-sdk/client-s3` which is the newest version of the package.

services/server/src/server/services/storageServices/S3RepositoryService.ts
```diff
@@ -0,0 +1,93 @@
...
+export class S3RepositoryService
+  extends RepositoryV2Service
+  implements WStorageService
+{
+  IDENTIFIER = WStorageIdentifiers.S3Repository;
+  private s3: S3Client;
+  private bucket: string;
+
+  constructor(options: S3RepositoryServiceOptions) {
+    super({ ipfsApi: "", repositoryPath: "" });
```
manuelwedler, line 31: `ipfsApi` is not used anymore. Would be nice to remove it also from the `RepositoryV2Service`.

services/server/src/server/services/storageServices/S3RepositoryService.ts
```diff
@@ -0,0 +1,93 @@
...
+      await this.s3.send(new DeleteObjectsCommand(deleteParams));
+    } catch (error) {
+      logger.error("Failed to delete partial match from S3", { error, prefix });
+      throw error;
+    }
+  }
+
+  async save(path: string | PathConfig, content: string) {
+    const filePath =
+      typeof path === "string" ? path : this.generateRelativeFilePath(path);
```
manuelwedler, line 78: I wondered why this function should be able to accept a string which could be a problem if `RepositoryV2Service` and `S3RepositoryService` would be called with the same string.

It seems the `save` method is only called with a `PathConfig` and never with a string. Can we remove the option to provide a string? `RepositoryV1Service` also wouldn't need it.

manuelwedler (changes requested):
Great solution to our problem with the s3 sync script :) 

I only have a few minor comments.

services/server/src/config/local-test.js
marcocastignoli: What do you mean? This config is just for tests, and they are passing

services/server/package.json
marcocastignoli: I'm already using it, but I probably forgot to install it. I'm wondering how can it work... But if you look at here, I'm actually using v3

https://github.com/ethereum/sourcify/blob/1676e81cd15b7ab5d8c5d0d26b86cec3e41835dc/services/server/src/server/services/storageServices/S3RepositoryService.ts#L5C1-L10C29

manuelwedler: Yeah saw that. I think it works because it is a dependency of the mock library.

PR #1744, opened 2024-11-20
Remove Kroma Etherscan

kuzdogan:
Removing the Kroma Etherscan instances as the tests have been failing and it seems the explorer will be sunset:

image


manuelwedler (changes requested):
Can you please also remove the test entries from `etherscanInstanceContracts.json`?

marcocastignoli (approved):

PR #1750, opened 2024-11-26
Remove Kroma etherscan test contracts

kuzdogan:
Following #1744 I should've removed the test contracts as well after removing Etherscan support for Kroma chain


manuelwedler (changes requested):
#1744 removed the tests for the chains in `services/server/test/chains/chain-tests.spec.ts`. We still need the tests as these chains are still supported.

kuzdogan:
Right 🤦 Can you check again @manuelwedler

manuelwedler (approved):
Thanks :)

PR #1759, opened 2024-11-27
[Vyper support] Add `VyperCheckedContract` in `lib-sourcify` and the new endpoint `/verify/vyper` in server

marcocastignoli:
This PR currently:
- Implements `VyperCheckedContract` and `SolidityCheckedContract`. `VyperCheckedContract` generates a metadata file for Vyper contracts.
- Implements `useVyperCompiler`, a function similar to `useCompiler` which handles dynamic download for Vyper versions (need some more work for the platform architectures), and vyper compilation.
- `verifyDeployed` now takes as parameter `CheckedContract`, an abstract class extended by both `VyperCheckedContract` and `SolidityCheckedContract`. It skips metadata-related functions for `VyperCheckedContract`.
- Stores match in all `StorageServices` (I still haven't checked everything in detail)


marcocastignoli:
Closed for #1760

PR #1776, opened 2024-12-10
Improve platform architecture management for Vyper compiler

manuelwedler:
I investigated which architectures the binaries are built for. I only had to add a check for Mac binary. I also improved the logging a bit because it was confusing when I tried to run a vyper verification.

### DIFF
--- a/packages/lib-sourcify/test/compiler/vyperCompiler.ts
+++ b/packages/lib-sourcify/test/compiler/vyperCompiler.ts
@@ -102,11 +102,13 @@ export async function getVyperExecutable(
     version,
     fileName,
   );
-  logDebug('Downloaded vyper', {
-    version,
-    vyperPath,
-    platform,
-  });
+  if (success) {
+    logDebug('Downloaded vyper', {
+      version,
+      vyperPath,
+      platform,
+    });
+  }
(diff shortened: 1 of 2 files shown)


packages/lib-sourcify/test/compiler/vyperCompiler.ts
```diff
@@ -10,7 +10,10 @@ import { logDebug, logError, logWarn } from '../../src/lib/logger';
...
 
 export function findVyperPlatform(): string | false {
-  if (process.platform === 'darwin') {
+  if (
+    process.platform === 'darwin' &&
+    (process.arch === 'x64' || process.arch === 'arm64')
+  ) {
     return 'darwin';
   }
   if (process.platform === 'linux' && process.arch === 'x64') {
```
marcocastignoli, line 19: Did you check anything for windows?

marcocastignoli (approved):

packages/lib-sourcify/test/compiler/vyperCompiler.ts
manuelwedler: Yes, I also checked Windows. I ran the `file` command to check for which architectures the three binary types are.

PR #1781, opened 2024-12-11
Add tests for Vyper support

manuelwedler:
adds some basic tests for the `VyperCheckedContract` and the new server endpoint.

It might be worth to add more tests for more complex Vyper contracts, or checking all data stored in the db.


marcocastignoli (approved):
lgtm!

PR #1783, opened 2024-12-11
Add support for Vyper cbor auxdata in bytecode-utils and in VyperCheckedContract

marcocastignoli:
This PR adds support for Vyper cbor auxdata, it is quite tricky to handle all the different versions of Vyper cbor auxdata so I had to rewrite most of the `bytecode-utils` package.

I also added the function `generateCborAuxdataPositions` in `VyperCheckedContract`, so `matchWithRuntimeBytecode` and `matchWithCreationBytecode` can successfully handle auxdata transformations adding the support for partial matches if Vyper integrity check doesn't match. 
That said, Vyper integrity check is still not supported as one of the cases to get a "perfect" match. I think we can wait for this, integrity check is in beta and we still have to evaluate it. So even if a Vyper contract has an integrity check match, it will lead to a "partial" match (in this case the auxdata is stored in the database).


services/monitor/src/ChainMonitor.ts
```diff
@@ -261,7 +264,7 @@ export default class ChainMonitor extends EventEmitter {
         return;
       }
       try {
-        const cborData = bytecodeDecode(bytecode);
+        const cborData = bytecodeDecode(bytecode, AuxdataStyle.SOLIDITY);
```
manuelwedler, line 267: Same as in `session-state.handler.ts`. Just add a comment that we decode it with Solidity style because it is the only published to ipfs.

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -32,92 +55,291 @@ export const decode = (bytecode: string): DecodedObject => {
...
-  const [, auxdata] = splitAuxdata(bytecode);
+  const [, auxdata] = splitAuxdata(bytecode, auxdataStyle);
+  // See more here: https://github.com/vyperlang/vyper/pull/3010
+  if (auxdataStyle === AuxdataStyle.VYPER) {
+    if (!auxdata) {
+      // Attempt fallback auxdata styles for older Vyper versions
+      const fallbackStyles = [
+        AuxdataStyle.VYPER_LT_0_3_10, // Vyper bytecode before version 0.3.10 uses SOLIDITY AuxdataStyle
+        AuxdataStyle.VYPER_LT_0_3_5, // Vyper bytecode before version 0.3.5 doesn't include auxdata length
+      ];
```
manuelwedler, line 66: I wonder what's the rationale of trying these fallbacks here? Wouldn't we call the `decode` function with the respective styles (`VYPER_LT_0_3_10`, ...) because we already know which version we want to verify for? 
I think I just need some explanation here.

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -32,92 +55,291 @@ export const decode = (bytecode: string): DecodedObject => {
   }
 
   // split auxdata
-  const [, auxdata] = splitAuxdata(bytecode);
+  const [, auxdata] = splitAuxdata(bytecode, auxdataStyle);
+  // See more here: https://github.com/vyperlang/vyper/pull/3010
+  if (auxdataStyle === AuxdataStyle.VYPER) {
```
manuelwedler, line 60: Also I wonder is it right that we only check for `AuxdataStyle.VYPER`? If we call this function with `VYPER_LT_0_3_10` or `VYPER_LT_0_3_5`, it would fail.

manuelwedler (changes requested):
Nice I like how well structured `splitAuxdata` is :)

Just some minor things and clarification needed

packages/lib-sourcify/src/lib/VyperCheckedContract.ts
```diff
@@ -181,4 +208,71 @@ export class VyperCheckedContract extends AbstractCheckedContract {
...
+              creationCborLengthHex ||
+                '0' /** handles vyper lower than 0.3.5 in which creationCborLengthHex is '' */,
+              16,
+            ),
+          value: `0x${auxdataFromRawCreationBytecode}`,
+        },
+      };
+      return true;
+    }
+    return true;
```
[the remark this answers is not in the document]
marcocastignoli: my idea here is to make it return false only if it cannot do it. But I don't want to return false for a contract that doesn't have auxadata. It successfully run but couldn't find anything.

packages/bytecode-utils/test/bytecode.spec.ts
```diff
@@ -42,39 +60,84 @@ describe('bytecode utils', function () {
...
+      });
+  });
+
+  it('bytecode decode Vyper cbor auxdata for version < 0.3.5', () => {
+    chai
+      .expect(decode(BYTECODE_VYPER_NO_AUXDATA_LENGTH, AuxdataStyle.VYPER))
+      .to.deep.equal({
+        compiler: '0.3.4',
+      });
   });
```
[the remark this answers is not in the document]
marcocastignoli: Because for Vyper version less than 0.3.5 there is no auxdata length, even if there is the auxdata itself

```
gte(0.3.5)
<bytecode><auxadata><two_bytes_auxdata_length>

lt(0.3.5)
<bytecode><auxadata>

packages/bytecode-utils/src/lib/bytecode.ts
marcocastignoli: My rationale here is that when you call `decode(` you don't know the compiler version, you just try to decode a bytecode and you just know that is solidity or vyper.

While when you call `splitAuxdata` you know also the version. 

I think this is useful for services like `Monitor` in which you just try to decode a bytecode without knowing the version in advance.

Does this make sense? If yes we should definitely explain this in a comment or change the names of the functions

packages/bytecode-utils/src/lib/bytecode.ts
marcocastignoli: `export const decode = <T extends AuxdataStyle.SOLIDITY | AuxdataStyle.VYPER>(`

`decode(` doesn't support all `AuxdataStyle` options

packages/lib-sourcify/src/lib/VyperCheckedContract.ts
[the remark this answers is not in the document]
manuelwedler: Ah that makes sense then. Maybe a comment on the function declaration would be helpful then.

packages/bytecode-utils/src/lib/bytecode.ts
manuelwedler: Ah I missed that

manuelwedler (approved):

PR #1774, opened 2024-12-09
Finalize vyper API: OpenAPI spec + error handling

marcocastignoli:
This PR finalizes Vyper API's OpenAPI specifications and improves errors in the Vyper API handler.

PR #1784, opened 2024-12-12
Add Vyper support to Sourcify

marcocastignoli:
Adds vyper support, here the initial PR: #1760. Below the related tasks:

- [x] Review what's stored in the database
- [x] ~~Create a common folder for the compilers code used by both lib-sourcify and server (idk if it fits in this new feature, imo yes because it's a lot of duplicated code now that we have also vyper stuff)~~ postponed
- [x] Rewrite platform architecture management (vyper versions for x86, arm, ...): #1776
- [x] Update READMEs #1786
- [x] Update documentation https://github.com/sourcifyeth/docs/pull/27
- [x] Handle vyper transformations: #1783
- [x] Add more tests, and finalize lib-sourcify test: #1781
- [x] Finalize `/verify/vyper` API errors + openapi responses: #1774
- [x] Check Vyper immutable field to see if we can support immutable transformation: #1790 

After merging:
- publish new docs


manuelwedler:
I think we should merge this by a merge commit in order to keep the individual commits

marcocastignoli:
@manuelwedler can you fix the conflicts here?

manuelwedler:
yes will do so now

manuelwedler (approved):

PR #1786, opened 2024-12-12
Add Vyper documentation in READMEs

marcocastignoli:
Please review also: https://github.com/sourcifyeth/docs/pull/27


packages/bytecode-utils/README.md
```diff
@@ -10,24 +10,59 @@ yarn add @ethereum-sourcify/bytecode-utils
...
+
+**Result example**
+
+```json
+{
+  "integrity": "...",
+  "runtimeSize": 1234,
+  "dataSizes": [32, 64, 128],
+  "immutableSize": 256,
+  "compiler": {
```
manuelwedler, line 55: I see in the test of your other PR that compiler looks like this: `compiler: '0.4.1'` (not an object). 

In the `VyperDecodedObject` it is defined like this:
```
compiler: {
    vyper: [number, number, number];
  };
  ```
  
  Not sure which of these three versions is right. This might also need a change in your other PR.

manuelwedler (approved):

PR #1809, opened 2024-12-24
Add binary search for creator tx hash

kuzdogan:
Adds a new function `findContractCreationTxByBinarySearch` to do a lower bound binary search for the transaction that created the contract.

Calls `eth_getCode` on the middle blocks to see if the contract has code. If yes, the contract should be created before this block. If no code, the contract should be created after this block.
Once the block is found, searches through all the tx's of the block to see if any of them created this contract.
Only supports EOA contract creations (tx.to === null). Tracing every single tx would've been quite expensive.

Adds unit tests and an integration test for the Mainnet Tether contract to test the function.

### DIFF
--- a/services/server/test/integration/verification-handlers/verify.stateless.spec.ts
+++ b/services/server/test/integration/verification-handlers/verify.stateless.spec.ts
@@ -29,7 +29,7 @@ import nock from "nock";
 
 chai.use(chaiHttp);
 
-describe("/", function () {
+describe.only("/", function () {
   const chainFixture = new LocalChainFixture();
   const serverFixture = new ServerFixture();
 
@@ -356,7 +356,7 @@ describe("/", function () {
           ];
         }
 
-        return [404]; // This response won't be used since the interceptor is removed
+        return nock.restore(); // Else let pass
       });
 
     let res = await chai
--- a/services/server/test/integration/verification-handlers/verify.stateless.spec.ts
+++ b/services/server/test/integration/verification-handlers/verify.stateless.spec.ts
@@ -338,7 +338,7 @@ describe("/", function () {
 
     // Intercept only eth_getTransactionReceipt calls but allow other RPC calls
     // This will block the binary search for the creation tx hash and creationMatch will be set to null
-    nock("http://localhost:8545")
+    const scope = nock("http://localhost:8545")
       .post("/")
       .reply(function (uri, requestBody) {
         const body =
@@ -355,8 +355,7 @@ describe("/", function () {
             },
           ];
         }
-        // Remove all interceptors and let this request pass through to the real endpoint
-        nock.cleanAll();
+
         return [404]; // This response won't be used since the interceptor is removed
       });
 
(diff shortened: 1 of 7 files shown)


manuelwedler:
Tests are failing

kuzdogan:
@manuelwedler  Fixed the failing tests after we enabled binary search for finding creation transactions. The test suite was breaking because the server now uses binary search to locate transaction hashes for contract creation.

I made two main changes to resolve this:
1. Added the necessary creation data to the test cases that needed it
2. Blocked the creation transaction fetching specifically for tests that verify database updates of creation info

Please have a look at the updated tests too. The tests are now passing again.

services/server/test/integration/verification-handlers/verify.stateless.spec.ts
```diff
@@ -335,13 +336,41 @@ describe("/", function () {
...
+            500,
+            {
+              error: {
+                message: "Blocked getTransactionReceipt",
+              },
+            },
+          ];
+        }
+        // Remove all interceptors and let this request pass through to the real endpoint
+        nock.cleanAll();
```
manuelwedler, line 359: I think it is not safe to call `nock.cleanAll` here. This will also remove the interceptors set up in `LocalChainFixture` and may interfere with other tests.

Maybe it would be better to mock the `SourcifyChain` function with sinon here?

manuelwedler (changes requested):
Looks good, clean implementation of the binary search.

Just one comment about one of the tests.

Also, I think we should evaluate how this feature affects our RPC calls in practice. If they get too high / costly it might make sense to add a max iteration parameter to the search.

services/server/test/integration/verification-handlers/verify.stateless.spec.ts
kuzdogan: Yeah good point. I first tried using stubs but it wouldn't work because we can't modify the the running server instance's SourcifyChain on the fly, right? Or I'm missing something.

manuelwedler: I haven't tried but I think it should be possible. You just need to access the instance the server is using and overwrite the related function, like `sourcifyChain.getTx = sandbox.stub()`. See `proxy-contract-util.spec.ts` for example. There we use a sandbox to restore after each test. 
For accessing the specific `SourcifyChain` instance, we might need to modify the `LocalChainFixture`.

kuzdogan: There you are passing the mocked SourcifyChain instances to the functions `detectAndResolveProxy`, they are unit tests. Current Server() setup does not allow "get"ing the chains and modifying them afaik. 

If you are ok with the final setup after the fix I think this is good enough.

manuelwedler: I'm fine with it. But there is still a `.only` modifier on the test 

Just for reference, shouldn't it be possible to just access the SourcifyChain by `serverFixture.server.chainRepository.sourcifyChainMap[chainId]`?

kuzdogan: Ooops good catch.

And turns out it was possible. Have a alook

manuelwedler: Looks good :)

manuelwedler (approved):

kuzdogan:
Also forgot to add, this is generally a least resort option. Many chains have alternative ways to find the creator tx hash so I don't think this will be used commonly. This is used more for the long tail of chains that don't have creator tx sources and provide public RPCs

Issue #1573, opened 2024-08-16
Missing creator transaction hashes after migration


manuelwedler:
We found that some contracts that had a `creator-tx-hash.txt` in the `/files/any/` endpoint shown in our old kubernetes deployment didn't have it after the migration for our new setup. The `transaction_hash` column in `contract_deployments` is simply NULL for these cases. 

I investigated two examples:

#### 0x00000000219ab540356cBB839Cbe05303d7705Fa on chain 1

![image](https://github.com/user-attachments/assets/a2220826-bfde-4e33-875b-c6b424d27f1e)

The logs show that there was a creatorTxHash found but fetching the creation bytecode failed. The latter probably happened due to an unresponsive RPC. 

We should still store the creatorTxHash in these cases. This is fixed by https://github.com/ethereum/sourcify/pull/1572

#### 0x4E7095a3519A33dF3D25774c2F9D7a89eB99745D on chain 11155111

![image](https://github.com/user-attachments/assets/86076b1e-11c6-4a48-94d5-dc48e10f65a0)

Here the log is a bit misleading. The fetching of the creatorTxHash failed actually because no `creatorTx` is in the jsonPayload. Therefore, lib-sourcify also didn't match with creation tx. I fixed the logging also in the above PR.

So in this case the cause was probably the Etherscan API being unresponsive.

### Database problem

What still remains is that we probably have a lot of contracts after the migration for which we are missing the creation tx hashes.   I queried the DB and it shows that there are 3.7M contract_deployments without tx_hash and 1.4M with the tx_hash. This seems a lot without tx hash to me. Unfortunately, we cannot know how it was before because we have never recovered the old production DB fully.

Two options two fix this problem now:

- We run the migration again and maybe add some checks for the transaction hashes to make sure they are fetched and stored.
- We go through the old repo, check the `creator-tx-hash.txt` files and add them manually to the database if not present yet.

A side note: There are a probably more problems after the migration. We should check the database more thoroughly. An unresponsive RPC could also have caused a change in the match status after the migration.

kuzdogan:
If we go with the second option we are going to have a dirty DB, that is contracts will have a creatorTxHash but no creation bytecodes. I think we should re-verify them.

We'll need a script that needs to connect to the both filesystems, the current one on GCP and the legacy one running on Argo. This script needs two lists:
1. The contracts that don't have a creationTxHash on the DB.
2. The contract that **have** a `creator-tx-hash.txt` in the legacy Argo repo

[...]

PR #1811, opened 2024-12-27
Creation tx backfill scripts

kuzdogan:
The script I've written to do the backfill of the #1573. It's basically a reverification script but it first queries the DB to find contracts that don't have the creation info and creates a new DB table to keep track of this. Then fetches these contracts with their deployment and compilation info in batches from the db and sends them to verification concurrently. Implements a cold start to avoid drowning the server and getting downtimes.

Start reviewing at `backfill-creation.mjs`
```
node backfill-creation.mjs --help
```

I also think we should refactor out the logger as it is being used in a lot of places, but something for the future.


kuzdogan:
Blockscout tests are failing because they've changed the API response from `creation_tx_hash` to `creation_transaction_hash`https://eth.blockscout.com//api/v2/addresses/0x00000000219ab540356cBB839Cbe05303d7705Fa 

We can ignore tests

services/database/scripts/ContractVerifier.mjs
```diff
@@ -0,0 +1,342 @@
...
+      ).toFixed(2),
+    });
+    await this.markContractAsVerified(
+      contract.chain_id,
+      Buffer.from(contract.address, "hex"),
+    );
+    this.verifiedCount++;
+  }
+
+  async handleAlreadyVerified(contract) {
```
manuelwedler, line 176: Are we actually updating the creation transaction hash in the db if we get this error?

manuelwedler (commented):
looks good to me. Just something about the comments and a question.

services/database/scripts/ContractVerifier.mjs
kuzdogan: No the scripts don't do manual edits to the DB, it has to be a full verification from scratch. 

The full workflow needs to be run multiple times. After running couple times until you get to reasonable leftover contracts, one needs to delete the table and create again (looking at the missing tx hashes in the DB). If a contract throwing this error was being verified, it will not show up again in the new table. Otherwise one needs to look at logs etc. why the contract is not receiving the creator tx hash.

manuelwedler: My questions was rather targeting the server. Does the server update the creation transaction hash in the database if the contract was already verified and the reverification doesn't lead to a better match?
If I read the `StorageService` code correctly, it throws before storing anything in that case.

kuzdogan: I think it implicitly does because it's actually a better match, from `creationMatch=null` to `creationMatch=partial/perfect`

Here 
https://github.com/ethereum/sourcify/blob/83cf76bda936530048401ef4787908ab0a4db961/services/server/src/server/services/utils/util.ts#L75-L99

and the check is used here:
https://github.com/ethereum/sourcify/blob/83cf76bda936530048401ef4787908ab0a4db961/services/server/src/server/services/StorageService.ts#L266-L292

kuzdogan: But I guess to your point, it does not save this info if it doesn't get a creation match. Question is, should it? Ideally yes but the amount of contracts is quite miniscule now. We only have around 600 contracts total left that can't be verified with a creation match

manuelwedler: Ah yes that's the point. It only doesn't if cannot get a creation match. I think it is fine then.

manuelwedler (approved):

PR #1834, opened 2025-01-07
Implement Vyper support in session etherscan API

marcocastignoli:
TODO
- [x] tests
- [x] minor refactoring


kuzdogan (approved):
LGTM! I think it's straightforward and not messy. We can merge if tests pass.

The only thing is, the JSON session would not work as is, because of this part:
https://github.com/ethereum/sourcify/blob/42cd05b48776047c21ca30aa6850cd95bcbac974/services/server/src/server/controllers/verification/solc-json/session/solc-json.session.handlers.ts#L36-L41

https://github.com/ethereum/sourcify/blob/42cd05b48776047c21ca30aa6850cd95bcbac974/services/server/src/server/services/VerificationService.ts#L142-L152

But IMO this is ok as we do this for showcasing purposes. I don't expect anyone to use the JSON session in the UI and the UI button says "Solidity JSON" anyway.

PR #1837, opened 2025-01-07
Don't run binary search test on forked PRs because of unavailable env vars

manuelwedler:
#1812 is failing because the binary search test on a live chain needs RPCs to function. In PRs from forked repos environment variables are not available.


marcocastignoli (approved):
👍

PR #1845, opened 2025-01-08
Verification flow refactoring

marcocastignoli:
This is the main pull request for the verification flow refactoring. Smaller feature-specific pull requests will be merged into this branch before it is finalized and merged into the staging environment.

See https://github.com/ethereum/sourcify/issues/1665 for more details


cshein45:
[MOVE](https://movementnetwork.com)

services/server/src/server/apiv1/verification/solc-json/stateless/solc-json.stateless.handlers.ts
```diff
@@ -38,59 +35,44 @@ export async function verifySolcJsonEndpoint(req: Request, res: Response) {
...
       compilerVersion,
     );
 
-  const checkedContracts = await checkFilesWithMetadata(
-    solc,
-    vyper,
+  // Create metadata contracts from the files
+  const metadataContracts = await createMetadataContractsFromFiles(
     metadataAndSourcesPathBuffers,
   );
```
kuzdogan, line 49: Why are we generating metadata files here? Can't we directly use the solcJson we have?

services/server/src/server/apiv1/verification/solc-json/stateless/solc-json.stateless.handlers.ts
```diff
@@ -38,59 +35,44 @@ export async function verifySolcJsonEndpoint(req: Request, res: Response) {
   const chain = req.body.chain;
   const address = req.body.address;
 
+  // Get metadata and sources from the Solidity JSON input
   const metadataAndSourcesPathBuffers =
     await services.verification.getAllMetadataAndSourcesFromSolcJson(
       solc,
       solcJson,
       compilerVersion,
     );
```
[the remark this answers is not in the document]
marcocastignoli: We don't know the compilationTarget

packages/lib-sourcify/test/validation.spec.ts
```diff
@@ -1,267 +0,0 @@
...
-    });
-  });
-});
-
-describe('Unit tests', function () {
-  const pathContent = {
-    path: './validation/files/hardhat-output/output.json',
-    content: JSON.stringify(hardhatOutput),
-  };
-  it('Should extractHardhatMetadataAndSources', async function () {
```
manuelwedler, line 219: This test was completely removed. I think we should not remvoe test cases for functions which we still have in the same form as before.

packages/lib-sourcify/test/functions.spec.ts
```diff
@@ -1,273 +0,0 @@
...
-    );
-    expect(contractWithPerfectMetadata).is.not.equal(null);
-    expect(
-      contractWithPerfectMetadata?.metadata?.sources['SimplyLog.sol']
-        ?.keccak256,
-    ).equals(
-      '0x8e7a1207ba791693fd76c6cf3e99908f53b8c67a5ae9f7b4ab628c74901711c9',
-    );
-  });
-  it('Should mock a metadata object for a Vyper contract', async () => {
```
manuelwedler, line 139: We also don't have a test for the metadata mock anymore

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -1,1318 +0,0 @@
...
-        childAddress,
-        {
-          msgSender: factoryAddress,
-        }
-      );
-
-      expectMatch(match, 'perfect', childAddress);
-    });
-    */
-    it('should fully verify a contract which is originally compiled and deployed with Unix style End Of Line (EOL) source code, but being verified with Windows style (CRLF) EOL source code', async () => {
```
manuelwedler, line 348: Is there a reason why we don't have this test anymore?

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -1,1318 +0,0 @@
...
-
-      const match = await checkAndVerifyDeployed(
-        contractFolderPath,
-        sourcifyChainHardhat,
-        contractAddress,
-      );
-      expectMatch(match, 'perfect', contractAddress);
-    });
-
-    it('should fully verify a contract when a not alphabetically sorted metadata is provided', async () => {
```
manuelwedler, line 367: same question here

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -1,1318 +0,0 @@
...
-
-      const match = await checkAndVerifyDeployed(
-        contractFolderPath,
-        sourcifyChainHardhat,
-        contractAddress,
-      );
-      expectMatch(match, 'perfect', contractAddress);
-    });
-
-    it('should fully verify a library with call protection when viaIR is enabled', async () => {
```
manuelwedler, line 417: and here

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -1,1318 +0,0 @@
...
-      const match = await checkAndVerifyDeployed(
-        contractFolderPath,
-        sourcifyChainHardhat,
-        contractAddress,
-      );
-      expectMatch(match, 'perfect', contractAddress);
-    });
-
-    // https://github.com/ethereum/sourcify/issues/1159
-    it('should compile nightly using Emscripten', async function () {
```
manuelwedler, line 437: here

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -1,1318 +0,0 @@
...
-          '1': '0x84188f8000a1657679706572830003090012',
-        },
-      });
-      expect(match.runtimeTransformations).to.be.deep.equal([]);
-      expect(match.runtimeTransformationValues).to.be.deep.equal({});
-    });
-  });
-
-  describe('Unit tests', function () {
-    describe('SourcifyChain', () => {
```
manuelwedler, line 784: all of these SourcifyChain tests were removed

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -1,1318 +0,0 @@
...
-      try {
-        expect(match.creationMatch).to.equal('perfect');
-        expect(match.address).to.equal(contractAddress);
-      } catch (e) {
-        console.log('Match: ', match);
-        throw e;
-      }
-    });
-
-    it('should find a perfect match for a contract that has a single metadata hash in the middle of the creation bytecode', async () => {
```
manuelwedler, line 1127: here

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -1,1318 +0,0 @@
...
-        creatorTxHash,
-        recompiledMetadata,
-        generateCreationCborAuxdataPositions,
-        recompiled.creationLinkReferences,
-      );
-      // Must check the creationMatch
-      expect(match.creationMatch).to.equal('perfect');
-    });
-
-    it('should find a perfect match for a contract that has multiple metadata hashes in the middle of the creation bytecode', async () => {
```
manuelwedler, line 1167: here

packages/lib-sourcify/test/verification.spec.ts
```diff
@@ -1,1318 +0,0 @@
...
-        contractAddress,
-        creatorTxHash,
-        recompiledMetadata,
-        generateCreationCborAuxdataPositions,
-        recompiled.creationLinkReferences,
-      );
-      expect(match.creationMatch).to.equal('perfect');
-    });
-
-    it('should not find a perfect match for a contract for which no cborAuxdataPositions could be generated', async () => {
```
manuelwedler, line 1206: here

manuelwedler (commented):
I wanted to understand if we removed some test case by accident, and checked if some tests from the old lib-sourcify version are missing. Here are the cases I found. Some might be removed on purpose. Please have a look.

packages/lib-sourcify/test/validation.spec.ts
marcocastignoli: Right! I completely lost that function since it's not used directly.

packages/lib-sourcify/test/verification.spec.ts
marcocastignoli: We replaced it with "should generate variations with different line endings", but I see now the test is not complete

packages/lib-sourcify/test/verification.spec.ts
marcocastignoli: I don't even know if this is needed anymore since now verification is not based on metadata

packages/lib-sourcify/test/verification.spec.ts
marcocastignoli: we have "should handle multiple auxdatas correctly"

packages/lib-sourcify/test/verification.spec.ts
marcocastignoli: we have "should handle multiple auxdatas correctly"

packages/lib-sourcify/test/verification.spec.ts
marcocastignoli: we have "should return partial match when there is perfect bytecode match but no auxdata"

packages/lib-sourcify/test/verification.spec.ts
manuelwedler: that should be fine then

packages/lib-sourcify/test/verification.spec.ts
manuelwedler: that's fine too

packages/lib-sourcify/test/verification.spec.ts
manuelwedler: I think this one is a bit different. Here we mock the `generateCreationCborAuxdataPositions` to fail. We don't have a test for this I think.

packages/lib-sourcify/test/verification.spec.ts
manuelwedler: How about a test that creates a SolidityMetadataContract with non-alphabetically sorted metadata and then tries to verify it?

packages/lib-sourcify/test/verification.spec.ts
marcocastignoli: The test you cited it's not the case it fails, it's the case in which it returns 

```
// This mocks the case where no cborAuxdataPositions could be generated
      const generateCreationCborAuxdataPositions = async () => {
        return {};
      };
      ```
      
We have that in "should handle case with no auxdata when metadata is disabled"

Comment (manuelwedler) [reply packages/lib-sourcify/test/verification.spec.ts:1206]: true. We could add one for the case that it throws.

Comment (marcocastignoli) [reply packages/lib-sourcify/test/verification.spec.ts:1206]: I cannot make this test work with `sinon`... Do you want to try?

File: packages/lib-sourcify/test/verification.spec.ts (hunk 6 above)
Comment (marcocastignoli) [reply packages/lib-sourcify/test/verification.spec.ts:367]: I don't think this is needed because when you import a json file in javascript via 

```
// Get the unsorted metadata
      const metadataPath = path.join(
        __dirname,
        '..',
        'sources',
        'StorageUnsortedMetadata',
        'metadata.json',
      );
      const metadataBuffer = fs.readFileSync(metadataPath);
      const unsortedMetadata = JSON.parse(metadataBuffer.toString());
      ```

it is automatically ordered.

manuelwedler: Right. This would only make sense if we use a string as input.

packages/lib-sourcify/test/verification.spec.ts
marcocastignoli: We decided to not implement it

manuelwedler (approved):

PR #1848, opened 2025-01-09
add core chain testnet2

LeeNorth:
# Add New Chain <chainId>

Thanks for your pull request to add a new support in Sourcify.

If you haven't done so, please follow the instructions on [how to request chain support](https://docs.sourcify.dev/docs/chain-support/) in docs.

Please check the following items before submitting your pull request for a speedy review.

## Checklist

- [x] I haven't modified the `chains.json` file directly.
- [x] In `sourcify-chains.json` file
- [x] `test-new-chain` test in Circle CI is passing.


leeyanbtcs:
@bertux

kuzdogan:
How do we get the API key? https://btcs.network/ does not load

kuzdogan:
I @LeeNorth I guess you should not share keys here 😄 we'd like to get our own one using our own email, how can we do that?

LeeNorth:
@kuzdogan Okay, you can register and log in here: https://scan.test2.btcs.network/, then obtain the key from the API-KEYS section

kuzdogan:
Now the Etherscan tests are failing because it can't verify the contract. Are you sure the contract deployed at addresses `0x06Fd8b725adBa12A76B21F0a03AccbF30d64DC55` and `0xa4807B287191a9044Cc3CEe537641E9B829cF2F1` are really those at https://scan.test2.btcs.network/address/0x06Fd8b725adBa12A76B21F0a03AccbF30d64DC55#code and https://scan.test2.btcs.network/address/0x06Fd8b725adBa12A76B21F0a03AccbF30d64DC55#code ? How are these contracts deployed? 

The bytecodes are quite different. 

For 0x06Fd8b725adBa12A76B21F0a03AccbF30d64DC55
image

The onchain bytecode we get from the RPC matches with the one in the block explorer, but when compiled we get a totally differnt one. My only explanation would be if the verification on the block explorer is not correct. Maybe it's something on our side but I can't think of anything.

LeeNorth:
@kuzdogan The contract was deployed using Remix. Here are the deployment parameters. 
image

kuzdogan:
Yeah I confirm it matches when `evmVersion` is set to `paris` but the API is returning `evmVersion: 'Default'`

```json
{
  "status": "1",
  "result": [
    {
      "SourceCode": "// SPDX-License-Identifier: GPL-3.0\n\npragma solidity >=0.7.0 <0.9.0;\n\n/**\n * @title Storage\n * @dev Store & retrieve value in a variable\n */\ncontract Storage {\n\n    uint256 number;\n\n    /**\n     * @dev Store value in variable\n     * @param num value to store\n     */\n    function store(uint256 num) public {\n        number = num;\n    }\n\n    /**\n     * @dev Return value \n     * @return value of 'number'\n     */\n    function retrieve() public view returns (uint256){\n        return number;\n    }\n}\n",
      "ABI": "[{\"outputs\":[{\"name\":\"\",\"internalType\":\"uint256\",\"type\":\"uint256\"}],\"inputs\":[],\"name\":\"retrieve\",\"stateMutability\":\"view\",\"type\":\"function\"},{\"outputs\":[],\"inputs\":[{\"name\":\"num\",\"internalType\":\"uint256\",\"type\":\"uint256\"}],\"name\":\"store\",\"stateMutability\":\"nonpayable\",\"type\":\"function\"}]",
      "ContractName": "Storage",
      "CompilerVersion": "v0.8.26+commit.8a97fa7a",
      "OptimizationUsed": "0",
      "Runs": "200",
      "ConstructorArguments": null,
      "EVMVersion": "Default",
      "Library": null,
      "LicenseType": "GNU GPLv3",
      "Proxy": "0",
      "Implementation": null,
      "SwarmSource": null
    }
  ],
  "message": "OK"
}
```

kuzdogan:
> @kuzdogan Yes, the scan API has not been updated, so the evmVersion is returning as 'default'. However, the EVM on our chain currently only supports 'Pair'. Does this matter?

That means we can't verify a portion of contracts from the block explorer, but for the tests to pass, can you please change the Ethercan test contracts to ones compiled with 0.8.18 or 0.8.19 (default: paris)?

LeeNorth:
@kuzdogan We have updated the API, and it now returns the correct EVM version. Could you please try again?

kuzdogan:
Ok yes it works. The only little things is, the match type of `0xa4807B287191a9044Cc3CEe537641E9B829cF2F1` needs to be `partial` instead of `perfect`. I don't have edit access, can you please change that?

LeeNorth:
@kuzdogan updated.

kuzdogan (approved):

PR #1850, opened 2025-01-09
Rename controllers folder to apiv1

manuelwedler:
For API v2, we wanted to rename the controllers folder. This PR renames it to `apiv1`. API v2 can be developed in a folder `apiv2` on the same level. 

I think it is better to make this change now directly to staging to avoid conflicts later.


marcocastignoli:
Tests are not passing

manuelwedler:
Oh wait, it seems Mocha skips the tests in the subfolders now

marcocastignoli (approved):

marcocastignoli:
> Oh wait, it seems Mocha skips the tests in the subfolders now

Too late, I simply trusted tests

PR #1851, opened 2025-01-09
Init class

### LINKED ISSUE #1821: Implement new "Validation step" classes and their unit tests
Implement classes to validate a set of sources / setting before passing them to the compiler. For example in the case of Solidity, the class will use information in `metadata.json` to validate existing sources and check for missing sources. The Vyper validation class could run the integrity check on the existing files.

In general this classes can return an instance of the ️`Compilation` (#1822) objects.

kuzdogan:
Fixes #1821 

PR to be merged to the main Verification Refactoring flow. I'll be documenting the changes and rationale along the way.

- Current `validation.ts` and the way we create `CheckedContract`s from `files` is by first splitting the files as metadata and sources. Then we take variations of all source files and from them we find the correct sources for each metadata. I decided to move the variation step into `SolidityMetadataContract` class. The downside is we might need to run "variation" and keccaking steps multiple times for each source file, when there are multiple metadata and a set of source files in the current files array. However, our API will accept a single metadata. So I don't expect this to happen in practice a lot. Even if it happens it's should be still ok, even if hashing could be a little expensive. This way it's cleaner conceptually, as all the mapping, variation etc. is within the class.
- Note: We have `JsonInput` type for Solidity but `VyperJsonInput` type for Vyper


kuzdogan:
Requesting an interim review before adding tests @marcocastignoli

packages/lib-sourcify/package.json
```diff
@@ -18,7 +18,7 @@
     "check": "run-s check:*",
     "check:eslint": "eslint . --ext .ts",
     "check:prettier": "prettier \"./**/*.ts\" --check",
-    "test": "c8 --reporter=none mocha -r ts-node/register test/**/*.spec.ts --no-timeout --exit",
+    "test": "c8 --reporter=none mocha -r ts-node/register test/**/*.spec.ts test/*.spec.ts --no-timeout --exit",
```
marcocastignoli, line 21: Why do you need this? shouldn't `test/*.spec.ts` be already included in the superset `test/**/*.spec.ts`?

packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
```diff
@@ -0,0 +1,383 @@
...
+  metadata: Metadata;
+  name: string;
+  path: string;
+  providedSources: PathContent[];
+  providedSourcesByHash: Map<string, PathContent>;
+  foundSources: StringMap;
+  missingSources: MissingSources;
+  invalidSources: InvalidSources;
+  unusedSourceFiles: string[];
+  metadata2provided: StringMap; // maps the file path as in metadata.sources to the path of the provided by the user. E.g. metadata can have "contracts/1_Storage.sol" but the user provided "/Users/user/project/contracts/1_Storage.sol"
```
marcocastignoli, line 60: I would rename this property, at least removing the 2: I like `metadataToProvidedPathMap`

packages/lib-sourcify/src/Validation/fetchUtils.ts
```diff
@@ -0,0 +1,136 @@
+import { logError, logInfo, logDebug } from '../lib/logger';
+import { id as keccak256str } from 'ethers';
+import { IpfsGateway } from '../lib/types';
+
+export async function performFetch(
```
marcocastignoli, line 5: Is this only related with Validation? If that's the case, it's ok to keep this here, if it is also used by Verification or other utils let's move it outside from /Validation

packages/lib-sourcify/src/Validation/processFiles.ts
```diff
@@ -0,0 +1,243 @@
...
+/**
+ * Regular expression matching metadata nested within another string.
+ * Assumes metadata's first key is "compiler" and the last key is "version".
+ * This is true for the metadata files generated by the compiler as it's canonicalized by sorting the keys.
+ */
+const NESTED_METADATA_REGEX =
+  /"{\\"compiler\\":{\\"version\\".*?},\\"version\\":1}"/;
+const HARDHAT_OUTPUT_FORMAT_REGEX = /"hh-sol-build-info-1"/;
+
+export function createMetadataContractsFromPaths(
```
marcocastignoli, line 18: Regarding this, I reiterate what I said yesterday, I would like the final library to be used like this:

```ts
import {Validation, Compilation, Verification} from "@ethereum-sourcify/lib-sourcify";

const metadataContracts = Validation.createMetadataContractsFromPaths(...)

const metadataContract = new Validation.SolidityMetadataContract(...)
```

What do you think? If you like it, then we can rename this file as `validation.ts` (like an index of this folder), in which we export everything related to validation:
- the functions already in this file
- and we import/export `SolidityMetadataContract`

packages/lib-sourcify/src/Validation/zipUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+      // Remove the zip file from the array and decrement the index to check the next file.
+      files.splice(i, 1);
+      i--;
+    }
+  }
+  // Add unzipped at the end to not check again if the extracted files are zips.
+  files.push(...allUnzipped);
+}
+
+function isZip(file: Buffer): boolean {
```
marcocastignoli, line 32: Same comment as `fetchUtils.ts`. I don't like them in the same folder as `SolidityMetadataContract.ts` and `processFiles.ts`. What about creating a `utils/` folder in which we put both this and `fetchUtils.ts`?

marcocastignoli (changes requested):
Just a few comments in particular for renaming and moving staff around. 

After solving the comments I think we can proceed by:
- merging this back in the main PR
- create a new PR in which we adjust `Validation` to use the new types from `Compilation`
- create a new PR in which we create `Validation` tests

packages/lib-sourcify/package.json
kuzdogan: Yes, but for some reason it didn't do it for me 🤷

packages/lib-sourcify/src/Validation/fetchUtils.ts
kuzdogan: I don't see this being used anywhere else, except fetching sources, which is part of the validation. If needed we can move it out later.

packages/lib-sourcify/src/Validation/zipUtils.ts
kuzdogan: As said, I don't see it being used anywhere else, so it's actually only validation related utils. If that's shared with some other parts of the code, then yes. For me this structure is quite clear.

packages/lib-sourcify/src/Validation/processFiles.ts
kuzdogan: Ok let's go with your way, it's nice. How do we do that do we declare a `Validation` class or a namespace, or function?

packages/lib-sourcify/package.json
marcocastignoli: Yes

packages/lib-sourcify/src/Validation/processFiles.ts
marcocastignoli: I think we can just export a default object with all the needed classes / functions in an index file

```ts
// index.ts
export default {
  Validation:{},
  Compilation:{},
  Verification:{},
}
```

marcocastignoli (approved):

PR #1852, opened 2025-01-09
Add APIv2 OpenAPI specification

manuelwedler:
`apiv2.yaml` is directly retrieved from Stoplight and can just be copy-pasted again when an update is needed in the future. Only if we add more endpoints we need to manually reference them in `openapi.yaml` due to limitations of the OpenAPI spec.

Needed to change the mode of the validator's refparser because the we reference the same file multiple times now.

~~I'm not sure about the tags for the swagger documentation yet. Maybe we should add a "deprecated" to APIv1 instead of the "API v1" prefix?~~

### DIFF
--- a/services/server/src/apiv2.yaml
+++ b/services/server/src/apiv2.yaml
@@ -1,0 +1,1314 @@
+openapi: 3.1.0
+info:
+  version: '0.2'
+  title: Sourcify v2 - Draft
+  summary: DRAFT - Sourcify v2
+  description: |-
+    Welcome to the Sourcify's APIv2 Draft. Our new API is currently in the design phase and we are seeking feedback!
+
+    Important differences between the [current API](https://docs.sourcify.dev/docs/api/):
+    - **Ticketing**: The verfication requests resolve into tickets/verification jobs. 
+      - Previously the verification happened during the HTTP request, which resulted in timeouts if compilation took longer
+    - **Standard JSON as default**: In the current design we take the standard JSON format as our main verification endpoint. We still support verification with metadata at `/verify/metadata`.
+    - **Lean API**: We keep the number of endpoints minimal compared to v1. We won't have a session API. 
+    - **Detailed contract response**: Prev. we only returned contract files of a contract. Now we can return details at `/contract/{chainId}/{address}`.
+  contact:
+    name: Sourcify
+    url: 'https://sourcify.dev'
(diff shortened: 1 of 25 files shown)


manuelwedler:
> I'm not sure about the tags for the swagger documentation yet. Maybe we should add a "deprecated" to APIv1 instead of the "API v1" prefix?

I think I will also mark API v1 endpoints as deprecated and change the tags

manuelwedler:
Ready for review again

kuzdogan (changes requested):
I think we should move (Deprecated) part to the beginning of titles. I went forward and done those myself, I'm guessing that's ok.

Otherwise I can't run it as is. I tried clearing the build cache etc. but I get the following output, (quite long so truncated)

```
  },
  {
    "instancePath": "/paths/~1v2~1contracts~1{chainId}/get/responses/500/content/application~1json/schema",
    "schemaPath": "#/properties/schema/oneOf",
    "keyword": "oneOf",
    "params": {
      "passingSchemas": null
    },
    "message": "must match exactly one schema in oneOf"
  },
  {
    "instancePath": "/paths/~1v2~1contracts~1{chainId}/get/responses/500",
    "schemaPath": "#/definitions/Reference/required",
    "keyword": "required",
    "params": {
      "missingProperty": "$ref"
    },
    "message": "must have required property '$ref'"
  },
  {
    "instancePath": "/paths/~1v2~1contracts~1{chainId}/get/responses/500",
    "schemaPath": "#/patternProperties/%5E%5B1-5%5D(%3F%3A%5Cd%7B2%7D%7CXX)%24/oneOf",
    "keyword": "oneOf",
    "params": {
      "passingSchemas": null
    },
    "message": "must match exactly one schema in oneOf"
  }
]
```

Also when I run the server and try to access any endpoint I get this:
image

manuelwedler:
Hm I cannot reproduce this. Just tried again with a clean install and build, and the server responds correctly as expected. Also the CI works. I can try again on Monday.

kuzdogan:
Let's also see what @marcocastignoli has, maybe a Linux/MacOs thing

kuzdogan (approved):
Ok clean install solved it, thanks @marcocastignoli 

LGTM. I'd just note I prefer Stoplight UI or another UI than Swagger http://localhost:5555/api-docs/ But that's something for later.

Issue #1827, opened 2025-01-06
Implement the new APIv2 lookup endpoints


marcocastignoli:
Implement:
- [Get verified contract](https://sourcify.stoplight.io/docs/sourcify-apiv2/branches/main/cqx62aqefyrje-get-verified-contract)
- [List of verified contracts per chain](https://sourcify.stoplight.io/docs/sourcify-apiv2/branches/main/kqms8dret0zja-list-of-verified-contracts-per-chain)

This task in not blocked by the rest of APIv2 tasks defined in the #1367 milestone

manuelwedler:
done in #1874 and #1857

Issue #1855, opened 2025-01-15
Implement GenericErrorResponse interface for all errors returned from /v2


manuelwedler:
In the API v2 design we created the GenericErrorResponse interface for all errors returned by the new API:

```ts
export interface GenericErrorResponse {
  customCode: ErrorCode;
  message: string;
  errorId: string;
}
```

We should make sure that all errors that are returned from the `/v2` endpoints are following this format. For that, we can implement some middleware, e.g. `use("/v2", errorHandlingMiddleware)` that matches unknown / uncaught errors to the new interface. This should also catch the ones returned by `safeHandler`. We should also ensure that validation errors follow this format. We might need to adapt our `genericErrorHandler` from api v1.

We will also need to define the custom error codes.

Also check https://github.com/ethereum/sourcify/issues/1829#issuecomment-2621689381 while working on this

PR #1857, opened 2025-01-16
Add `v2/contracts/{chainId}` API endpoint

manuelwedler:
See #1827

Adds the `v2/contracts/{chainId}` endpoint as described in https://sourcify.stoplight.io/docs/sourcify-apiv2/kqms8dret0zja-list-of-verified-contracts-per-chain

Includes the initial setup of routes and types needed for API v2, as it's the first endpoint. Also refactors some API v1 code to the apiv1 folder.

### Pagination issue
I discovered an issue about the pagination design of the API. The issue is that we include the `totalResults` in the response. To gather this information, we run a very long db query for counting the matches. It took 60-90 seconds in my trials on production. This issue is already present in API v1. 

I think we can still review this PR, and resolve this issue in another when we decided how to go about it.


manuelwedler:
We decided to implement id based pagination to circumvent the above issue. 

We need to add the `id` from `sourcify_matches` to our VerifiedContractMinimal object from the API design.

The API query should look like this then: `/v2/contracts/{chainId}?limit=20&afterId=5500608&sort=desc`

@marcocastignoli you can wait with reviewing until I implemented this.

kuzdogan:
Just to make sure, should it be from `sourcify_matches` or `verified_contracts`?

manuelwedler:
It should be `sourcify_matches.id`. `verified_contracts.id` wouldn't be unique to the chainId and address, right?

services/server/src/server/apiv2/errors.ts
```diff
@@ -0,0 +1,23 @@
...
+  errorId: string;
+}
+
+export class ChainNotFoundError extends NotFoundError {
+  payload: GenericErrorResponse;
+
+  constructor(message: string) {
+    super(message);
+    this.payload = {
+      customCode: "unsupported_chain",
```
marcocastignoli, line 18: I would specify a list of all the custom codes that we have. Not a complete one, but let's initialize it

services/server/src/server/apiv2/errors.ts
```diff
@@ -0,0 +1,23 @@
+import { NotFoundError } from "../../common/errors";
+import { v4 as uuidv4 } from "uuid";
+
+export type ErrorCode = "unsupported_chain";
```
marcocastignoli, line 4: this is wrong right?

marcocastignoli (changes requested):
I'll review again after you implement the new pagination. Great job! 👍

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -228,20 +244,59 @@ export class SourcifyDatabaseService
...
+      parseInt(chainId),
+      page,
+      limit,
+      descending,
+    );
+
+    const results: VerifiedContractMinimal[] = sourcifyMatchesResult.rows.map(
+      (row) => ({
+        match:
+          row.runtime_match === "perfect" || row.creation_match === "perfect"
```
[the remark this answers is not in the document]
manuelwedler: Yes, I can put this in an util function.

But I don't think we should start pushing `exact_match` into the database. For now, it should just be an API level label. If we also want it in the database, we should create a migration script to also rename the old data. Otherwise the database is not consistent.

services/server/src/server/apiv2/errors.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class ChainNotFoundError extends NotFoundError {
+  payload: GenericErrorResponse;
+
+  constructor(message: string) {
+    super(message);
+    this.payload = {
+      customCode: "unsupported_chain",
+      message,
+      errorId: uuidv4(),
```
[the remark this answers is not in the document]
manuelwedler: I guess Kaan proposed it to be a UUID and I think it makes a lot of sense :) Users can give use their error message and we can find exactly their request in the logs.

services/server/src/server/apiv2/errors.ts
manuelwedler: I already created an issue for that: https://github.com/ethereum/sourcify/issues/1855
I would rather keep it in the scope of another PR

services/server/package.json
```diff
@@ -16,9 +16,9 @@
     "postgres-test:start": "docker-compose -f test/docker-compose.yml up -d",
     "postgres-test:stop": "docker-compose -f test/docker-compose.yml rm -fsv",
     "postgres-test:migrate": "cd ../database && npm run migrate:up -- --env test",
-    "test:unit": "mocha  --exit test/unit/**/*.spec.ts",
-    "test:integration": "npm run postgres-test:migrate && mocha --exit test/integration/**/*.spec.ts",
-    "test": "npm run postgres-test:migrate & npx c8 --reporter=none mocha --exit test/unit/**/*.spec.ts test/integration/**/*.spec.ts",
+    "test:unit": "mocha  --exit --recursive test/unit/**",
```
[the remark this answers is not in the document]
manuelwedler: See the commit message: "Some shells require mocha's `recursive` option to go through all subdirectories"

I think the problem was introduced by moving the apiv1 tests in another subfolder.

Interestingly, the problem wasn't there with vscode/cursor integrated run option.

services/server/src/server/services/utils/util.ts
```diff
@@ -96,3 +96,14 @@ export function isBetterMatch(newMatch: Match, existingMatch: Match): boolean {
...
   return false;
 }
+
+export function toV2MatchLevel(status: Status): V2MatchLevel {
+  switch (status) {
+    case "perfect":
+      return "exact_match";
+    case "partial":
+      return "match";
+    default:
```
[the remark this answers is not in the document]
manuelwedler: I wouldn't do it soon. People might also rely on the old labels when they use the parquet export.

What's your opinion on this @kuzdogan ?

services/server/src/server/apiv2/errors.ts
manuelwedler: How do you mean? This is just the first error code that I implemented. There should follow a lot more.

manuelwedler: So this type should be extended then

services/server/src/server/apiv2/errors.ts
manuelwedler: To clarify, the list should go into the `ErrorCode` type

marcocastignoli (approved):
LGTM!

Issue #1835, opened 2025-01-07
Investigate failing Etherscan API tests


manuelwedler:
Our etherscan instance tests are failing since last Friday. I think they might have changed something about their API. They sent an e-mail that you should move your instances API keys to main Etherscan.

manuelwedler:
Etherscan seems to have changed something about their instances. Some instances seem to not have required an API key, and for some of these we didn't set up a key in the CI or didn't configure them properly in the `sourcify-chains-default.json`.

Working on a fix.

manuelwedler:
After my recent fixes, there are still 5 tests failing. I would wait a day, see which fail consistently and act then.

manuelwedler:
Added another missing api key to CI, and now only two tests are failing. Chains 252 and 2522. The issue is that the test expects the verification to be partial but it is actually perfect. The weird thing is that it's partial as expected when I run the test locally.

manuelwedler:
It seems the issue of the contracts on chain 252 and 2522 is a bug that #1839 introduced. lib-sourcify cannot generate the cborauxdata positions for these contracts, but due to the logic from #1839 the creation code matches falsely to perfect.

PR #1860, opened 2025-01-17
Add missing Etherscan api key env names

manuelwedler:
See #1835 

Etherscan seems to have changed something about their instances. Some instances seem to not have required an API key, and for some of these we didn't set up a key in the CI or didn't configure them properly in the `sourcify-chains-default.json`.

This PR fixes the latter. I already added the missing env variables to the CI.


marcocastignoli (approved):
Did you add the new keys both in prod and staging on GCP?

manuelwedler:
Yes, I went on both production and staging, and checked all keys. I found even more missing, so I added all that were missing.

PR #1861, opened 2025-01-20
Implement functionalities to allow sync in batch

marcocastignoli:
This PR implements a new `SourcifyDuneSyncClient` that generalizes batch synchronization for all the sourcify table to Dune.

### DIFF
--- a/services/database/dune/fetchRows.ts
+++ b/services/database/dune/fetchRows.ts
@@ -59,7 +59,6 @@ async function fetchWithQuery(table: string, customQuery?: string) {
       return null;
     }
 
-    console.log(`Found ${result.rows.length} ${table}`);
     return result.rows;
   } catch (error) {
     console.error(`Error fetching ${table}:`, error);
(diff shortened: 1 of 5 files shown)


services/database/dune/fetchRows.ts
```diff
@@ -96,73 +102,85 @@ export async function fetchCompiledContracts(): Promise<
...
-  PgSourcifyMatch[] | null
-> {
-  return fetchWithQuery("sourcify_matches");
+export async function fetchSourcifyMatches(
+  page: number,
+  pageSize: number,
+): Promise<PgSourcifyMatch[] | null> {
+  return fetchWithQuery(
+    "sourcify_matches",
+    `SELECT * FROM sourcify_matches ORDER BY created_at ASC OFFSET ${page} LIMIT ${pageSize}`,
```
manuelwedler, line 158: I wonder why you don't paginate by using `where id > ${last_id} limit ${pageSize}` 
Using offset is generally slow with high offsets because the database has to go over all rows smaller than the offset. I think it would be better to use the `id` based approach also here.

Regarding ordering, also the queries can be very slow because `created_at` is not indexed. I think it would be better to use `order by id` too.

This counts for all other queries too. Using indexed columns for iterating makes everything much more performent.

manuelwedler (changes requested):
I think the efficiency of the script can be improved

services/database/dune/fetchRows.ts
marcocastignoli: The problem is that not all tables have an incremental id

manuelwedler: Does that matter? you can even do it on the code table where we have hashes as id.

Just tried this successfully for example:
```
where code_hash > decode('52E56732A80EA66B14A846E383A99CB3455871F2C899E23CE81F02EAA5F94C18', 'hex') order by code_hash limit 10;
```

manuelwedler (approved):
Looks good!

PR #1867, opened 2025-01-23
Fix matching creation bytecode when no cborAuxdataPositions could be generated

### LINKED ISSUE #1835: Investigate failing Etherscan API tests

manuelwedler:
Fixes #1835

Fixes a bug #1839 introduced.

In case `cborAuxdataPositions = {}`, `Object.values(cborAuxdataPositions).every(...)` evaluates to true and therefore matches the creation match to perfect.


marcocastignoli (approved):
LGTM

Issue #1823, opened 2025-01-06
Implement new "Verification step" classes,and  their unit tests


marcocastignoli:
Implement the `Verification` class. A class that takes as input the newly defined `Compilation`(#1822) class + onchain information (address,chainId) and returns as output the verification status + transformations + `Compilation` properties .

The properties of this class are going to be used by `AbstractDatabaseService` to fill the database tables.

This task is blocked by #1822, but if needed we can temporarily implement this task by taking `CheckedContract` as input instead of `Compilation` (only if we have time and `Compilation` takes a long time to develop).

PR #1879, opened 2025-01-30
Verification flow verification classes

marcocastignoli:
This PR implements the new Verification class defined in #1823.

TODO:
- [x] ~~Implement calling `matchWithRuntimeBytecode` and `matchWithCreationTx` in parallel~~
- [x] Fix `should verify a contract with viaIR:true, optimizer disabled, and compiler <0.8.21`
- [x] Implement Vyper constructor argument transformation test
- [x] Implement custom SourcifyLibError, with custom code/message
- [x] Include all transformations functions inside the `Verification` class
- [x] Move transformations into `Verification/Transformations.ts`
- [x] Do not include any dependency from the old `./src/lib` folder


packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,691 @@
...
+    // Here we use bytecodes from the context because they are already processed
+    let normalizedRecompiledBytecode = context.normalizedRecompiledBytecode;
+    const onchainBytecode = context.isCreation
+      ? this.onchainCreationBytecode
+      : this.onchainRuntimeBytecode;
+
+    const cborAuxdata = context.isCreation
+      ? this.compilation.creationBytecodeCborAuxdata
+      : this.compilation.runtimeBytecodeCborAuxdata;
+
```
manuelwedler, line 245: One optional thing that you could do if you like it. If we pass `onchainBytecode`, `cborAuxdata` and `linkReferences` in the context (transformations should not be needed anymore when you address my other comment), we can also make this function pure. Then, I would also add a `matchingFunction` to the context like this: 

- In matchWithRuntimeBytecode: `matchingFunction: (onchainBytecode, normalizedRecompiledBytecode) => onchainBytecode === normalizedRecompiledBytecode`
- In matchWithCreationTx: `matchingFunction: (onchainBytecode, normalizedRecompiledBytecode) => onchainBytecode.startsWith(normalizedRecompiledBytecode)`

manuelwedler (changes requested):
I only had a glimpse on the tests. Not sure if I can take a closer look before my vacation.

packages/lib-sourcify/src/Verification/Verification.ts
marcocastignoli: I prefer having matchBytecodes inside `Verification`, from my pov it's a grouping of multiple transformations functions otherwised called directly in `matchWithRuntimeBytecode` and `matchWithCreationTx`.

packages/lib-sourcify/src/Verification/VerificationTypes.ts
```diff
@@ -0,0 +1,36 @@
...
+export interface BytecodeMatchingResult {
+  match: 'perfect' | 'partial' | null;
+  libraryMap?: StringMap;
+  normalizedRecompiledBytecode: string;
+  transformations: Transformation[];
+  transformationValues: TransformationValues;
+  message?: string;
+}
+
+export enum VerificationErrorCode {
```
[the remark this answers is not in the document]
manuelwedler: This one is not tackled yet. What's your opinion on this?

manuelwedler (approved):
Good job!

My comments were addressed successfully. I still didn't have a close look on the tests, but from what I saw they looked good. If you want to have a look on the tests next week @kuzdogan , go ahead. Otherwise, this is good to go from my point of view.

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,500 @@
...
+    /* eslint-disable indent */
+    const matchesNormalizedBytecode = context.isCreation
+      ? onchainBytecode.startsWith(
+          auxdataTransformationResult.normalizedRecompiledBytecode,
+        )
+      : auxdataTransformationResult.normalizedRecompiledBytecode ===
+        onchainBytecode;
+    /* eslint-enable indent */
+
+    if (matchesNormalizedBytecode) {
```
kuzdogan, line 351:
```suggestion
    const doNormalizedBytecodesMatch = context.isCreation
      ? onchainBytecode.startsWith(
          auxdataTransformationResult.normalizedRecompiledBytecode,
        )
      : auxdataTransformationResult.normalizedRecompiledBytecode ===
        onchainBytecode;
    /* eslint-enable indent */

    if (doNormalizedBytecodesMatch) {
```

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,500 @@
...
+  get libraryMap() {
+    return {
+      runtime: this.runtimeLibraryMap,
+      creation: this.creationLibraryMap,
+    };
+  }
+
+  get abiEncodedConstructorArguments() {
+    return this._abiEncodedConstructorArguments;
+  }
```
kuzdogan, line 499: Do we assign this anywhere at all?

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,500 @@
...
+
+  async handleSolidityBugCases(
+    forceEmscripten: boolean,
+    compiledRuntimeBytecode: string,
+  ): Promise<SolidityBugType> {
+    // Handle Solidity specific verification bug cases
+    if (
+      this.compilation instanceof SolidityCompilation &&
+      this.runtimeMatch === null
+    ) {
```
kuzdogan, line 183: IMO this should be outside the function. It was weird to read handleSolidityBugCases in the verify function wihtout these checks

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,500 @@
...
+          compiledRuntimeBytecode,
+          AuxdataStyle.SOLIDITY,
+        );
+        // Metadata hashes match but bytecodes don't match.
+        if (
+          deployedAuxdata === recompiledAuxdata &&
+          settings.optimizer?.enabled
+        ) {
+          throw new VerificationError(
+            "It seems your contract's metadata hashes match but not the bytecodes. You should add all the files input to the compiler during compilation and remove all others. See the issue for more information: https://github.com/ethereum/sourcify/issues/618",
```
kuzdogan, line 224: This doesn't make sense: "...all all the files ... and remove all others"

We might have to change all occurances of this message in the codebase, but also we should always use error codes instead of messages when checking ie. in the tests
```suggestion
            "It seems your contract's metadata hashes match but not the bytecodes. If you are verifying via metadata.json, use the original full standard JSON input file that has all files including those not needed by this contract. See the issue for more information: https://github.com/ethereum/sourcify/issues/618",
```

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,500 @@
...
+        logWarn('Error checking for extra-file-input-bug', {
+          chain: this.sourcifyChain.chainId,
+          address: this.address,
+          error: e.message,
+        });
+        throw new VerificationError(
+          e.message,
+          'FAILED_TO_CHECK_EXTRA_FILE_INPUT_BUG',
+        );
+      }
```
kuzdogan, line 241: Why do we have this explicit check? Is there a known case of such errors? Seemed unnecessary to me

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -0,0 +1,348 @@
...
+  immutables?: {
+    [id: string]: string;
+  };
+  cborAuxdata?: {
+    [id: string]: string;
+  };
+}
+
+// returns the full bytecode with the call protection replaced with the real address
+export function checkAndCreateCallProtectionTransformation(
```
kuzdogan, line 86: Maybe rename these all functions `transformCallProtection` etc. ?

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,500 @@
...
+    }
+    return SolidityBugType.NONE;
+  }
+
+  private async matchBytecodes(
+    context: BytecodeMatchingContext,
+  ): Promise<BytecodeMatchingResult> {
```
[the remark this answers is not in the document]
marcocastignoli: I agree with you. We can use two parameters, before refactoring there were multiple parameters so I thought it would have been better to use a single object instead of a lot of parameters. Now that there are only 2 parameters it makes no sense.

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
```diff
@@ -176,20 +187,22 @@ export class VyperCompilation extends AbstractCompilation {
         this.auxdataStyle === AuxdataStyle.VYPER_LT_0_3_10 ||
         this.auxdataStyle === AuxdataStyle.VYPER_LT_0_3_5
       ) {
-        this.runtimeBytecodeCborAuxdata = this.tryGenerateCborAuxdataPosition(
-          this.getRuntimeBytecode(),
+        this._runtimeBytecodeCborAuxdata = this.tryGenerateCborAuxdataPosition(
+          this.runtimeBytecode,
```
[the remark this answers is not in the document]
marcocastignoli: Just because one case is creation and the other is runtime. Otherwise we should pass a `isCreation` parameter or similar. Do you prefer a `isCreation` parameter?

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
```diff
@@ -176,20 +187,22 @@ export class VyperCompilation extends AbstractCompilation {
...
       const [, creationAuxdataCbor, creationCborLengthHex] = splitAuxdata(
-        this.getCreationBytecode(),
+        this.creationBytecode,
         this.auxdataStyle,
       );
 
-      this.creationBytecodeCborAuxdata = this.tryGenerateCborAuxdataPosition(
-        this.getCreationBytecode(),
+      this._creationBytecodeCborAuxdata = this.tryGenerateCborAuxdataPosition(
+        this.creationBytecode,
```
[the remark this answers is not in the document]
marcocastignoli: Let's proceed this conversation in the comment below

packages/lib-sourcify/src/Compilation/SolidityCompilation.ts
```diff
@@ -58,6 +57,7 @@ export class SolidityCompilation extends AbstractCompilation {
         ],
       },
     };
+    delete this.jsonInput.settings.compilationTarget;
```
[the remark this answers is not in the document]
kuzdogan: IMO that should be a new type that extends Settings like MetadataCompilerSettings to avoid any type mistakes like mistakenly including the compilationTarget in the std-json output response.

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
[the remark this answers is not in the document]
kuzdogan: Ah right, no it's fine

packages/lib-sourcify/src/Verification/Verification.ts
kuzdogan: Sorry but saw this is resolved and not moved outside?

packages/lib-sourcify/src/Verification/Verification.ts
kuzdogan: Yes I think that's a little unnecessary. I don't think it's likely anything will go wrong here. We can propagate the error without catching or we could have a catch-all clause with a generic VerificationError. `FAILED_TO_CHECK_EXTRA_FILE_INPUT_BUG` seemed a bit too much

packages/lib-sourcify/src/Verification/Transformations.ts
kuzdogan: Ok I see and agree. Shall we say something like `extractConstructorArgsTransformation` instead of `checkAndCreate...` or do you want to keep that part too?

packages/lib-sourcify/src/Verification/Verification.ts
marcocastignoli: It is moved outside

https://github.com/ethereum/sourcify/blob/ed4fb932ea9c74ddc87f8718b7f6021a18b79594/packages/lib-sourcify/src/Verification/Verification.ts#L115-L126

packages/lib-sourcify/src/Verification/Verification.ts
marcocastignoli: I'll just remove the error for now

packages/lib-sourcify/src/Verification/Verification.ts
kuzdogan: Ok sorry my bad. I was looking at the old code I guess. Can be closed

marcocastignoli:
Should we wait merging because you also want to check the tests? @kuzdogan

marcocastignoli:
TODO

- [ ] https://github.com/ethereum/sourcify/pull/1898

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -0,0 +1,348 @@
...
+      // Extract the immutable value from the onchain bytecode.
+      const immutableValue = onchainRuntimeBytecode.slice(
+        start * 2,
+        start * 2 + length * 2,
+      );
+
+      // Save the transformation value
+      if (transformationValues.immutables === undefined) {
+        transformationValues.immutables = {};
+      }
```
kuzdogan, line 158: What's the purpose of this check? Isn't it always `undefined`? we assign `{}` to `transformationValues` just a few lines above.

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -0,0 +1,348 @@
...
+          normalizedRecompiledBytecode.slice(start * 2 + length * 2);
+      } else if (auxdataStyle === AuxdataStyle.VYPER) {
+        // For Vyper, insert the immutable value.
+        normalizedRecompiledBytecode =
+          normalizedRecompiledBytecode + immutableValue;
+      }
+    });
+  });
+  return {
+    normalizedRecompiledBytecode: '0x' + normalizedRecompiledBytecode,
```
kuzdogan, line 175: Same as the comment in `extractCallProtectionTransformation`, isn't this actually a `populatedBytecode` instead of "normalized"?

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,475 @@
...
+    };
+
+    // Replace library placeholders
+    const librariesTransformationResult = extractLibrariesTransformation(
+      normalizedRecompiledBytecode,
+      onchainBytecode,
+      linkReferences,
+    );
+    const libraryMap = librariesTransformationResult.libraryMap;
+    normalizedRecompiledBytecode =
```
kuzdogan, line 265: Same comment as `extractCallProtectionTransformation` and immutables transformation function regarding "normalized" vs "populated". Adding for completeness

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,475 @@
...
+      ? this.compilation.creationBytecodeCborAuxdata
+      : this.compilation.runtimeBytecodeCborAuxdata;
+
+    const transformations = isCreation
+      ? this.creationTransformations
+      : this.runtimeTransformations;
+    const transformationValues = isCreation
+      ? this.creationTransformationValues
+      : this.runtimeTransformationValues;
+
```
kuzdogan, line 246: Another reason (feeling) I notice why I wanted to review this file: Isn't this unnecessary? In `matchBytecodes` function the only new transformations we extract are the library and cborAuxdata transformations, because those are the only ones shared by both creation and runtime bytecodes. 

On these lines the `this.creation/runtimeTransformations` would be empty anyway?

Also the reason I say this is unnecessary is actually about the transformations assignments after lines 298 and additionally after 332. We spread the `result.transformations` with the library and cborAuxdata transformations and return it. 

But then outside the `matchBytecodes` function, ie. in `matchWithRuntime` and `matchWithCreation` we do the complete merging with `...transformations` again. 

I find this a bit confusing to where we really combine all transformations. To my understanding this `matchBytecodes` function should only return the library transformations and cbor if exist, and the merging of all transformations should only happen in one place, that is at the end of `matchWithRuntime` and `matchWithCreationBytecode` functions.

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,475 @@
...
+      : this.compilation.runtimeLinkReferences;
+
+    const result: BytecodeMatchingResult = {
+      match: null,
+      transformations: [...transformations],
+      transformationValues: { ...transformationValues },
+      normalizedRecompiledBytecode,
+    };
+
+    // Replace library placeholders
```
kuzdogan, line 258:
```suggestion
    // Library transformations can be in both the runtime bytecode and creation bytecode, hence done here in `matchBytecodes` method.
```

kuzdogan (changes requested):
Some small and some major things. I haven't looked at the tests yet to not make this review too large. I'll review the tests as the final next step.

My comments about the transformations array merging might be confusing. We can do a short call about that.

I'm reviewing this a bit thoroughly as this is an important piece of code but otherwise great work!

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
```diff
@@ -129,33 +130,43 @@ export class VyperCompilation extends AbstractCompilation {
...
+  get runtimeLinkReferences(): LinkReferences {
+    // Vyper doesn't support libraries
+    return {};
+  }
+
+  get creationLinkReferences(): LinkReferences {
+    // Vyper doesn't support libraries
+    return {};
+  }
+
```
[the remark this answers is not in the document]
marcocastignoli: Right now in `Verification` we just call compilation.runtimeLinkReferences without having to deal with errors / undefined. If we throw, then in `Verification` we need to handle in an `if(solidity)` when getting those values. 

What do you think?

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -0,0 +1,348 @@
...
+  const callProtection = `0x${push20CodeOp}${'00'.repeat(20)}`;
+
+  if (template.startsWith(callProtection)) {
+    const replacedCallProtection = real.slice(0, 0 + callProtection.length);
+    const callProtectionAddress = replacedCallProtection.slice(4); // remove 0x73
+    transformations.push(CallProtectionTransformation());
+    transformationValues.callProtection = '0x' + callProtectionAddress;
+
+    return {
+      normalizedRecompiledBytecode:
```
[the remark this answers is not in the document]
marcocastignoli: Actually I wondered many times if normalized was the right name, but I couldn't find anything better. Populated is perfect! Great suggestion

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,475 @@
...
+    const compiledRuntimeBytecode = this.compilation.runtimeBytecode;
+    const compiledCreationBytecode = this.compilation.creationBytecode;
+
+    if (compiledRuntimeBytecode === '0x' || compiledCreationBytecode === '0x') {
+      throw new VerificationError(
+        `The compiled contract bytecode is "0x". Are you trying to verify an abstract contract?`,
+        'COMPILED_BYTECODE_IS_ZERO',
+      );
+    }
+
```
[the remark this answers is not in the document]
marcocastignoli: I'm wondering are we sure that that's the case? I'm thinking about two cases:
1. creation bytecode length differs because of the constructor arguments
2. In Vyper, immutables generated in deployment phase are appended to the end of the bytecode. (check test `should add constructor transformation with correct offset and arguments for Vyper contract` in `packages/lib-sourcify/test/Verification/Verification.spec.ts`)

packages/lib-sourcify/src/Verification/Transformations.ts
marcocastignoli: We assign `const transformationValues: TransformationValues = {};` but not

```ts
const transformationValues: TransformationValues = {
  immutables: {}
};
```

then we enter a loop in which multiple immutables are stored. For the first one `transformationValues.immutables` is empty and we have to initialize it.

packages/lib-sourcify/src/Verification/Verification.ts
marcocastignoli: I think I understood but I didn't get what's your alternative solution? Just replicate the functions calls directly inside `matchWithRuntime` and `matchWithCreationBytecode`?

packages/lib-sourcify/src/Verification/Verification.ts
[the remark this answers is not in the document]
marcocastignoli: > I was referring to runtime bytecode lengths

Ok

> This can only apply to Solidity. I think in Vyper we can check if the recompiled bytecode is longer than onchain

so basically we do 

```ts
if (
(solidity && recompiledRuntime.length !== onchainRuntime.length) 
or
(vyper && onchainRuntime.length < recompiledRuntime.length)
) {
throw
}
```

packages/lib-sourcify/src/Verification/Transformations.ts
kuzdogan: Ok I see. Just, this check always passes so we can just assign {} inside the loop without the if. 

Just a small thing

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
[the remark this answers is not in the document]
marcocastignoli: > Does this mean it will also write {} to the linkrefernces field in the DB downstream

The cases of what we store in the database are not related to the output of the lib-sourcify library. In `server/.../*database*.ts` classes we handle every case appropriately.

> I think it can at least be undefined.

I see your point, but if you don't have a strong opinion I prefer to keep an empty object since it's how we deal with other similar fields such as `AbstractCompilation.run/creBytecodeCborAuxdata`, `Verification.run/creTransformations`, `Verification.run/creTransformationsValues`, `Verification.run/creLibraryMap`. They all return empty objects (even fields that are nullable are always set to `{}` when they are initialized if empty).

packages/lib-sourcify/src/Verification/Transformations.ts
marcocastignoli: You mean outside the loop at the beginning? If we were to assign it without the if we would override any previously assigned key/value

kuzdogan: Ahh ok now I see. This check does not always pass, it will only pass in the first loop run. 

Just ignore my comment.

packages/lib-sourcify/src/Verification/Verification.ts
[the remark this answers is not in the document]
kuzdogan: Yes. Also let's add some comments too

packages/lib-sourcify/src/Verification/Transformations.ts
marcocastignoli: np :)

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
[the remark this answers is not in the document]
kuzdogan: Ok fine for me

packages/lib-sourcify/src/Verification/Verification.ts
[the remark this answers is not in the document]
marcocastignoli: I implemented it but then I had a problem with this test:

```ts
@ethereum-sourcify/lib-sourcify:   2) Verification Class Tests
@ethereum-sourcify/lib-sourcify:        Basic Contract Verification
@ethereum-sourcify/lib-sourcify:          should detect extra file input bug when optimizer is enabled:
@ethereum-sourcify/lib-sourcify:       AssertionError: expected 'BYTECODE_LENGTH_MISMATCH' to equal 'EXTRA_FILE_INPUT_BUG'
@ethereum-sourcify/lib-sourcify:       + expected - actual
@ethereum-sourcify/lib-sourcify:       -BYTECODE_LENGTH_MISMATCH
@ethereum-sourcify/lib-sourcify:       +EXTRA_FILE_INPUT_BUG
```

Apparently if we do the bytecode length check then `extra file input bug` is not compatible anymore. I think for now it's better to rollback these changes. What do you think @kuzdogan ?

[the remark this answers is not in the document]
kuzdogan: Hmm weird this should not happen. extra-file-input only changes AST IDs and the bytecode lengths should not be affected. I tried to debug into it and my tests passed.

Regardless IMO this is a quite important validation for verification and will save us a lot of unnecessary runs, whereas extra-file-input is quite rare and it's also just a thing when verifying with metadata.json. If you can't resolve it you can disable the test for now and open another issue and we'll think in detail how to handle this and if indeed the bytecode lengths really change.

packages/lib-sourcify/src/Verification/VerificationTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+export type VerificationErrorCode =
+  | 'CANT_FETCH_BYTECODE'
+  | 'CONTRACT_NOT_DEPLOYED'
+  | 'COMPILED_BYTECODE_IS_ZERO'
+  | 'EXTRA_FILE_INPUT_BUG'
+  | 'CREATION_BYTECODE_MATCH_ERROR'
+  | 'NO_MATCH'
+  | 'ONCHAIN_RUNTIME_BYTECODE_NOT_AVAILABLE'
+  | 'ONCHAIN_CREATION_BYTECODE_NOT_AVAILABLE'
+  | 'BYTECODE_LENGTH_MISMATCH';
```
manuelwedler, line 23: Can we still rename the error codes to use lower snake_case? That way they would match the format of API v2 errors and we could directly use them.

packages/lib-sourcify/src/Verification/Verification.ts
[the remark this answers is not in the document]
marcocastignoli: On the test I implemented in lib-sourcify that's exactly the case. Now I refactor the code a bit to check the bug before throwing the bytecode length mismatch. Tell me if you like this solution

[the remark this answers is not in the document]
kuzdogan: It's ok like this but I'd still like to see why this has happened ie. why lengths are different. I'll defer that to a separate issue.

packages/lib-sourcify/test/Verification/Verification.spec.ts
```diff
@@ -0,0 +1,1213 @@
...
+        compilation,
+        sourcifyChainHardhat,
+        UNUSED_ADDRESS,
+      );
+
+      try {
+        await verification.verify();
+      } catch (err: any) {
+        expect(err.code).to.equal('contract_not_deployed');
+      }
```
kuzdogan, line 221: I think this is an incorrect way to test thrown errors. If the verification does not throw, the test will still pass because it will not enter any `expect()` clauses. 

Such errors should be tested with 

```ts
await expect(verification.verify()).to.be.rejectedWith().and.have.property('code', 'contract_not_deployed');
```
or
```ts
expect(() => verification.verify()).to.throw().with.property('code', 'contract_not_deployed');
```

You can ask an LLM for this

packages/lib-sourcify/test/Verification/Verification.spec.ts
```diff
@@ -0,0 +1,1213 @@
...
+  for (const [sourcePath] of Object.entries(metadata.sources)) {
+    // Extract filename from source path (e.g. "contracts/Storage.sol" -> "Storage.sol")
+    const fileName = sourcePath.split('/').pop() || sourcePath;
+    const filePath = path.join(sourcesPath, fileName);
+    const content = fs.readFileSync(filePath, 'utf8');
+    sources.push({
+      path: sourcePath,
+      content,
+    });
+  }
```
kuzdogan, line 59: I guess this is passing for all cases now, but shouldn't this be actually getting the sources from the filesystem instead of assuming the paths in the metadata file?

Ie. here you are looking at the file paths in metadata.sources field. This assumes the paths in the metadata structure is the same as in the tests. But this doesn't has to be the case since we actually check the files exist with their hashes and not their paths. So the "correct" way would be to recursively traverse the "sources" folder and add every source file to the sources object.

Not a big issue but to me this feels to be the right way.

packages/lib-sourcify/test/Verification/Verification.spec.ts
```diff
@@ -0,0 +1,1213 @@
...
+
+      expectVerification(verification, {
+        status: {
+          runtimeMatch: 'partial',
+          creationMatch: null,
+        },
+      });
+    });
+
+    it('should return null match when there is no perfect match and no auxdata', async () => {
```
kuzdogan, line 356: I don't fully get what we are testing here? To me this is simply testing null match when bytecodes don't match?

packages/lib-sourcify/test/Verification/Verification.spec.ts
```diff
@@ -0,0 +1,1213 @@
...
+
+      expectVerification(verification, {
+        status: {
+          runtimeMatch: 'perfect',
+          creationMatch: null,
+        },
+      });
+    });
+
+    it('should fail when chain is temporarily unavailable', async () => {
```
kuzdogan, line 504:
```suggestion
    it('should fail when bytecode could not be fetched', async () => {
```

packages/lib-sourcify/test/Verification/Verification.spec.ts
```diff
@@ -0,0 +1,1213 @@
...
+      );
+      await verification.verify();
+
+      expectVerification(verification, {
+        status: {
+          runtimeMatch: 'perfect',
+          creationMatch: 'perfect',
+        },
+      });
+    });
```
kuzdogan, line 598: Don't we need to check for transformations too?

kuzdogan (changes requested):
Couple issues. I didn't really check how comprehensive these tests are and if more are needed. Also there are a lot of potential for refactoring. For the sake of moving forward I'm leaving these out for this PR.

packages/lib-sourcify/test/Verification/Verification.spec.ts
marcocastignoli: I changed the test's description. But it's this case: https://github.com/ethereum/sourcify/blob/de425e2bb787d6bb2457eb66f4a6fb46e466168d/packages/lib-sourcify/src/Verification/Verification.ts#L337-L340

packages/lib-sourcify/test/Verification/Verification.spec.ts
marcocastignoli: I see your point. It was just to keep the previous sources structure and I thought in this way it would work. What do you think it's best to do here? Maybe we'll do a final iteration on this tests util details later?

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -0,0 +1,501 @@
...
+      }
+
+      result.libraryMap = libraryMap;
+      result.transformations = librariesTransformationResult.transformations;
+      result.transformationValues =
+        librariesTransformationResult.transformationValues;
+      return result;
+    }
+
+    // If there is no perfect match and no auxdata, return null
```
kuzdogan, line 337:
```suggestion
    // If bytecodes don't match and no auxdata to ignore, return null
```

packages/lib-sourcify/test/Verification/Verification.spec.ts
```diff
@@ -0,0 +1,1227 @@
...
+
+      expectVerification(verification, {
+        status: {
+          runtimeMatch: 'partial',
+          creationMatch: null,
+        },
+      });
+    });
+
+    it('should throw an error when there is no perfect match and no auxdata', async () => {
```
kuzdogan, line 357:
```suggestion
    it('should throw an error when bytecodes dont match and no auxdata to ignore for a partial match', async () => {
```

packages/lib-sourcify/test/Verification/Verification.spec.ts
kuzdogan: Ah ok got it. "no perfect match and no auxdata" was confusing to me. I made the changes to the test title and code comment to make it clear.

PR #1881, opened 2025-01-30
Reorganize database migrations

### LINKED ISSUE #1865: Reorganize database migrations
We want to reorganize our database migration scripts. There should only be two scripts, that will be extended if we make schema changes in the future. The scripts should be:

1. VerA schema (exactly as in https://github.com/verifier-alliance/database-specs)
2. Sourcify's changes to the VerA schema

Note that the validation functions are currently missing from VerA spec and they should be included now.

manuelwedler:
Closes #1865

This reduces our db migrations to two files: one for VerA db spec and one for the Sourcify changes to it. The VerA database-specs is added as a git submodule, and its `schema.sql` file is just read in the first migration, instead of duplicating the code. 

In the future, when VerA db specs are updated, we just need to update the submodule commit. Sourcify additions should just be added to the Sourcify migration file.


services/database/migrations/20231109160022-create-alliance-schema.js
```diff
@@ -17,412 +23,122 @@ exports.setup = function (options, seedLink) {
...
+        'DROP FUNCTION IF EXISTS "validate_runtime_transformations";',
+      ),
+      db.runSql.bind(
+        db,
+        'DROP FUNCTION IF EXISTS "validate_creation_transformations";',
+      ),
+      db.runSql.bind(db, 'DROP FUNCTION IF EXISTS "validate_transformations";'),
+      db.runSql.bind(
+        db,
+        'DROP FUNCTION IF EXISTS "validate_transformations_call_protection";',
```
kuzdogan, line 58: Why were were dropping all these checks again? Maybe we should add a comment here.

kuzdogan (approved):
Nice an clean. Just a comment

services/database/migrations/20231109160022-create-alliance-schema.js
manuelwedler: It's the down function. It should restore the database state before the migration. I can add a comment for that.

daniil-parsiq:
I just wanted to ask you guys if you're okay there, living happy life while merging all migrations into one file? Who cares about people upgrading from older versions of Sourcify anyway?

manuelwedler:
@daniil-parsiq 

The issue here was that, most of the time, the migrations were not correctly updated. Therefore, the migrations were hardly usable for updating a Sourcify database. To address the issue you are talking aobut, we would need migrations on the upstream repo  (https://github.com/verifier-alliance/database-specs) and make sure they are maintained there whenever a schema change is made. Everything else is not maintainable, imo.

Are you running a Sourcify instance? We are open to address this issue if there is the need. Feel free to reach out to us via matrix / discord / e-mail and we can discuss this.

daniil-parsiq:
@manuelwedler 

thanks for the explanation 😅 We did solve the problem on our end, however through manual interventions. We were running Sourcify 2.2.1 previously and got quite surprised when the new version didn't work while migration jobs saying all good and successful. I understand difficulties with the upstream repo, but this change defied the whole purpose of migrations and versioning hence the angry commenting!

daniil-parsiq:
Hi @marcocastignoli,  @manuelwedler  coming back to this... Was bumping Sourcify version from 3.0.0 to 3.7.0 today, I hope you know how I feel about you all :DDD 

Another funny case that actually made me update Sourcify version is that updating foundry complete broke verification:
```
2025-11-19T14:56:54.895484Z DEBUG reqwest::connect: starting new connection: https://sourcify.rnk.dev/

Submitting verification for [WrappedReact] "0xe00CBca00d36c89819289dE34e352881E8F475Fd".
2025-11-19T14:56:55.488586Z DEBUG reqwest::connect: starting new connection: https://sourcify.rnk.dev/
Warning: error decoding response body; expected value at line 1 column 1; waiting 5 seconds before trying again (4 tries remaining)
^C
```

While logs of Sourcify server were just quiet. Updating to 3.7.0 helped

marcocastignoli:
hey @daniil-parsiq, I cannot understand what's the problem. Is it because the new Foundry wasn't compatible with sourcify server 3.0.0?

manuelwedler:
Hi @daniil-parsiq , Foundry moved to Sourcify API v2 which is available from server version 3.1.0 and above. We reached out to all custom Sourcify instances we know of about this already one or two months ago. We have let them know they should upgrade to still be compatible with Foundry (and Hardhat v3 which also moved).

We usually have a direct communication channel with anyone running custom Sourcify instances. I invited you earlier to get in touch with us. This would have helped with you being notified about stuff like this. I'd like to invite you another time to reach out, either to hello@sourcify.dev or via [Matrix](https://matrix.to/#/#ethereum_source-verify:gitter.im), such that we can create a direct communication channel, too.

PR #1893, opened 2025-02-03
Fix not publishing to npm and throw on errors in the bash scripts

kuzdogan:
Moves the function definition above the commands to fix the issue. Adds `set -e` lines to throw on errors


manuelwedler (approved):
lgtm

Issue #1826, opened 2025-01-06
Iterate on the new `verification_jobs` table


marcocastignoli:
Iterate on the new table definition, making sure that it contains all the need information. In particular focus on the foreign key `verification_id` in the case of a successful verification and understand what information to include in the case of a failed verification:
- jobId
- endpoint: `/v2/verify/etherscan/`
- chainId
- address
- status
- server hardware
- compilation time
- errors (compiler failures)
- createdAt
- finishedAt (success or fail)
- when successful: verification-id
- Do we want to store information when the compilation succeeds but match fails?
  - onchain bytecodes
  - recompiled bytecodes

manuelwedler:
I mostly agree with above fields. Errors need multiple fields due to the API design in Stoplight. I propose to use exactly these names for the columns:

### Table: `verification_jobs`

Needed for the API response:
- `id`: bigserial, primary key, incremental
- `status`: varchar, not null
- `started_at`: timestamptz, not null
- `completed_at`: timestamptz
- `chain_id`: numeric, not null
- `contract_address`: bytea, not null
- `verified_contract_id`: bigserial, foreign key to `verified_contracts.id`
- `error_code`: varchar
- `error_message`: varchar
- `error_id`: uuid

These are only needed for stats:
- `verification_endpoint`: varchar, not null
- `server_hardware`: jsonb, not null
- `compilation_time`: interval

Regarding `server_hardware`, I am not sure if we can get helpful information for it, since we run it in Cloud Run and CPU and memory allocations are just limits. We can try to use the Node.js functions for those (`os.cpus()`, `os.totalmem()`, ...). Let's see what they return under this environment. I think it makes sense to store this information for each verification job in the following json format:

```
{
  "cpu": string,
  "memory": string
}
```

Anything else that should be added?


### Additional information?

> Do we want to store information when the compilation succeeds but match fails?
> * onchain bytecodes
> * recompiled bytecodes

In my opinion, we should not store more information than necessary, because I don't want to blow the database. The error_code with the error_message should be enough to tell the user what went wrong. If we want to debug something, we should be able to find the corresponding logs by using the error_id.

kuzdogan:
Small comment on `chain_id`, I think it should be `bigInt` instead of `numeric`, see https://github.com/verifier-alliance/database-specs/issues/23

kuzdogan:
[...]

PR #1902, opened 2025-02-05
Add migrations for `verification_jobs` and `verification_jobs_ephemeral` tables

### LINKED ISSUE #1884: Add `verification_jobs` table to migrations
After finalizing the discussion with #1826 implement the changes and add them to Sourcify migrations

manuelwedler:
Closes #1884

I created the migrations for the `verification_jobs` table as in my last proposal in #1826 , not including the specific server hardware, because its use case is not entirely clear to me. The `hardware` column should be `"cloud_run:${process.env.K_REVISION}"` as Marco proposed.


manuelwedler:
In chat, we decided to remove the onchain codes from the ephemeral table and add the `creator_transaction_hash` instead there. Last commit adds this changes.

marcocastignoli (approved):
👍

kuzdogan:
I thought we also agreed to include a `verification_endpoint` field no? This will be esp. useful if we add verification jobs for v1 verifications too (for compialtion time etc.). Let's discuss this on Monday

manuelwedler:
`verification_endpoint` is included here.

PR #1914, opened 2025-02-07
Compilers package

### LINKED ISSUE #1348: "packagizing" the `useCompiler`
We use a function `useCompiler` in various places to compile contracts. This function provides a small wrapper around the solc compiler in order to download it and invoke the compilation.

The code was duplicated across multiple places, namely in the tests of `lib-sourcify`, in `serverless/compiler-lambda` and in the server code. In order to minimize the effort for maintenance and synchronization of these code parts, they should be moved into a public package. Basically the content of `services/server/src/server/services/compiler/local/solidityCompiler.ts` should be the content of the new package.

### Tasks
- [x] Create new package `compiler-wrapper` making the functionality of `solidityCompiler.ts` available
- [x] Use the package inside the tests of `lib-sourcify`
- [x] Use the package inside the server

manuelwedler:
Closes #1348 

This simply moves the useCompiler functions to a new public package without refactoring the code much.


packages/compilers/test/solidityCompiler.spec.ts
```diff
@@ -0,0 +1,156 @@
+import { expect } from 'chai';
+import {
+  getSolcExecutable,
+  getSolcJs,
```
marcocastignoli, line 4: I see you didn't include `VyperCompiler.spec.ts`

packages/compilers/src/lib/vyperCompiler.ts
```diff
@@ -37,8 +35,8 @@ export function findVyperPlatform(): string | false {
 export async function useVyperCompiler(
   vyperRepoPath: string,
   version: string,
-  vyperJsonInput: VyperJsonInput,
-): Promise<VyperOutput> {
+  vyperJsonInput: any,
```
marcocastignoli, line 38: In the new verification flow PR there are a lot of new types like this one. I'm wondering if we should move them in this new package. Also I see here you removed the type, why?

packages/compilers/package.json
```diff
@@ -0,0 +1,83 @@
+{
```
marcocastignoli, line 1: We should also setup running the tests in the debugger with .vscode configuration

marcocastignoli (changes requested):
I see many compilers types were removed, I think we should coordinate to use the types we declared in the verification refactoring PR

packages/compilers/test/solidityCompiler.spec.ts
marcocastignoli: Ah ok, I see you just included existing tests. Let's implement the Vyper tests in a separate PR

packages/compilers/src/lib/vyperCompiler.ts
marcocastignoli: Mhhh I already don't like this :D The only solution I see other than including the types in the compiler package is to create another `compiler-types` package

manuelwedler: I'm not sure how to move forward here. Another package for the types would definitely solve this, but it also seems a bit too much to me. @kuzdogan What do you think about this issue?

kuzdogan: We discussed this shortly in the call and the conclusion was to use any for now here but open another issue to export types to a new package. 

Correct me if I'm wrong.

manuelwedler: issue: #1958

marcocastignoli (approved):
lgtm

PR #1934, opened 2025-02-18
Rename `*` field of v2 contract lookup endpoint to `all`

### LINKED ISSUE #1926: Change or add a new wildcard for `?fields=*` for APIv2
Since * is outside the URL spec and allowed characters, it is repeatedly being excluded in pasted URLs and examples of APIv2.

I think we should instead opt into an allow character of maybe just an `all` wildcard to include everything.

E.g.

https://sourcify.dev/server/v2/contract/1/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48?fields=*

or this Tweet

Image

manuelwedler:
Closes #1926


services/server/src/apiv2.yaml
```diff
@@ -905,13 +905,6 @@ components:
                     chainId: '11155111'
                     address: '0x2738d13E81e30bC615766A0410e7cF199FD59A83'
                     verifiedAt: '2024-07-24T12:00:00Z'
-                pagination:
-                  currentPage: 3
-                  totalPages: 243
-                  resultsPerPage: 5000
-                  totalResults: 24245
-                  hasNextPage: true
-                  hasPreviousPage: true
```
marcocastignoli, line 914: Why this was removed?

marcocastignoli (commented):
just a question

services/server/src/apiv2.yaml
manuelwedler: We don't have it anymore, because we changed the pagination style. This is only here because I copied the api v2 spec from Stoplight to be up-to-date. Seems like this change was forgotten earlier.

marcocastignoli (approved):

PR #1944, opened 2025-02-19
Ensure to return NULL for any NULL value on the GET contract endpoint

### LINKED ISSUE #1941: Invalid argument when `fields=all` is provided to lookup endpoint
https://staging.sourcify.dev/server/v2/contract/10/0x06C6D063896ac733673c4474E44d9268f2402A55?fields=all

This API call fails for the following error:

```
{
"error": "invalid address (argument=\"address\", value=\"0x\", code=INVALID_ARGUMENT, version=6.13.5)",
"message": "invalid address (argument=\"address\", value=\"0x\", code=INVALID_ARGUMENT, version=6.13.5)"
}
```

Weirdly, another address is just fine:

https://sourcify.dev/server/v2/contract/1/0x36911E4118D890c9e9897E5b4D663225CDabd4A9?fields=all

Without the `?fields=all` both addresses are also fine

manuelwedler:
Fixes #1941 

The `concat('0x', ...)` function simply returns `0x` in case the second argument is NULL. This caused the problem from the linked issue.

### DIFF
--- a/services/server/src/server/services/utils/database-util.ts
+++ b/services/server/src/server/services/utils/database-util.ts
@@ -230,11 +230,11 @@ export const STORED_PROPERTIES_TO_SELECTORS = {
   verified_at:
     'to_char(sourcify_matches.created_at, \'YYYY-MM-DD"T"HH24:MI:SS"Z"\') as verified_at',
   address:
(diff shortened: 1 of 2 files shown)


marcocastignoli (approved):
lgtm

PR #1956, opened 2025-02-28
Move api-docs route before fallback handler

### LINKED ISSUE #1953: Serving the swagger docs is broken
I realized that the error handling PR for API v2 (https://github.com/ethereum/sourcify/pull/1946) broke our Swagger UI docs.

See https://staging.sourcify.dev/server/api-docs/

returns:
```json
{"customCode":"route_not_found","message":"The requested resource was not found","errorId":"cd41cfb4-b40e-4c3e-95f3-42f3a83bfe87"}
```

Seems like the swagger docs route is matched later than the 404 fallback.

manuelwedler:
Fixes #1953 

The api-docs route was created in the cli file because it uses asynchronous code (`loadSwagger`). For that reason, it is added after the route not found fallback, which is located in the server constructor. This causes the issue linked above.

To move the api-docs route before the fallback, I changed the server creation method from the constructor to a static async create method. This way we can call asynchronous code in the creation method.

This means instead of calling `new Server(...)`, we need to create the server by `await Server.create(...)`. I did it like this because I think it is the cleaner approach, but this is also a breaking change for anyone using the server module of course.


kuzdogan (changes requested):
TBH this doesn't feel right. I don't think we require a breaking change how to load the Server module to fix this issue.

Does this line really have to be there? 
```ts

    // Any request that could not be handled by a route earlier will match this
    this.app.use((req, res, next) => {
      next(new RouteNotFoundError("The requested resource was not found"));
    });
```
We already had error handlers and they weren't causing this problem. This leads me to think the error is caused by this line being not interpreted as an error handler. Is this really the correct way to throw when no matching routes are found? 

Or do we even need that line? If I completely remove it we already get a 404:

image

kuzdogan:
Here's also another workaround from Claude:

## Response

I see the issue. The problem is that your `/api-docs` endpoint is being caught by the catch-all route handler that throws a `RouteNotFoundError` before it can be properly handled. Since you're adding the `/api-docs` endpoint in the cli.ts file (after the Server class is instantiated), it's not being registered before the catch-all handler.

Here's how you can fix this:

```typescript:services/server/src/server/server.ts
// ... existing code ...

    this.app.use("/", routes);

    // Add a method to register routes after initialization
    public registerRoutes(path: string, router: express.Router) {
      this.app.use(path, router);
    }

    // Any request that could not be handled by a route earlier will match this
    this.app.use((req, res, next) => {
      next(new RouteNotFoundError("The requested resource was not found"));
    });

// ... existing code ...
```

This adds a new `registerRoutes` method to the Server class that allows you to register additional routes after the Server has been initialized but before the catch-all 404 handler processes the request.

Then in your cli.ts file, you can use this method to register the Swagger UI route after initializing the server but before starting it:

```typescript
// In cli.ts where you set up the server
const server = new Server(/* your options */);

// After server initialization but before server.listen()
const swaggerUi = require('swagger-ui-express');
const swaggerDocument = await server.loadSwagger(/* your swagger options */);
const swaggerRouter = express.Router();
swaggerRouter.use('/', swaggerUi.serve, swaggerUi.setup(swaggerDocument));
server.registerRoutes('/api-docs', swaggerRouter);

// Then start the server
await server.listen();
```

This approach allows you to add the Swagger UI routes after the Server is constructed but ensures they're registered before the catch-all 404 handler processes the request.

manuelwedler:
> Here's also another workaround from Claude:

That code is simply not working. You cannot add a method in the constructor like this. Also if you do, the `app.use` still gets called later for api docs than the fallback route.

> Does this line really have to be there?

Imo, it should be. We can move it to `cli.ts` to call it later than the api-docs routes but then we would need to also move the error handlers which we should not do.

> We already had error handlers and they weren't causing this problem. This leads me to think the error is caused by this line being not interpreted as an error handler. Is this really the correct way to throw when no matching routes are found?

This line is not an error handler. It's a fallback handler and throws an error which is then handled by our error handlers. Yes, this is the right way. See express docs: https://expressjs.com/en/starter/faq.html#how-do-i-handle-404-responses

> Or do we even need that line? If I completely remove it we already get a 404:

I think that's the other option. We can just rely on the default 404 of express.
However, I still think adding the api-docs routes in the `cli.ts` is bad design.

kuzdogan:
Ok yeah the Claude code is not an option, I rushed to post that a bit.

Can we instead move the swagger call to server.ts and make it synchronous, and move above the new 404 fallback? Either we make the `loadSwagger` part ie. essentially the `loadRefs` function synchronous with a different method or library, or the worse case, we keep the async call but don't await it, it should eventually load the swagger docs.

I also see us dropping the swagger-ui as it's worse than stoplight IMO, but that's something for later.

manuelwedler:
I tried a few other options / libraries for making the `loadSwagger` call synchronous, but could not get anything working. 

I also don't want to call the method without awaiting it. That will make tests flaky (I already had a failing test when I tried this locally). 

Should we just remove the fallback handler? That seems the easiest solution here, but I still think adding routes in `cli.ts` is an anti-pattern and should be resolved in the future.

manuelwedler:
Removed the fallback handler in the new commit

kuzdogan (approved):

PR #1969, opened 2025-03-06
Fix creation match upgrade

marcocastignoli:
This PR:
- [x] Fixes the test "Should upgrade creation match from 'null' to 'perfect'" to catch the not-updated verified_contract
- [x] Updates verified_contracts creation table at creation match upgrade.

See https://github.com/ethereum/sourcify/issues/1974


services/server/src/server/services/storageServices/AbstractDatabaseService.ts
```diff
@@ -419,48 +419,43 @@ export default abstract class AbstractDatabaseService {
...
-        | undefined;
       let onchainCreationCodeInsertResult:
         | QueryResult<Pick<DatabaseUtil.Tables.Code, "bytecode_hash">>
         | undefined;
-      // Check if contracts_deployed needs to be updated
-      if (
-        existingVerifiedContractResult[0].transaction_hash === null &&
-        match.creatorTxHash != null &&
-        databaseColumns.onchainCreationCode
-      ) {
```
marcocastignoli, line 433: @kuzdogan @manuelwedler I want your feedback about this. Should we insert a new contract_deployment only if the existing row doesn't have creation information or always?

I think we should **always do it** because the `verified_contracts` information are linked with the new `contract_deployment` that we are about to insert. If we skip insertion then the `verified_contract` will reference to the old `contract_deployment` that is potentially different.

marcocastignoli: (Notice that even if we leave this "if" condition, the bug we encountered is solved)

marcocastignoli: Resolved in a call: it's better to always update because the `contract_deployments` inserts are all idempotent.

PR #1978, opened 2025-03-07
Add information about connecting to an RPC on the Docker host

sealer3:
This adds a short subsection to the server docs explaining how to connect Sourcify to an RPC server running on the Docker host. I had to spend some time searching online for this before I was able to get Sourcify to connect to a development node on my machine, so maybe this information would be useful for others, too.


kuzdogan (approved):
Good to know, thanks!

PR #1993, opened 2025-03-12
Improve auxdata position generation return types and error handling

marcocastignoli:


packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -125,7 +125,12 @@ export class Verification {
 
     // We need to manually generate the auxdata positions because they are not automatically produced during compilation
     // Read more: https://docs.sourcify.dev/blog/finding-auxdatas-in-bytecode/
-    await this.compilation.generateCborAuxdataPositions();
+    try {
+      await this.compilation.generateCborAuxdataPositions();
+    } finally {
```
manuelwedler, line 130:
```suggestion
    } catch {
```

packages/lib-sourcify/src/Compilation/CompilationTypes.ts
```diff
@@ -125,3 +126,15 @@ export interface CompilationTarget {
 }
 
 export type CompilationLanguage = 'Solidity' | 'Vyper';
+
+export type CompilationErrorCode =
+  | 'cannot_generate_cbor_auxdata_positions'
+  | 'invalid_compiler_version'
+  | 'contract_not_found_in_compiler_output'
```
manuelwedler, line 133: I don't see this error code used anywhere. Also, there are still usual `Error` instances thrown in the compilation classes. Will you change that later?

marcocastignoli: > . Also, there are still usual Error instances thrown in the compilation classes. Will you change that later?

I kept standard Errors because they are not actually CompilationError, and I thought it would be confusing to throw them as CompilationError. What do you think?

manuelwedler: ah right these are the getter functions. Then it's fine

manuelwedler (approved):

Issue #1910, opened 2025-02-07
Implement POST `/v2/verify/{chainId}/{address}` endpoint


manuelwedler:
New verify endpoint using standard json as input

Specification in [Stoplight](https://sourcify.stoplight.io/docs/sourcify-apiv2/h55xvo5bsu9qz-verify-contract-standard-json)

marcocastignoli:
Related: https://github.com/ethereum/sourcify/issues/1875

PR #1994, opened 2025-03-12
Add VerificationExport

manuelwedler:
Needed for API v2 verify endpoints (https://github.com/ethereum/sourcify/issues/1910). The export will be returned from the verification worker.

### DIFF
--- a/packages/lib-sourcify/src/Verification/VerificationTypes.ts
+++ b/packages/lib-sourcify/src/Verification/VerificationTypes.ts
@@ -1,4 +1,20 @@
-import { StringMap } from '../Compilation/CompilationTypes';
+import { Abi } from 'abitype';
+import {
+  CompilationLanguage,
+  CompilationTarget,
+  CompiledContractCborAuxdata,
+  Devdoc,
+  LinkReferences,
+  Metadata,
+  StringMap,
+  Userdoc,
+} from '../Compilation/CompilationTypes';
+import {
+  ImmutableReferences,
+  SoliditySettings,
+  StorageLayout,
+} from '../Compilation/SolidityTypes';
+import { VyperSettings } from '../Compilation/VyperTypes';
 import { SourcifyLibError } from '../SourcifyLibError';
 import { Transformation, TransformationValues } from './Transformations';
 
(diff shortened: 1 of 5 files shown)


packages/lib-sourcify/src/Verification/VerificationTypes.ts
```diff
@@ -40,3 +56,67 @@ export type VerificationStatus =
...
+          sourceMap?: string;
+          linkReferences?: LinkReferences;
+        };
+        deployedBytecode: {
+          sourceMap?: string;
+          linkReferences?: LinkReferences;
+          immutableReferences?: ImmutableReferences;
+        };
+      };
+    };
```
marcocastignoli, line 112: I'm wondering here can we use `SolidityOutputContract | VyperOutputContract` as type?

packages/lib-sourcify/src/Verification/VerificationTypes.ts
```diff
@@ -40,3 +56,67 @@ export type VerificationStatus =
...
+    runtime?: StringMap;
+    creation?: StringMap;
+  };
+  compilation: {
+    language: CompilationLanguage;
+    compilationTarget: CompilationTarget;
+    compilerOutput: {
+      // The export should not include the AST object to reduce the size
+      sources?: Record<string, { id: number }>;
+    };
```
marcocastignoli, line 95: Why here we are exporting only sources? I understand we want to reduce the size, but the full objects contain much more information:

```
export interface SolidityOutput {
  errors?: SolidityOutputError[];
  sources?: SolidityOutputSources;
  contracts: {
    [globalName: string]: {
      [contractName: string]: SolidityOutputContract;
    };
  };
}

export interface VyperOutput {
  compiler: string;
  errors?: VyperError[];
  sources: {
    [sourcePath: string]: VyperOutputSource;
  };
  contracts: VyperOutputContracts;
}
```

marcocastignoli (changes requested):

packages/lib-sourcify/src/Verification/VerificationTypes.ts
manuelwedler: They contain more output than we use. Therefore, I minimized it that way.

packages/lib-sourcify/src/Verification/VerificationTypes.ts
marcocastignoli: Right

packages/lib-sourcify/src/Verification/VerificationTypes.ts
marcocastignoli: Ok for now let's keep it like this, if in the future someone needs more fields we can add them

marcocastignoli (approved):
lgtm

PR #1995, opened 2025-03-12
Use VerificationExport for storeVerification methods of storage services

manuelwedler:
Follow up PR from #1994

### DIFF
--- a/services/server/src/server/services/storageServices/AbstractDatabaseService.ts
+++ b/services/server/src/server/services/storageServices/AbstractDatabaseService.ts
@@ -1,4 +1,4 @@
-import { Verification } from "@ethereum-sourcify/lib-sourcify";
+import { VerificationExport } from "@ethereum-sourcify/lib-sourcify";
 import * as DatabaseUtil from "../utils/database-util";
 import { bytesFromString } from "../utils/database-util";
 import { Database, DatabaseOptions } from "../utils/Database";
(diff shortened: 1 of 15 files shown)


services/server/src/server/services/storageServices/AbstractDatabaseService.ts
```diff
@@ -16,12 +16,10 @@ export default abstract class AbstractDatabaseService {
...
   }
 
-  validateVerificationBeforeStoring(verification: Verification): boolean {
+  validateVerificationBeforeStoring(verification: VerificationExport): boolean {
     if (
       verification.compilation.runtimeBytecode === undefined ||
-      verification.compilation.creationBytecode === undefined ||
-      verification.status.runtimeMatch === undefined ||
-      verification.status.creationMatch === undefined
+      verification.compilation.creationBytecode === undefined
```
marcocastignoli, line 22: No, here we want to store only if we have creation and runtime match

marcocastignoli: because this is the generic `validateVerificationBeforeStoring` used by the VerA database

marcocastignoli (changes requested):
I like it much more now! A couple of comments:

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -535,7 +535,7 @@ export class SourcifyDatabaseService
     return response;
   };
 
-  validateVerificationBeforeStoring(verification: Verification): boolean {
+  validateVerificationBeforeStoring(verification: VerificationExport): boolean {
     if (
       verification.compilation.runtimeBytecode === undefined &&
       verification.compilation.creationBytecode === undefined
```
marcocastignoli, line 541: probably this is confusing, in here we should check that we should skip only if runtime and creation match are both `undefined`

marcocastignoli (approved):
lgtm

PR #2004, opened 2025-03-17
Deprecate v1 repository endpoints in openapi spec

manuelwedler:
Updates the API docs in order to encourage users to use v2 lookup.


marcocastignoli (approved):
lgtm

PR #2005, opened 2025-03-17
[Verification refactoring] Add missing tests in lib-sourcify

marcocastignoli:


packages/lib-sourcify/test/SourcifyChain.spec.ts
```diff
@@ -286,8 +286,16 @@ describe('SourcifyChain', () => {
 
   describe('Unit tests', () => {
     let hardhatNodeProcess: ChildProcess;
+    let sourcifyChain: SourcifyChain;
```
manuelwedler: Can we move this describe block out to the topmost level? I don't like the shadowing of the `sourcifyChain` variable.

manuelwedler (approved):
Looks good!

PR #2010, opened 2025-03-19
Update update lib sourcify docs

kuzdogan:
IMHO the previous documentation was too long and contain a lot of unnecessary information. 

I rewrote it by removing many parts and making it more concise. Let me know what you think.


marcocastignoli (approved):
I agree that this version feels less verbose while keeping the most important information

PR #2018, opened 2025-03-20
Upgrade contract private endpoint

### LINKED ISSUE #1974: Inconsistent `creation_match` values between `sourcify_matches` and `verified_contracts`
### Problem:

Currently, we have inconsistent data where `sourcify_matches.creation_match` is set to `'partial'` or `'perfect'`, but the corresponding linked entry in `verified_contracts.creation_match` remains `false`. This inconsistency should not occur, as whenever `sourcify_matches` indicates a `'partial'` or `'perfect'` match, the related `verified_contract` should always reflect this with `creation_match = true`.

#### Example:

To illustrate the issue, the following SQL query identifies an example:

```sql
SELECT *
FROM sourcify_matches sm
JOIN verified_contracts vc ON vc.id = sm.verified_contract_id
WHERE
    (sm.creation_match IS NOT NULL AND vc.creation_match = false)
    AND (vc.runtime_metadata_match IS NULL OR vc.runtime_metadata_match = false)
    AND (vc.creation_metadata_match IS NULL OR vc.creation_metadata_match = false)
ORDER BY sm.id DESC
LIMIT 1;
```

#### Impacted records:

- Approximately **300k** contracts with `creation_match = 'perfect'` are affected.
- Approximately **2.5 million** contracts with `creation_match = 'partial'` are affected.

### Root cause:

This issue arises during the upgrade of matches initially lacking creation information. Specifically:

1. `contract_deployments` is updated correctly with new creation information (without creating a new entry).
2. When there is a new `compiled_contracts` entry, updates occur successfully because the UNIQUE constraint in `verified_contracts` includes a distinct `compilation_id`, thus avoiding conflicts and ensuring consistency.
3. If there isn't a new `compiled_contracts` but better match data is obtained:
   - Existing entries in `verified_contracts` are not updated or replaced due to constraints (`deployment_id` and `compilation_id` uniqueness). Infact `deployment_id` remains unchanged (but gets updated), which prevents new entries from being created.
   - Consequently, `sourcify_matches` points to outdated entries in `verified_contracts`.

#### Recent upgrade
During the initial database creation, we encountered challenges obtaining complete creation information. To address this, we subsequently performed a second re-verification to populate the missing creation details in the production database. However, this re-verification process introduced the inconsistencies described above, affecting approximately 2.5 million contracts.

### Severity:

The issue is of low severity since the data in `sourcify_matches` remains accurate. The missing information in `verified_contracts` includes only `creation_match`, `creation_transformations`, and `creation_transformations_values`.

### Proposed code solution:

To resolve future inconsistencies:

- Update the UNIQUE constraint in `contract_deployments` to include the `contract_id`. We need to modify the code so that instead of updating the existing `contract_deployments`, a new entry is always created whenever new creation information becomes available.
- Consequently, creating a new entry rather than updating ensures a new `deployment_id`, which guarantees successful insertions into `verified_contracts` without triggering uniqueness constraints.

### Remediation of existing data:

Currently missing `creation_transformations` and `creation_transformations_values` can only be restored by re-verifying affected contracts using either a custom or existing re-verification script.

marcocastignoli:
This PR adds a new private endpoint that we'll use to fix #1974, this endpoint calls the `upgradeContract` handler that:
- extract compilation and onchain information from the database
- initialize `Compilation` and `Verification` with the extracted data, overriding getter functions
- disable unwanted `Compilation` methods, such as `Compilation.compile()` and `Compilation.generateCborAuxdataPositions()`, we already have in the database all the needed information
- mock `SourcifyChain` methods so that it returns onchain information without calling the RPC
- calls `Verification.verify()` to regenerate transformations
- updates misaligned columns in `verified_contracts` and `sourcify_matches`

### DIFF
--- a/services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
+++ b/services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
@@ -1,0 +1,349 @@
+import { Response } from "express";
+import { LegacyVerifyRequest, extractFiles } from "../../verification.common";
+import {
+  ISolidityCompiler,
+  SolidityMetadataContract,
+  createMetadataContractsFromFiles,
+  Verification,
+  SolidityCompilation,
+  SolidityJsonInput,
+  SourcifyChain,
+} from "@ethereum-sourcify/lib-sourcify";
+import { BadRequestError, NotFoundError } from "../../../../../common/errors";
+import { StatusCodes } from "http-status-codes";
+import { Services } from "../../../../services/services";
+import { ChainRepository } from "../../../../../sourcify-chain-repository";
+import logger from "../../../../../common/logger";
+import { getApiV1ResponseFromVerification } from "../../../controllers.common";
+
+export async function verifyDeprecated(
+  req: LegacyVerifyRequest,
+  res: Response,
+): Promise<any> {
+  const solc = req.app.get("solc") as ISolidityCompiler;
+  const services = req.app.get("services") as Services;
+  const chainRepository = req.app.get("chainRepository") as ChainRepository;
+
+  const inputFiles = extractFiles(req);
+  if (!inputFiles) {
+    const msg =
+      "Couldn't extract files from the request. Please make sure you have added files";
+    throw new NotFoundError(msg);
+  }
+
+  let metadataContracts: SolidityMetadataContract[];
+  try {
+    metadataContracts = await createMetadataContractsFromFiles(inputFiles);
+  } catch (error: any) {
+    throw new BadRequestError(error.message);
+  }
+
+  if (metadataContracts.length !== 1 && !req.body?.chosenContract) {
+    const contractNames = metadataContracts.map((c) => c.name).join(", ");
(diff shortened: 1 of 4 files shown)


services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -0,0 +1,392 @@
...
+        },
+      },
+    };
+
+    compilation["compile"] = async () => {
+      // Override so that it doesn't compile
+    };
+
+    compilation["generateCborAuxdataPositions"] = async () => {
+      // Override so that it doesn't generate auxdata positions
```
manuelwedler, line 295: Don't we need to generate these in order to have the `matchBytecodes` function working?

manuelwedler: Ah I see you mock the properties below. Nevermind then :)

manuelwedler:
I'm still not sure if we should run this inside the server. We could also just run this as a script in GCP.

marcocastignoli:
@kuzdogan can I have a feedback also from you? Also, what do you think of Manu's comment?

> I'm still not sure if we should run this inside the server. We could also just run this as a script in GCP.

marcocastignoli:
Closed in favor of #2042

PR #2020, opened 2025-03-20
Fix on conflict clause for insertContractDeployment

marcocastignoli:
The `ON CONFLICT (chain_id, address, transaction_hash, contract_id)` clause should always match the constraint stored in the database, I didn't know that you can actually use the different syntax `ON CONFLICT ON CONSTRAINT contract_deployments_pseudo_pkey` instead of listing the columns.


manuelwedler (commented):
Looks good!

Can we maybe also add the tests for VerA here? Or do you think it's too much effort?

PR #2027, opened 2025-03-20
Improve express error handler

manuelwedler:
Rename `code` of `IResponseError` to `statusCode`
Needed to avoid interfering with postgres errors that also have a `code` property.

Adds logging for all unknown errors.

Also renames deprecated `getStatusText` method of HttpStatus to `getReasonPhrase`

### DIFF
--- a/services/server/src/common/errors/GenericErrorHandler.ts
+++ b/services/server/src/common/errors/GenericErrorHandler.ts
@@ -1,5 +1,6 @@
 import * as HttpStatus from "http-status-codes";
 import { Request, Response } from "express";
+import logger from "../logger";
 
 export default function genericErrorHandler(
   err: any,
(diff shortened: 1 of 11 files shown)


services/server/src/common/errors/GenericErrorHandler.ts
```diff
@@ -9,7 +10,10 @@ export default function genericErrorHandler(
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   _next: any,
 ): void {
-  const errorCode = +err.code || err.status || 500;
+  const errorCode = +err.statusCode || err.status || 500;
+  if (errorCode === 500) {
+    logger.error("Unexpected server error", { error: err });
+  }
```
marcocastignoli, line 16: If we add this logger here, then we have to remove the log from any other place in which we throw I suppose?

marcocastignoli (changes requested):

services/server/src/common/errors/GenericErrorHandler.ts
manuelwedler: I moved it mainly here, because we had a log in the `safeHandler` which is removed now. So before removing it, I think there were already duplicate logs. 

Imo, the duplicate logs are no problem, because in the best case we should never reach this point. It should rather be a way to remind us that we have to handle something differently or fix something.

marcocastignoli (approved):
lgtm

PR #2040, opened 2025-03-26
Add POST `/v2/verify/{chainId}/{address}` endpoint

### LINKED ISSUE #1910: Implement POST `/v2/verify/{chainId}/{address}` endpoint

manuelwedler:
Closes #1910 

As specified in https://sourcify.stoplight.io/docs/sourcify-apiv2/h55xvo5bsu9qz-verify-contract-standard-json

Includes the needed verification worker which is implemented with the Piscina library.

See commit messages for details.

We should still do some testing on the worker configuration (number of threads, ...) and some benchmarking after this.

### DIFF
--- a/services/server/src/server/services/workers/workerTypes.ts
+++ b/services/server/src/server/services/workers/workerTypes.ts
@@ -15,6 +15,12 @@ export interface VerifyFromJsonInputs {
   creationTransactionHash?: string;
 }
 
+// We need to return an object from the worker for being able to return an
+// error with custom properties. Structured clone only copies the default
+// properties (name and message) of Error objects when they are thrown, but
+// we want to return our custom `MatchingErrorResponse`.
+// Any instance of the `VerifyFromJsonOutput` interface should only contain
+// one of the two properties.
 export interface VerifyFromJsonOutput {
   verificationExport?: VerificationExport;
   errorResponse?: MatchingErrorResponse;


manuelwedler:
Some explanations: 
- The `unknown_error` was renamed to `internal_error` to match the specification.
- I needed to make some changes to the `SourcifyChainMap` in order to pass its properties to the workers. That is why also some changes to the monitor are included.

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  // The contract path can include a colon itself,
+  // therefore we need to take the last element as the contract name
+  const splitIdentifier = contractIdentifier.split(":");
+  const contractName = splitIdentifier[splitIdentifier.length - 1];
+  const contractPath = splitIdentifier.slice(0, -1).join(":");
+  const compilationTarget: CompilationTarget = {
+    name: contractName,
+    path: contractPath,
+  };
```
marcocastignoli, line 74: Can we actually handle this case from outside the worker? It feels to me that having clear inputs for the worker is somehow better

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -0,0 +1,158 @@
...
+    await verification.verify();
+  } catch (error: any) {
+    return {
+      errorResponse: createMatchingError(error, verification),
+    };
+  }
+
+  return {
+    verificationExport: verification.export(),
+  };
```
marcocastignoli, line 133: Can we pass directly the export? Why do we need an object?

marcocastignoli (approved):
Great job! 🎉 I just added a few nitpicky comments that don't even have to be addressed. Just implement them if you also think they make sense.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -825,17 +897,6 @@ export class SourcifyDatabaseService
...
-      // it means that the new match wasn't better (perfect > partial) than the existing one
-      if (verifiedContractId === false) {
-        logger.info("Not Updated in SourcifyDatabase", {
-          address: verification.address,
-          chainId: verification.chainId,
-          runtimeMatch: verification.status.runtimeMatch,
-          creationMatch: verification.status.creationMatch,
-        });
-        return;
-      }
```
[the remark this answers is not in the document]
manuelwedler: This is never run, because `verifiedContractId` can never be false. This seems to be some leftover, because we check if it is a better match earlier in `StorageService`.

services/server/src/server/services/workers/verificationWorker.ts
manuelwedler: Yes, I think that makes sense.

services/server/src/server/services/workers/verificationWorker.ts
manuelwedler: I was actually doing this earlier, but realized then that we also need to return errors in the response. Potentially, you could just throw errors, but the structured clone algorithm would only copy the default properties (name and message) of the Error class back to the main thread. We also need to pass our custom code and bytecodes back. So I needed the response object to be able to also return an error (instead of throwing).

marcocastignoli: Ok that's clear, maybe let's add a comment to explain that

PR #2042, opened 2025-03-31
Upgrade contract private endpoint

marcocastignoli:


services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -0,0 +1,406 @@
...
+        sourcifyMatchId,
+      ],
+    );
+
+    // 2. Update sourcify_matches table using direct query instead of updateSourcifyMatch
+    await poolClient.query(
+      `UPDATE sourcify_matches SET 
+        creation_match = $1
+      WHERE verified_contract_id = $2`,
+      [verificationStatus.creationMatch, sourcifyMatchId],
```
manuelwedler, line 380: here it should be `WHERE id = $2`

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -0,0 +1,406 @@
...
+        creation_transformations = $2,
+        creation_values = $3,
+        creation_metadata_match = $4
+      WHERE id = $5`,
+      [
+        creationMatch,
+        creationTransformations,
+        creationValues,
+        creationMetadataMatch,
+        sourcifyMatchId,
```
manuelwedler, line 371: This should be the id of the verified contract, not `sourcifyMatchId`

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -0,0 +1,406 @@
...
+      [verificationStatus.creationMatch, sourcifyMatchId],
+    );
+
+    // Commit the transaction
+    await poolClient.query("COMMIT");
+
+    res.send({
+      result: {
+        message: "Contract upgrade successful",
+        verifiedContractId: sourcifyMatchId,
```
manuelwedler, line 389: wrong id here too

manuelwedler (changes requested):
Found some issues about the use of the ids. Other than that, this should fix our data :)

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
marcocastignoli: Amazing catch 🙏 In my test I had those two ids with the same values and I didn't catch it

services/server/src/openapi.yaml
```diff
@@ -18,6 +18,10 @@ tags:
   - name: Session Verification
     description: API v1 - Used by the Sourcify UI to verify contracts. Not intended for external usage. Will require CORS.
 paths:
+  /private/verify-deprecated:
+    $ref: "server/apiv1/verification/private/stateless/private.stateless.paths.yaml#/paths/~1private~1verify-deprecated"
+  /private/upgrade-contract:
+    $ref: "server/apiv1/verification/private/stateless/private.stateless.paths.yaml#/paths/~1private~1upgrade-contract"
```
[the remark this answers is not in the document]
marcocastignoli: I just realized we don't put any real restriction on them :D You are totally right, they should be behind some private token of some sort

manuelwedler (approved):
Looks good!

kuzdogan:
Do we need to publish this endpoint? I thought we will keep this internal.

marcocastignoli:
> Do we need to publish this endpoint? I thought we will keep this internal.

In order to call it we either include it in sourcify-server-production or we deploy an ad-hoc service just to call this. From my pov if the /private endpoint is protected by an Authorization method, we can just merge it into master and keeping it active until we need it, then we can disable it in the settings

kuzdogan:
It's also okay this way and I'm not strongly against it but IMO it bloats the codebase unnecessarily and might confuse people, considering this is something only needed now and not any time in the future.

marcocastignoli:
> might confuse people, considering this is something only needed now and not any time in the future.

I see your point, on the other hand it's the easiest solution to run it. Also, from the user prospective, it's clear from my pov that those `/private` endpoints are only used by us; and we can always remove the code when we are done.

manuelwedler:
This should definitely be removed from the codebase once the data is fixed.

PR #2045, opened 2025-04-01
Add POST `/v2/verify/metadata/{chainId}/{address}` endpoint

### LINKED ISSUE #1911: Implement POST `/v2/verify/metadata/{chainId}/{address}` endpoint
New verify endpoint using Solidity metadata.json as input

Specification in [Stoplight](https://sourcify.stoplight.io/docs/sourcify-apiv2/mry40ifs9wvb7-verify-contract-using-solidity-metadata-json)

manuelwedler:
Closes #1911 

As specified in https://sourcify.stoplight.io/docs/sourcify-apiv2/branches/main/mry40ifs9wvb7-verify-contract-using-solidity-metadata-json


manuelwedler:
Finally, this is ready. Took some time to understand what was wrong, because it was only failing in CI initially. Reason is that locally we don't run unit and integration tests in the same run.

packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
```diff
@@ -261,11 +268,12 @@ export class SolidityMetadataContract {
     };
 
     if (!compilationTarget || Object.keys(compilationTarget).length != 1) {
-      throw new Error(
+      logDebug(
         `Can't create JsonInput from metadata: Invalid compilationTarget in metadata: ${Object.keys(
           this.metadata.settings.compilationTarget,
         ).join(',')}`,
       );
```
marcocastignoli, line 275: it's better to always log using the object to log logger's metadata. In this case, something like this:

```ts
logDebug(
        `Can't create JsonInput from metadata: Invalid compilationTarget in metadata`,
        {
              compilationTargets: Object.keys(
                 this.metadata.settings.compilationTarget,
              )
        }
);
```

marcocastignoli: (It wasn't like this before because it was passed directly to the user.

packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
```diff
@@ -261,11 +268,12 @@ export class SolidityMetadataContract {
...
+      throw new ValidationError('invalid_compilation_target');
```
marcocastignoli, line 276: Related to the comment above, here we must pass the `compilationTargets` to the user

packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
```diff
@@ -236,11 +242,12 @@ export class SolidityMetadataContract {
...
     ) {
-      throw new Error(
-        `Can't create JsonInput from metadata: Missing or invalid sources in metadata: ${JSON.stringify(
+      logDebug(
+        `Can't create JsonInput from metadata: Missing or invalid sources in metadata: missing:${JSON.stringify(
           this.missingSources,
-        )}`,
+        )}; invalid: ${JSON.stringify(this.invalidSources)}`,
       );
+      throw new ValidationError('missing_or_invalid_source');
```
marcocastignoli, line 250: Are we passing the missing sources to the user? Idk what we used to pass before but it feels like it's an important info for the user (missing keccak / missing path)

packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
```diff
@@ -222,10 +226,12 @@ export class SolidityMetadataContract {
...
     if (Object.keys(this.missingSources).length) {
-      const error = new Error(
-        `Resource missing; unsuccessful fetching: ${Object.keys(this.missingSources).join(', ')}`,
+      logDebug(
+        `Resource missing; unsuccessful fetching: ${Object.keys(
+          this.missingSources,
+        ).join(', ')}`,
       );
-      throw error;
+      throw new ValidationError('missing_source');
```
marcocastignoli, line 234: Same problem here of missing information

marcocastignoli (changes requested):
I just have the following doubt regarding error handling

manuelwedler:
One question from my side: 

The tests were failing because the metadata object is modified inside `SolidityMetadataContract` and we pass the same testing metadata object around. This means properties of this object are modified unexpectedly.

At the moment, we pass standard json input or metadata as objects to lib-sourcify and modify (nested) properties of these objects. I think this can lead to confusion for users of the libraries. In the previous lib-sourcify version, we were cloning the metadata object. I wonder if we should do the same again. What do you think @marcocastignoli ?

packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
manuelwedler: I fixed this by passing the compilationTargets to the error message. Users of lib-sourcify will see this error message.

However, when you would use the job lookup endpoint, you wouldn't see them. For that, we would need to store the compilationTargets in the database (maybe in `verification_jobs_ephemeral`). Do you think that makes sense?

marcocastignoli: In `verification_jobs_ephemeral` we already store the error messages, right?

manuelwedler: No, we don't store the error message. We just store the code, and construct the message with the getter from the code

marcocastignoli:
Yeah it makes sense to clone the metadata object in the constructor of `SolidityMetadataContract`

manuelwedler:
> Yeah it makes sense to clone the metadata object in the constructor of `SolidityMetadataContract`

Okay, will make that change here then.

packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
manuelwedler: Same issue here. We don't store the error message in the db.

packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
manuelwedler: Imo, it's better to store these additional fields in the database than storing the messages themselves. We could add a json column `error_payload`, which contains just the neeeded fields for reconstructing the error messages.

marcocastignoli: let's also have an opinion by @kuzdogan  since this would also affect the API design. Do we need to update the GET verificationJob result?

manuelwedler: We could just include these fields inside the error message, then no API change would be needed.

But we could also return that `errorPayload` field as a json object in the API. This would make it also machine-readable.

marcocastignoli: > Probably the latter is even better considering that we could use it in the UI.

That's exactly what I'm thinking of, if from the UI we need to show which files are invalid or the keccak it's better to have those information available separately from the error message

manuelwedler: I decided with Marco to implement my proposal. I named the new column `error_data`. Now, we have customized error messages dependent on parameters without storing the error messages themselves in the database.

@marcocastignoli you can have another look.

marcocastignoli (approved):
lgtm!

PR #2055, opened 2025-04-08
Delete h5ai-nginx

kuzdogan:


manuelwedler (approved):

manuelwedler (changes requested):
Wait, I think we also need to remove it from `.gitmodules`.

There are also CI configurations for building it.

manuelwedler (approved):
:+1:

PR #1381, opened 2024-04-30
Add New Chains Stratis 105105 205205

zeptin:
# Add New Chain 105105, 205205

Thanks for your pull request to add a new support in Sourcify.

If you haven't done so, please follow the instructions on [how to request chain support](https://docs.sourcify.dev/docs/chain-support/) in docs.

Please check the following items before submitting your pull request for a speedy review.

## Checklist

- [x] The branch is named as `add-chain-<chainId>`.
- [x] I haven't modified the `chains.json` file directly.
- [x] In `sourcify-chains.json` file
  - [x] I've set `supported: true`.
  - [x] I haven't added an `rpc` field but the one in [chains.json](../../src/chains.json) is used (if not, please explain why).
- [x] I've added a test in [chain-tests.js](../../test/chains/chains-test.js) file.
- [ ] `test-new-chain` test in Circle CI is passing.

PR #2057, opened 2025-04-08
Update verifyContract for Auroria (chain 205205)

zeptin:
The Auroria testnet was restarted and this contract had to be re-deployed.

See #1381 and #2056


kuzdogan (approved):
I've manually ran the tests and it's working. Thanks. Next time please name the branch as `add-chain-205205`

Just, is the testnet expected to reset regularly? If so please update the contracts here.

zeptin:
No, further resets are not likely

PR #2058, opened 2025-04-08
Remove environment variables  from lib-sourcify

marcocastignoli:
In this PR I'm removing environment variables from `lib-sourcify`. 
- We already have `setLibSourcifyLoggerLevel` to change the log level of lib-sourcify
- I added `SourcifyChain.setGlobalRpcTimeout` to change the rpc timeout in `SourcifyChain`
- I added `SolidityMetadataContract.setGlobalIpfsGateway` to change the `IpfsGateway`. I decided to not create a custom function outside of `SolidityMetadataContract` because that's the only place in which we use the `IpfsGateway`
- The `LibSourcify.SourcifyChainsExtensionsObject` doesn't have anymore a `headerEnvName` parameter. Now it has only a `headerValue` parameters. Consequently I had to create a custom `SourcifyServer.SourcifyChainsExtensionsObjectWithHeaderEnvName`, that includes `headerEnvName`, and at server's boot it loads the environment variables values inside `headerValue`

**TODO**
- [x] Tests passing
- [x] Update docs

### DIFF
--- a/packages/lib-sourcify/src/logger.ts
+++ b/packages/lib-sourcify/src/logger.ts
@@ -6,7 +6,7 @@ export interface ILogger {
 
 // Default logger behavior
 export const DefaultLogger: ILogger = {
-  logLevel: process.env.NODE_ENV === 'production' ? 2 : 4,
+  logLevel: 2,
   setLevel(level: number) {
     this.logLevel = level;
   },
--- a/packages/lib-sourcify/src/logger.ts
+++ b/packages/lib-sourcify/src/logger.ts
@@ -60,6 +60,10 @@ export function setLogger(logger: ILogger) {
   AppLogger = logger;
 }
 
+export function getLevel(): number {
+  return AppLogger.logLevel;
+}
+
 export function setLevel(level: number) {
   AppLogger.setLevel(level);
 }
--- a/services/server/src/server/cli.ts
+++ b/services/server/src/server/cli.ts
@@ -40,20 +40,17 @@ if (process.env.IPFS_GATEWAY || process.env.IPFS_GATEWAY_HEADERS) {
 if (process.env.RPC_TIMEOUT) {
   try {
     libSourcifyConfig.rpcTimeout = parseInt(process.env.RPC_TIMEOUT);
-    logger.info("lib-sourcify RPC timeout set", {
-      rpcTimeout: process.env.RPC_TIMEOUT,
-    });
   } catch (error) {
     logger.error("Error setting lib-sourcify RPC timeout", { error });
     throw new Error("Error setting lib-sourcify RPC timeout");
   }
 }
 
-if (process.env.NODE_ENV !== "production") {
-  // Set the log level to 4 (debug) in non-production environments
-  libSourcifyConfig.logLevel = 4;
-  logger.info("lib-sourcify log level set to debug");
-}
+// This variable is used to set the log level for the server and lib-sourcify
+const logLevel = parseInt(
+  process.env.NODE_LOG_LEVEL ||
+    (process.env.NODE_ENV === "production" ? "1" : "5"),
+);
 
 // Solidity Compiler
 
@@ -92,6 +89,7 @@ const server = new Server(
     upgradeContract: config.get("upgradeContract"),
     sessionOptions: getSessionOptions(),
     sourcifyPrivateToken: process.env.SOURCIFY_PRIVATE_TOKEN,
+    logLevel,
     libSourcifyConfig,
   },
   {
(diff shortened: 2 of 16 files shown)


services/server/src/server/cli.ts
```diff
@@ -23,6 +23,46 @@ import { SolcLocal } from "./services/compiler/local/SolcLocal";
...
+    logger.error("Error setting lib-sourcify RPC timeout", { error });
+    throw new Error("Error setting lib-sourcify RPC timeout");
+  }
+}
+
+if (process.env.NODE_ENV !== "production") {
+  // Set the log level to 4 (debug) in non-production environments
+  setLibSourcifyLoggerLevel(4);
+  logger.info("lib-sourcify log level set to debug");
+}
```
marcocastignoli, line 65: This is how you actually set lib-sourcify's configs

kuzdogan (changes requested):
One question regarding setting the logging level of lib-sourcify. Also see my video sent about the ExtendedSourcifyChainsObject etc.

packages/lib-sourcify/src/logger.ts
```diff
@@ -6,7 +6,7 @@ export interface ILogger {
 
 // Default logger behavior
 export const DefaultLogger: ILogger = {
-  logLevel: process.env.NODE_ENV === 'production' ? 2 : 4,
+  logLevel: 2,
```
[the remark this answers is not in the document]
marcocastignoli: This is the point where I set it: https://github.com/ethereum/sourcify/pull/2058/files/27827cd4f4ebeecc44644d035c9aa8f5d1fc466c#diff-06c65ad1bf1384622366b536a841a7d2dd1c9f722e1bb11b3110ca13a94f5583R61-R65

services/server/src/server/cli.ts
marcocastignoli: I didn't think about this! It's actually a good idea, let's pass them as Server's parameter!

marcocastignoli: Notice that this variable should still be set in `cli.ts` because that's where we extract all the configs / env variables in order to pass them to `Server`.

kuzdogan: wdym? It's only dependent on the `NODE_ENV` variable. We can just look at that in Server and only set there. `cli` calls Server anyway. Why is it needed to be set in `cli`?

services/server/src/sourcify-chains.ts
```diff
@@ -18,7 +18,30 @@ import dotenv from "dotenv";
...
+  [chainId: string]: {
+    sourcifyName: string;
+    supported: boolean;
+    etherscanApi?: {
+      apiURL: string;
+      apiKeyEnvName?: string;
+    };
+    fetchContractCreationTxUsing?: any; // Using any to avoid importing the full type
+    rpc?: Array<string | BaseRPC | APIKeyRPC | ExtendedFetchRequestRPC>;
+  };
```
manuelwedler, line 41: You don't need to redeclare the whole type here. You can just overwrite the `rpc` property of `SourcifyChainExtension`.

services/server/src/server/cli.ts
manuelwedler: I think it is weird that the log level is not consistent between lib-sourcify and server. Can we have one server constructor parameter that sets log level on server and lib-sourcify (and all other packages)?

Inside cli.ts we can set the value of this parameter from process.env.
Let's use this statement for setting it: 
```
const level = process.env.NODE_LOG_LEVEL || (process.env.NODE_ENV === "production" ? "info" : "debug");
 ```

File: services/server/src/server/server.ts [hunk 4]
```diff
@@ -62,10 +80,35 @@ export class Server {
     verificationServiceOptions: VerificationServiceOptions,
     storageServiceOptions: StorageServiceOptions,
   ) {
+    if (options.logLevel) {
+      if (options.logLevel && !validLogLevels.includes(options.logLevel)) {
```
manuelwedler, line 84:
```suggestion
    if (options.logLevel !== undefined) {
      if (!validLogLevels.includes(options.logLevel)) {
```

the logLevel can be 0

services/server/src/server/server.ts
```diff
@@ -76,6 +80,14 @@ export class Server {
...
+        throw new Error(`Invalid log level: ${options.logLevel}`);
+      }
+      setLogLevel(LogLevels[options.logLevel]);
+      setLibSourcifyLoggerLevel(options.logLevel);
```
manuelwedler, line 88: Can you also set the log level of compilers package here?

manuelwedler (changes requested):

manuelwedler (approved):
looks good!

PR #2059, opened 2025-04-08
Update Hoodi Testnet with RPCs and creatorTxHash

kuzdogan:


services/server/src/sourcify-chains-default.json
```diff
@@ -127,7 +127,30 @@
...
+      "blockscoutApi": {
+        "url": "https://eth-hoodi.blockscout.com/"
+      },
+      "etherscanApi": true,
+      "routescanApi": {
+        "type": "testnet"
+      }
+    },
+    "rpc": [
+      "https://rpc.hoodi.ethpandaops.io",
```
manuelwedler, line 145: Doesn't this need authorization headers?

kuzdogan: Nope, Hoodi and Holesky RPCs are open. Only mainnet is with Auth

manuelwedler: Alright, just wanted to make sure :)

PR #2060, opened 2025-04-08
Add `sourceIds` field to v2 lookup

### LINKED ISSUE #2007: Add `sourceIds` field to v2 lookup
Following a suggestion on Matrix chat, we should add a field `sourceIds` on the highest level of the response of the `v2/contract/:chainId/:address` endpoint. It should contain the source file ids as in the standard json output. It should return a mapping of type `Record<string, { id: number }>`, where the keys are the file paths.

It should be optional and only be returned if queried via `?fields`.

manuelwedler:
Closes #2007

### DIFF
--- a/packages/lib-sourcify/src/Verification/VerificationTypes.ts
+++ b/packages/lib-sourcify/src/Verification/VerificationTypes.ts
@@ -14,6 +14,8 @@ import {
   Metadata,
   Userdoc,
   VyperSettings,
+  SolidityOutputSource,
+  VyperOutputSource,
 } from '@ethereum-sourcify/compilers-types';
 import { ErrorMessagePayload, SourcifyLibError } from '../SourcifyLibError';
 import { Transformation, TransformationValues } from './Transformations';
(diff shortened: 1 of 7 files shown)


packages/lib-sourcify/src/Verification/VerificationTypes.ts
```diff
@@ -95,7 +97,10 @@ export interface VerificationExport {
     sources: StringMap;
     compilerOutput: {
       // The export should not include the AST object to reduce the size
-      sources?: Record<string, { id: number }>;
+      sources?: Record<
+        string,
+        Pick<SolidityOutputSource, 'id'> | Pick<VyperOutputSource, 'id'>
```
marcocastignoli, line 102: Does it make sense from a type point of view to say number or number? It seems like it's always number.

marcocastignoli (approved):
lgtml. Just a nitpick comment, feel free to not address it.

packages/lib-sourcify/src/Verification/VerificationTypes.ts
manuelwedler: Do you mean to just use `Record<string, {id: number}>`?

PR #2069, opened 2025-04-09
Revert "Disable v2 verification and job status endpoints and do not run Piscina workers (#2061)"

manuelwedler:
Enabling this again because the release is done


marcocastignoli (approved):
👍

Issue #1877, opened 2025-01-29
Test asynchronous verification with GCP downscaling


manuelwedler:
With the new server architecture in API v2, we will use multithreading for background verification. We found that potentially GCP Cloud Run could kill a server instance that has no open requests but runs a verification in the background (https://github.com/ethereum/sourcify/issues/1830#issuecomment-2621148018). 

To make sure API v2 works as expected, we should trigger the downscaling on staging GCP deployment and see if verification still succeeds when instances are killed, or if jobs are kept in pending state in the db and we cannot recover.

Since we need to deploy API v2 to GCP for testing this, this issue is a last step of the API v2 milestone before we can publish to master.

manuelwedler:
To tackle this, a script would be good that does the following:

- compile a complicated contract (compilation time should be long such that the worker will run for longer)
- deploy the contract a couple of times (maybe 100 times if possible)
- send a verification request for each deployed contract
- output the verification_ids

manuelwedler:
I created such script as described before and did 4 test runs with different scenarios. The tests were conducted directly on our staging server deployment. Here are the results from my investigations.

From debugging a previous issue (https://github.com/ethereum/sourcify/issues/1531), I still had the standard json input for a contract that takes very long to compile. It takes around 103 seconds to compile it locally on my machine.

The script I used for testing, compiles this contract, deploys it 10 times and sends a verification request for each deployment. It can be configured to send the compilation requests concurrently, or to wait 10 seconds between the requests.

The script: https://gist.github.com/manuelwedler/f756ffe779d4f2ea25cbffe2ca2351ea#file-scalingtest-ts
The standard json input: https://gist.github.com/manuelwedler/f756ffe779d4f2ea25cbffe2ca2351ea#file-input-json

All tests were run on GCP Cloud Run. The container was configured to use 4 GB memory and 1 vCPU. This is the configuration we use currently for both deployments, production and staging. We should change the number of CPUs later in order to take effect of the worker thread approach on performance, but the downscaling effects are more easily tested with 1 CPU.

To evaluate the results, I took screenshots from our GCP metrics and noted the contract compilation times which are stored in the database. The good thing about staging is that the traffic is comparably low and the server instance count is usually just 1. Due to that, we can see the effects of my simple script very clearly in the GCP metrics. Between the tests, I always waited until the instance count was 1 again.

## Test 1: concurrent requests

[...]

PR #2070, opened 2025-04-09
Fix SIGTERM handling of server containers

manuelwedler:
I found why our server does not handle SIGTERM signals in GCP (#1877). The issue is our Dockerfile which used `npm start` to run the server. Due to this, `node` does not have PID 1 which is needed to properly handle signals from the outside in the container (see https://www.docker.com/blog/docker-best-practices-choosing-between-run-cmd-and-entrypoint/).

I created a docker compose file for locally testing if the SIGTERM is handled. The compose file makes it much easier to run the server with the db through Docker locally. I thought it would be good to also add it to the repo.

### DIFF
--- a/services/database/database.json
+++ b/services/database/database.json
@@ -15,6 +15,14 @@
     "database": "sourcify",
     "port": { "ENV": "DOCKER_HOST_POSTGRES_TEST_PORT" }
   },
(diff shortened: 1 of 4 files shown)


marcocastignoli (approved):
lgtm

PR #2094, opened 2025-04-19
copy `fetchContractCreationTxUsing` for an unknown chain

daniil-parsiq:
Currently, when you use custom `sourcify-chains.json` there is a problem for networks that are not present `chains.json`, it is then categorized as `unknownChain` and not all of the properties are being copied when populating `sourcifyChainsMap`.

I've added `fetchContractCreationTxUsing` to prevent contract creation transaction being fetched only through binary search wtih RPC.


kuzdogan (approved):
Good catch, thanks!

kuzdogan:
Tests are failing because of a timeout but it should not be related to this so I'll merge.

Issue #2071, opened 2025-04-09
Test performance of verification via worker


manuelwedler:
In #1877, we tested if GCP scaling works with our asynchronous verification worker approach of API v2. Since this is fixed, we should start to evaluate what properties make the worker performant.

For this we can reuse the script from #1877 and see how the server behaves in different scenarios. I think both scenarios from #1877 are relevant (concurrent requests, and waiting between requests). ~~Additionally to the complex contract, we should try to verify a very simple contract with a lot of more requests (long compilation vs. a lot of short compilations).~~ Edit: I decided to go with just one medium complex contract, because I think it is better to test with just one average contract. From our logs I can see that the median compilation time is roughly 3 - 5 seconds.

As a result of this issue, we should decide on values for the properties below which optimize the tested metrics for us.

### Metrics for evaluation

We should get some numbers on the following metrics to evaluate performance vs. costs.

- ~~Compilation times~~
- Job run times: Instead of compilation times, the total time a job took is more interesting for these tests because it includes worker start up times.
- Instance count and billable container instance time
- CPU and memory utilization

### Properties to tweak

We should run the test scenarios with different values for these properties:

- GCP container CPU setting
- GCP container memory setting
- Piscina's `concurrentTasksPerWorker` option
- Piscina's `idleTimeout` option

The properties with the biggest effects are the GCP container settings. I expect the Piscina options to lead to more fine-grained changes in the result. Maybe it's worth to first evaluate the Piscina options with a single CPU as their effects will be bigger then.

####  Some thoughts about the `concurrentTasksPerWorker` option

To quote the Piscina docs:

> This generally only makes sense to specify if there is some kind of asynchronous component to the task. Keep in mind that Worker threads are generally not built for handling I/O in parallel.

The compiler is run as a subprocess. That's why the compilation is parallelized anyway and our worker setup mainly tries to scale I/O. Therefore, I think how we designed it is questionable in general (but that just as a side note). We should try a few values for this option.

manuelwedler:
## TLDR

I will configure staging and production Sourcify servers to use the following values for the parameters:

- `idleTimeout`: 30 second
- `concurrentTasksPerWorker`: 10 (for now, maybe reevaluate later)
- Number of CPUs: 2
- Memory: 4 GB

## Worker configuration tests

As stated above, at first, I tested the worker parameters with the current GCP configuration.
I chose a contract from current staging Sourcify that took 4853 seconds to compile there.

[...]

PR #2101, opened 2025-04-22
Make Piscina worker parameters configurable through environment variables

manuelwedler:
For #2071, it would be good to have these parameters configurable.


marcocastignoli (approved):
lgtm

PR #2112, opened 2025-04-24
Set logLevel and inject traceIds in verificationWorker threads

### LINKED ISSUE #2074: Inject `traceId` into the loggers of the packages
Currently, only server logs have the `traceId` injected for GCP. I wanted to follow the logs of a request starting from a compilers log, which is not easy without the `traceId`.  We should also inject it into the loggers of ~~lib-sourcify~~ and compilers packages.
Edit: This is only a problem of compilers package.

manuelwedler:
Closes #2074

### DIFF
--- a/services/server/src/server/services/workers/verificationWorker.ts
+++ b/services/server/src/server/services/workers/verificationWorker.ts
@@ -24,14 +24,16 @@ import type {
   VerifyFromJsonInput,
   VerifyFromMetadataInput,
   VerifyOutput,
+  VerificationWorkerInput,
 } from "./workerTypes";
-import logger from "../../../common/logger";
+import logger, { setLogLevel } from "../../../common/logger";
 import {
   ProcessedEtherscanResult,
   isVyperResult,
   processSolidityResultFromEtherscan,
   processVyperResultFromEtherscan,
 } from "../utils/etherscan-util";
+import { asyncLocalStorage } from "../../../common/async-context";
 
 export const filename = resolve(__filename);
 
@@ -43,6 +45,8 @@ const initWorker = () => {
   if (chainRepository && solc && vyper) {
     return;
   }
+
+  setLogLevel(Piscina.workerData.logLevel || "info");
 
   const sourcifyChainInstanceMap = Piscina.workerData
     .sourcifyChainInstanceMap as { [chainId: string]: SourcifyChainInstance };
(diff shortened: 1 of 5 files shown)


services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -63,16 +67,41 @@ const initWorker = () => {
   vyper = new VyperLocal(Piscina.workerData.vyperRepoPath);
 };
 
-export async function verifyFromJsonInput({
+async function runWorkerFunction<T extends VerificationWorkerInput>(
```
kuzdogan, line 70: Some descriptive naming would be helpful, I had to dig in what this does
```suggestion
async function runWorkerFunctionWithContext<T extends VerificationWorkerInput>(
```

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -63,16 +67,41 @@ const initWorker = () => {
...
+  workerFunction: (input: T) => Promise<VerifyOutput>,
+  input: T,
+): Promise<VerifyOutput> {
+  initWorker();
+  const context = { traceId: input.traceId };
+  return asyncLocalStorage.run(context, workerFunction, input);
```
kuzdogan, line 76: I see this now runs the worker funciton with this context. But I couldn't see where 
1. This context is being used 
2. Where the traceId is being logged.

Doesn't the `traceId` need to be within these fields? Maybe TS is missing this because it's run with `asynclocalstorage.run()`?
https://github.com/ethereum/sourcify/blob/8eba637e823cd955ff04f9f7bb6c9fad822f7310/services/server/src/server/services/workers/verificationWorker.ts#L97-L105

From what I can see the traceId is passed at VerificationService in the `input` when calling each worker but I don't see it anywhere after that.

kuzdogan (commented):
Some questions about there `traceId` is actually logged. You can also send me a Loom video or do a short call if it's easier to explain than write here.

Additionally, we have unused functions in server.ts `getIp()` and `hash()`

services/server/src/server/services/workers/verificationWorker.ts
manuelwedler: Our logger module gets the trace automatically:

https://github.com/ethereum/sourcify/blob/8eba637e823cd955ff04f9f7bb6c9fad822f7310/services/server/src/common/logger.ts#L71-L75

Since the worker is running in it's own thread, we need to set it inside the worker again (server sets it per request for the main thread).

The logging happens here:

https://github.com/ethereum/sourcify/blob/8eba637e823cd955ff04f9f7bb6c9fad822f7310/services/server/src/common/logger.ts#L100-L110

kuzdogan (approved):

PR #2124, opened 2025-04-29
Cleanup openapi.yaml

kuzdogan:
- Moved description etc. to the main openapi.yaml instead of v2 a3563b5b77c7b66c36eee6e2aa48b698648f78c5
- Converted "examples:" fields with one example to "example:" to show on swagger ui 51e54a0b527b8414e8d1297642c7c10d6dd77e84 SwaggerUI wasn't able to parse example arrays and show. Now we show examples values in path paramter {chainId} for example

### DIFF
--- a/services/server/src/openapi.yaml
+++ b/services/server/src/openapi.yaml
@@ -1,13 +1,30 @@
 openapi: 3.1.0
 info:
-  version: 1.0.0
-  title: Sourcify API
-  description: API to interact with Sourcify
+  version: 2.0.0
+  title: Sourcify APIv2
+  description: |-
+    Welcome to the Sourcify's APIv2.
+
+    Important differences between the deprecated [legacy API](https://docs.sourcify.dev/docs/api/) and the new APIv2:
+    - **Ticketing**: The verfication requests resolve into tickets/verification jobs. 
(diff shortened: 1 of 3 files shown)


services/server/src/openapi.yaml
```diff
@@ -1,13 +1,30 @@
...
+    url: 'https://sourcify.dev'
+    email: hello@sourcify.dev
 servers:
-  $ref: "servers.yaml"
+  - url: https://sourcify.dev/server
+    description: Production server
+  - url: https://staging.sourcify.dev/server
+    description: Staging server
+  - url: http://localhost:5555
+    description: Local development server address on default port 5555.
```
marcocastignoli, line 27: We can still use the `servers.yaml` file or remove it if we don't need it anymore

marcocastignoli (changes requested):

services/server/src/openapi.yaml
kuzdogan: I thought I deleted it, thanks

marcocastignoli (approved):

kuzdogan:
Merging, a single flaky test is failing with timeout.

PR #2131, opened 2025-04-30
Skip Etherscan tests for external PRs

### LINKED ISSUE #2122: Disable v2 Etherscan tests for external PRs
As seen on https://github.com/ethereum/sourcify/pull/2116 the test run is failing because we are not skipping Etherscan test runs on external PRs. Since they require API keys and env vars are not available for external PRs they fail: https://app.circleci.com/pipelines/github/ethereum/sourcify/8597/workflows/a1451d5f-e052-4edc-92be-6c093c029855/jobs/50829

We should skip them similar to how we do in v1

https://github.com/ethereum/sourcify/blob/6cca91c9d8e44562e5a4ab05e9c83229189d4821/services/server/test/integration/apiv1/verification-handlers/etherscan.spec.ts#L38-L43

manuelwedler:
closes #2122

### DIFF
--- a/services/server/test/integration/apiv2/verification/verify.etherscan.spec.ts
+++ b/services/server/test/integration/apiv2/verification/verify.etherscan.spec.ts
@@ -32,6 +32,11 @@ import {
 use(chaiHttp);
 
 describe("POST /v2/verify/etherscan/:chainId/:address", function () {
+  // Don't run if it's an external PR. Etherscan tests need API keys that can't be exposed to external PRs.
+  if (process.env.CIRCLE_PR_REPONAME !== undefined) {
+    return;
+  }
+
   const chainFixture = new LocalChainFixture();
   const serverFixture = new ServerFixture();
   const sandbox = sinon.createSandbox();


manuelwedler:
actually yes. let me double check what's the problem

manuelwedler:
I don't understand why they are not working on external PRs. I tried removing the environment variable from the CI and running these tests and they still worked.

kuzdogan:
FWIW I didn't check why they didn't work. I saw the pattern and assumed it was the reason, without knowing we have been mocking the responses. If that's not the reason, then there's no issue.

manuelwedler:
I  checked again the Renovate PR: https://app.circleci.com/pipelines/github/ethereum/sourcify/8655/workflows/a801dd09-b69b-43a5-a073-65b00511af5a/jobs/51109/parallel-runs/0/steps/0-106

The issue is not the Etherscan API key (the response is mocked successfully). The issue is the RPC API key. The tests fail because they are not able to fetch the bytecode.

Should we just merge this PR? I can add a comment to explain this. @marcocastignoli

marcocastignoli (approved):
From my pov it's better if we mock every interaction with external services while running our integration tests, including mainnet rpc calls. Let's merge this and create an issue to mock mainnet rpc calls

kuzdogan:
I guess we should not merge this then. We are mocking in the end and the tests should not be skipped

manuelwedler:
> I guess we should not merge this then. We are mocking in the end and the tests should not be skipped

I agree that it would be better to mock the RPC calls, but I am currently not sure how to do it for API v2 tests because the RPC is called in the worker thread. 

I already created an issue for this: https://github.com/ethereum/sourcify/issues/2134

I'm still merging this, since it's better to have this fixed now. We can pull this issue into the next sprint if you like @kuzdogan

PR #2132, opened 2025-04-30
Fix flaky tests

### LINKED ISSUE #2107: Investigate flaky CI test: timeout error on verify via json test
We currently see this error frequently when running the tests in the CI.

```
sourcify-server:   1) POST /v2/verify/:chainId/:address
sourcify-server:        should verify a contract with Solidity standard input JSON:
sourcify-server:      Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/sourcify/services/server/test/integration/apiv2/verification/verify.json.spec.ts)
sourcify-server:       at listOnTimeout (node:internal/timers:581:17)
sourcify-server:       at processTimers (node:internal/timers:519:7)
```

It is not happening on every run but only every few times. We should check why this test is flaky.

manuelwedler:
Fixes #2107

Regarding the flaky tests on the new verification endpoints: The issue seems to be that they are the first test in the describe block (server is started at the beginning of the describe block) and it takes some time to spin up the worker thread. The first tests per describe block run for around 1 minute because they need to wait for the thread to be ready. Therefore, they hit only sometimes the 1 minute timeout.

On the compilers package, we also have a test where the compilation takes over 1 minute sometimes. This wasn't a problem before we had the compilers package, because this test was inside the lib-sourcify tests and there we have not timeout configured.

This simply increases the Mocha timeouts to 2 minutes. Hopefully, this is enough.

### DIFF
--- a/packages/compilers/.mocharc.json
+++ b/packages/compilers/.mocharc.json
@@ -1,5 +1,5 @@
 {
   "extension": [".spec.ts"],
   "require": ["ts-node/register"],
-  "timeout": 60000
+  "timeout": 120000
 }
--- a/services/server/.mocharc.json
+++ b/services/server/.mocharc.json
@@ -5,5 +5,5 @@
     "./test/load-env.js",
     "./test/global-mocks.ts"
   ],
-  "timeout": 60000
+  "timeout": 120000
 }


marcocastignoli (approved):
let's see if this fixes it :)

manuelwedler:
Yep, I hope so.. nothing guaranteed of course

Issue #2091, opened 2025-04-17
Investigate missing `constructorArguments` transformation in contract - 2


kuzdogan:
Following #2086 I think this time it's an unexpected issue I encountered. This time `creationMatch: true` but the constructorArguments transformation seems to be missing looking at the bytecodes.

API: https://sourcify.dev/server/v2/contract/1/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48?fields=all
Repo: https://repo.sourcify.dev/1/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48

The appended constructor args `0x0000000000000000000000000882477e7895bdc5cea7cb1552ed914ab157fe56` is not found under the `transformations`

marcocastignoli:
I notice that this contract was updated with the script on the 9th of April, we shoulld check other updated contract for this same error

kuzdogan:
Another case: https://repo.sourcify.dev/1/0x178608fFe2Cca5d36f3Fc6e69426c4D3A5A74A41

marcocastignoli:
In the upgrade script the compilation.metadata object was not mocked. extractConstructorArgumentsTransformation needs the metadata.abi object in order to extract the constructor arguments from the onchain creation bytecode. Without the metadata the constructorArguments transformation coulnd't be found.

marcocastignoli:
We fixed the upgrade script with https://github.com/ethereum/sourcify/pull/2135. We are executing again the upgrade script

manuelwedler:
The upgrade script ran successfully. All affected contracts should have the constructor transformations now. 
I had a few failed contracts in the first run, but I could upgrade all of them afterwards by manually calling the API. Probably they just failed due to network errors in the first run.

PR #2135, opened 2025-05-05
Fix missing compilation.metadata field in upgrade script

marcocastignoli:
See #2091

In the upgrade script the `compilation.metadata` object was not mocked. `extractConstructorArgumentsTransformation` needs the metadata.abi object in order to extract the constructor arguments from the onchain creation bytecode. Without the metadata the constructorArguments transformation coulnd't be found.

This PR adds the metadata field on the mock compilation

### DIFF
--- a/services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
+++ b/services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
@@ -135,11 +135,11 @@ export async function verifyDeprecated(
 }
 
 /**
- * Upgrades a contract by fetching the sourcify_match from the database and then verifying it again with the new verification parameters.
- * Mocks both the compilation and the sourcifyChain information so that recompilation and rpc calls are not performed.
+ * This function is used to fix the creation information (match, transformations list, transformations values, metadata match) of contract with misaligned creation data.
  *
- * Finally, it updates the creation fields (match, transformations list, transformations values, metadata match) in
- * the verified_contracts and the sourcify_matches tables with the new verification creation fields results.
+ * We beging by mocking both the compilation and the sourcifyChain information with data from the database, so that recompilation and rpc calls are not performed.
+ * Then we verify the contract again with the new mocked compilation and sourcifyChain objects.
+ * Finally we upgrade the misaligned contract by UPDATING the verified_contracts and sourcify_matches creation fields.
  */
 export async function upgradeContract(
   req: LegacyVerifyRequest,
@@ -268,6 +268,7 @@ export async function upgradeContract(
     );
 
     compilation.compilerOutput = {
+      sources: compilationArtifacts.sources,
       contracts: {
         [compilationTarget.path]: {
           [compilationTarget.name]: {
--- a/services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
+++ b/services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
@@ -134,6 +134,13 @@ export async function verifyDeprecated(
   }
 }
 
+/**
+ * Upgrades a contract by fetching the sourcify_match from the database and then verifying it again with the new verification parameters.
+ * Mocks both the compilation and the sourcifyChain information so that recompilation and rpc calls are not performed.
+ *
+ * Finally, it updates the creation fields (match, transformations list, transformations values, metadata match) in
+ * the verified_contracts and the sourcify_matches tables with the new verification creation fields results.
+ */
 export async function upgradeContract(
   req: LegacyVerifyRequest,
   res: Response,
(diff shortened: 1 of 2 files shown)


kuzdogan:
Could other fields have been missed on the compilation? I think we need to look into why metadata was missed. E.g. I also don't see the `compilationTime` being mocked. 

Also if metadata was missed:
1. how did these verifications succeed? 
2. That means the `sourcify_matches.metadata` field was empty for the affected contracts?

kuzdogan:
Another thing, why didn't `extractConstructorArgumentsTransformation` throw when abi was not found?

marcocastignoli:
> Could other fields have been missed on the compilation?

We should check every field used by the verification. I missed metadata.abi because it was inside `transformtions.ts`, I'll check again

> how did these verifications succeed? That means the sourcify_matches.metadata field was empty for the affected contracts?

No because we weren't updating that field in the database

marcocastignoli:
> Another thing, why didn't extractConstructorArgumentsTransformation throw when abi was not found?

That's why I didn't notice this error. Probably because a failed creation match shouldn't block a successful runtime match.

kuzdogan:
> That's why I didn't notice this error. Probably because a failed creation match shouldn't block a successful runtime match.

It's fine if it did and it should do block the creation match. It should've thrown here:

https://github.com/ethereum/sourcify/blob/9e2a5a663c65764a7bf080f4962d5788477b4b58/packages/lib-sourcify/src/Verification/Transformations.ts#L210-L215

Then this should have been caught here: 

https://github.com/ethereum/sourcify/blob/9e2a5a663c65764a7bf080f4962d5788477b4b58/packages/lib-sourcify/src/Verification/Verification.ts#L207-L216

But the `creationMatch` is already set to `true`, because it is set before the constructor arguments are being extracted: 

https://github.com/ethereum/sourcify/blob/9e2a5a663c65764a7bf080f4962d5788477b4b58/packages/lib-sourcify/src/Verification/Verification.ts#L476-L481

Setting the matches etc. should always be done at the very end of the matching function

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -134,6 +134,13 @@ export async function verifyDeprecated(
   }
 }
 
+/**
+ * Upgrades a contract by fetching the sourcify_match from the database and then verifying it again with the new verification parameters.
```
kuzdogan, line 138: > Upgrades a contract...

by creating a new `verified_contracts` entry or modifies the existing entry?

> with the new verification parameters.

What are the new verification parameters?

marcocastignoli: What about this?

```
/**
* This function is used to fix the creation information (match, transformations list, transformations values, metadata match) of contract with misaligned creation data.
 * This function mocks both the compilation and the sourcifyChain information so that recompilation and rpc calls are not performed. The mocked information is fetched from the database.
 * Finally it upgrades a contract by UPDATING the verified_contracts and sourcify_matches creation fields.
 */
```

marcocastignoli:
> Setting the matches etc. should always be done at the very end of the matching function

The reason being we re-use the same function `matchBytecodes` for both `matchWithCreationTx` and `matchWithRuntimeBytecode`, but in the creation case we execute an additional transformation after the shared logic. Should we change that? The alternative is resetting the match inside this catch case:

https://github.com/ethereum/sourcify/blob/9e2a5a663c65764a7bf080f4962d5788477b4b58/packages/lib-sourcify/src/Verification/Verification.ts#L207-L216

Like this

```ts
      } catch (e: any) {
        logWarn('Error matching with creation tx', {
          chain: this.sourcifyChain.chainId,
          address: this.address,
          creatorTxHash: this.creatorTxHash,
          error: e.message,
        });
        this.creationMatch = null;
        this.creationTransformations = [];
        this.creationTransformationValues = {};
        this.creationLibraryMap = {};
        this.creatorTxHash = undefined;
      }
```

What do you think?

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
marcocastignoli: I updated the comment, let me know if not it's more clear.

marcocastignoli:
> It's fine if it did and it should do block the creation match. It should've thrown here:
> `Failed to match with creation bytecode: constructor ABI Inputs are missing`, 

I'm not sure about this, because there could be contract without a constructor object in the ABI because they don't have a constructor. That should not block the verification.

marcocastignoli:
> Setting the matches etc. should always be done at the very end of the matching function

Yes, you are right! I pushed https://github.com/ethereum/sourcify/pull/2135/commits/473d9fc3207ef6b4b023b160ed27c648850aca16 to fix this

kuzdogan (approved):
Approved but can you create a separate issue for us to check all occurrences where we might be assigning the class variables too early like in this case?

marcocastignoli:
I added the new issue here: https://github.com/ethereum/sourcify/issues/2138

PR #2137, opened 2025-05-06
Fix traceId not logged inside worker when verifying via Etherscan import

### LINKED ISSUE #2074: Inject `traceId` into the loggers of the packages

manuelwedler:
Fixes #2074

Call internal `_verifyFromJsonInput` inside `_verifyFromEtherscan` instead of public function. Needed because the traceId was overwritten with undefined when calling the public function.

### DIFF
--- a/services/server/src/server/services/workers/verificationWorker.ts
+++ b/services/server/src/server/services/workers/verificationWorker.ts
@@ -256,7 +256,7 @@ async function _verifyFromEtherscan({
     processedResult = processSolidityResultFromEtherscan(etherscanResult, true);
   }
 
-  return verifyFromJsonInput({
+  return _verifyFromJsonInput({
     chainId,
     address,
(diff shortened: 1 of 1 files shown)


kuzdogan (approved):
Approved.

But how does calling via internal makes a difference?

manuelwedler:
in the public function we set the trace. The traceId is passed when we call from the VerificationService, but the trace is not passed from the _verifyFromEtherscan function (we could also pass it inside the _verifyFromEtherscan as an alternative fix, but imo less clean alternative).

PR #2142, opened 2025-05-07
Release process fix 2

kuzdogan:
Fixes to the release process following the last release call:

-  Use a temporary file for storing package version etc. in release scripts instead of memory variables
- use `--include-merge-tags` flag to correctly determine which packages are changes
- Fix GUI editors like `code` exiting before writing to temp changelog file.

I ran it to the extent that I can (without publishing a release etc.) and seems stable. We can check again on our next release if it goes well 🤞

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ workspace/
 # Ignore the server local config file. See .nxignore
 **/config/local.js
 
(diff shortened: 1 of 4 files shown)


marcocastignoli (approved):
lgtm

PR #2143, opened 2025-05-07
Handle `:` characters in contract paths correctly

### LINKED ISSUE #2043: Handle `:` characters in metadata library paths
I found an issue in the way the `SolidityMetadataContract` handles libraries. It is about this line:

https://github.com/ethereum/sourcify/blob/c6e61655bdd9f2ba54b8faefc7e1e9aa149a0e78/packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts#L300

Here, we assume that the result of `split` is always a two element array. But actually the path can contain `:` characters itself. I know about this because the default test contract in the server tests has a `:` in the path.

manuelwedler:
Closes #2043


kuzdogan (approved):
lgtm

Issue #2138, opened 2025-05-06
Check for early variable assignments in the new verification flow


marcocastignoli:
In [this comment](https://github.com/ethereum/sourcify/pull/2135#issuecomment-2851253015), Kaan outlines a problem where creation_match values are assigned in the middle of the `matchWithCreationTx` process. As a result, creation matches are stored even if errors occur later in the `matchWithCreationTx` process after the assignment. In PR #2135, we fix this issue by moving the creation_match assignment to the end of the `matchWithCreationTx` process.

We should double-check if there are other occurrences of early variable assignments in the rest of the verification flow, in particular in the Verification class.

marcocastignoli:
I checked for early variable assignments and **I didn't find any bad cases**.

### Early assignment of `onchainRuntimeBytcode`

https://github.com/ethereum/sourcify/blob/680e034e25cc24d364b49fab9bbbcc5deb352c31/packages/lib-sourcify/src/Verification/Verification.ts#L70-L88

At the beginning of the verify method we assign the `onchainRuntimeBytecode` so even if the verification fails, the value `onchainRuntimeBytecode` will still be exported from the class. I think we can accept such cases because we read from a `Verification` instance when we have a successful verification.

### Early assignment of creation information

_Premise, also these cases are not used if not when we have a successful match, so it's not an important fix__

The only weird behavior I noticed is that in the catch of `matchWithCreationTx` we assign `this.creatorTxHash = undefined;` but not the other fields. 

https://github.com/ethereum/sourcify/blob/680e034e25cc24d364b49fab9bbbcc5deb352c31/packages/lib-sourcify/src/Verification/Verification.ts#L206-L216

Similarly we could could assign undefined to the other creation information: `blockNumber`, `deployer`, `onchainCreationBytecode` and `txIndex`. From my pov it makes sense to assign them later because one of the reason we don't find a creation match is because the wrong `creatorTxHash` was provided.

kuzdogan:
I think in the second case with the creation information, it's weird that we just set the creatorTxHash to `undefined` and not other anyway. It should be either all or none. 

About having the creation information or not if there's no match: I was going to say we should not have creation info if we don't have a match because we wouldn't know if this creatorTxHash is correct. But actually we do check inside `getContractCreationBytecodeAndReceipt`  if this txHash actually created this contract:

[...]

PR #2154, opened 2025-05-13
Remove creatorTxHash's undefined assignment when creation match fails

marcocastignoli:
See #2138

### DIFF
--- a/services/server/test/integration/apiv2/verification/verify.json.spec.ts
+++ b/services/server/test/integration/apiv2/verification/verify.json.spec.ts
@@ -400,4 +400,46 @@ describe("POST /v2/verify/:chainId/:address", function () {
     chai.expect(verifyRes.body).to.have.property("errorId");
     chai.expect(verifyRes.body).to.have.property("message");
   });
+
+  it("should fail matching with creation tx if the provided creationTransactionHash does not match the contract address", async () => {
+    // Deploy contract A
+    const deploymentA = await deployFromAbiAndBytecodeForCreatorTxHash(
+      chainFixture.localSigner,
+      chainFixture.defaultContractArtifact.abi,
+      chainFixture.defaultContractArtifact.bytecode,
+    );
+
+    // Deploy contract B
+    const deploymentB = await deployFromAbiAndBytecodeForCreatorTxHash(
+      chainFixture.localSigner,
+      chainFixture.defaultContractArtifact.abi,
+      chainFixture.defaultContractArtifact.bytecode,
+    );
+
+    const { resolveWorkers } = makeWorkersWait();
+
+    // Try to verify contract A, but provide B's creatorTxHash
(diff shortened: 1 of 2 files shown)


kuzdogan (approved):

kuzdogan (changes requested):
The tests are failing

kuzdogan:
I'd like to be sure the check inside here fully checks if this creatorTxHash really corresponds to this contract 

https://github.com/ethereum/sourcify/blob/ede4d41509f5310c7f501c50b3570335057642dc/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts#L645-L646

Do we have a test for that, for what happens when the user provides an incorrect creatorTxHash?

marcocastignoli:
There is already a test in lib-sourcify

https://github.com/ethereum/sourcify/blob/7407627e089d5bb30a0d77414d3045bbda31573e/packages/lib-sourcify/test/Verification/Verification.spec.ts#L1269-L1305

But I added also one in apiV2 integration tests.

services/server/test/integration/apiv2/verification/verify.json.spec.ts
```diff
@@ -400,4 +400,46 @@ describe("POST /v2/verify/:chainId/:address", function () {
...
+
+    await resolveWorkers();
+
+    // Fetch the job result
+    const jobRes = await chai
+      .request(serverFixture.server.app)
+      .get(`/v2/verify/${verifyRes.body.verificationId}`);
+
+    chai.expect(jobRes.status).to.be.oneOf([200]);
+    chai.expect(jobRes.body.contract.creationMatch).to.be.null;
```
manuelwedler, line 443: Can you also check if the job is done? While the job is pending, creationMatch is also null.

marcocastignoli: Thanks for the heads up

services/server/test/integration/apiv2/verification/verify.json.spec.ts
```diff
@@ -400,4 +400,49 @@ describe("POST /v2/verify/:chainId/:address", function () {
...
+    // Fetch the job result
+    const jobRes = await chai
+      .request(serverFixture.server.app)
+      .get(`/v2/verify/${verifyRes.body.verificationId}`);
+
+    chai.expect(jobRes.status).to.be.oneOf([200]);
+    chai.expect(jobRes.body).to.include({
+      isJobCompleted: true,
+    });
+    chai.expect(jobRes.body.contract.creationMatch).to.be.null;
```
kuzdogan, line 446: We should also check if the runtimeMatch is true, and the errors are empty

marcocastignoli: Done

PR #2158, opened 2025-05-14
Use Etherscan API v2

### LINKED ISSUE #2125: Switch to Etherscan API v2
Etherscan is discontinuing v1 and ask users to "switch to API V2 by May 31, 2025". We should plan how to make the switch, primarily which account to use as the main one or if we want to keep using separate accounts etc. 

https://x.com/etherscan/status/1849772046089322588

manuelwedler:
Closes #2125

I will also update all of our API keys after merging this.


marcocastignoli (approved):
lgtm!

Issue #1980, opened 2025-03-07
Contract failing to verify with creation bytecode with possibly wrong normalization


kuzdogan:
While looking for a contract with many cborAuxdata fields for the new repo UI, I came across this one:

Address: 0x80a477a11b9AaFfaDc97Dc2785E9DB32859815F5
Chain: Base Sepolia (84532)

The contract is verified with the runtime bytecode but not with creation. 

Apparently Etherscan also wasn't able to verify with the creation bytecode https://sepolia.basescan.org/address/0x80a477a11b9aaffadc97dc2785e9db32859815f5#code 

<img width="790" alt="Image" src="https://github.com/user-attachments/assets/a8c61679-625b-4dea-95c7-02e3a8a2262e" />

The APIv2 returns the creation bytecode regardless (as discussed in https://github.com/ethereum/sourcify/issues/1979). One can spot several differences between the onchain vs. recompiled creation bytecode and most of them seemed to be the CBORAuxdata to me. So I investigated further and checked why it fails even after normalization.

The bytecodes at this line are the following:

https://github.com/ethereum/sourcify/blob/69e1927e791f81f766c244684d14d48fa1bac9ad/packages/lib-sourcify/src/lib/verification.ts#L660-L665

normalized onchain creation bytecode
[normalized-onchain-84532-0x80a477.txt](https://github.com/user-attachments/files/19129505/normalized-onchain-84532-0x80a477.txt)

normalized recompiled creation bytecode
[normalized-recompiled-84532-0x80a477.txt](https://github.com/user-attachments/files/19129498/normalized-recompiled-84532-0x80a477.txt)

You can check the diff in https://www.diffchecker.com/text-compare/ (choose Highlight: Character on the left sidebar)

From what I can see there are three differences after normalization:

### 1. Library 

onchain `bad4a09a79f4dd0cac798eaf7ba390be21c48794` vs. recompiled `42cc7d9bbcad1032b95aa9c26e95167f1750d8a3`. This apparently is a library address given in the compiler settings:

```json
    "libraries": {
      "src/TraitCategory.sol": {
        "TraitCategory": "0x42cc7d9bbcad1032b95aa9c26e95167f1750d8a3"
      }
    },
```

This one is expected as this is a linked library and the address is linked at the compile time, so there won't be a library placeholder we can look for. Apparently the user changed the library before deploying onchain. There's not much we can do about it.

### 2. CBOR Auxdatas

[...]

PR #2159, opened 2025-05-14
Fix `findAuxdataPositions` handling the case of multiple equal auxdata

marcocastignoli:
See #1980


kuzdogan:
Can you please explain what this solves? The issue provided is only providing some possible reasons to why this verificaition fails. I also don't fully remember this case and what I mean by "zeroed cborAuxdata" etc.

The linked library on the onchain vs. recompiled are different so I wouldn't expect this contract to be creation-verified anyway. Was it still not verifying even if you've set the library address equal to the one onchain?

Also what are you testing against in your bytecode? The bytecode in the artifact.json is identical to the recompiled bytecode in the issue. So it is expected for the test to pass. 

Feel free to send a short video or hop on a call if it's difficult to explain.

Bonus: It's still really difficult to read the `findAuxdataPositions` function code 😢

marcocastignoli:
> Bonus: It's still really difficult to read the findAuxdataPositions function code

I opened https://github.com/ethereum/sourcify/pull/2160 with a proposal of a refactored `findAuxdataPositions` function

kuzdogan (approved):

PR #2160, opened 2025-05-15
Refactor findAuxdataPositions and add comments

marcocastignoli:


packages/lib-sourcify/src/Compilation/auxdataUtils.ts
```diff
@@ -88,56 +88,75 @@ function getAuxdatasDiff(originalAuxdatas: string[], editedAuxdatas: string[]) {
 }
 
 /**
- * Finds the positions of the auxdata in the bytecode.
- * The compiler outputs the auxdata values in the `legacyAssembly` field. However we can't use these values to do a simple string search on the compiled bytecode because an attacker can embed these values in the compiled contract code and cause the correspoding field in the onchain bytecode to be ignored falsely during verification.
- * A way to find the *metadata hashes* in the bytecode is to recompile the contract with a slightly edited source code and compare the differences in the raw bytecodes. However, this will only give us the positions of the metadata hashes in the bytecode. We need to find the positions of the whole *auxdata* in the bytecode.
- * So we go through each of the differences in the raw bytecode and check if an auxdata diff value from the legacyAssembly is included in that difference. If it is, we have found the position of the auxdata in the bytecode.
+ * Maps every legacy-assembly auxdata block to its `{ offset, value }`
```
manuelwedler, line 91: I found `legacy-assembly` a bit confusing at the beginning when I read it because the term is not used anymore afterwards. It took some time until I understood that the auxdata strings are from legacy assembly. Maybe we can add a note at the parameter description below.

manuelwedler (approved):
Looks much better understandable now :+1: Only an improvement to the function comment could be made.

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
kuzdogan: Yes let's give background about the compiler's legacyAssembly output. IMO the previous background paragraphs were quite comprehensive

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
```diff
@@ -88,56 +88,75 @@ function getAuxdatasDiff(originalAuxdatas: string[], editedAuxdatas: string[]) {
...
+ * @param editedAuxdatas   – auxdata strings coming from the edited compile
+ *
+ * @returns an object whose keys are the indexes of the auxdata items
+ *          (as they appear in `legacyAssembly`) and whose values are
+ *          their position and value in the originalBytecode.
  */
 export function findAuxdataPositions(
   originalBytecode: string,
   editedBytecode: string,
   originalAuxdatas: string[],
```
kuzdogan, line 106: I'm wondering if renaming this something like `auxdatasFromCompiler` `editedAuxdatasFromCompiler` is easier to understand. The "auxdata" term is lacking specificity and causing confusion

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
```diff
@@ -88,56 +88,75 @@ function getAuxdatasDiff(originalAuxdatas: string[], editedAuxdatas: string[]) {
...
   editedAuxdatas: string[],
 ): CompiledContractCborAuxdata {
-  const auxdataDiffObjects = getAuxdatasDiff(originalAuxdatas, editedAuxdatas);
+  // 1. Build auxdataDiffs that explain how each auxdata is changing after recompilation with slightly different sources
 
-  const diffPositionsBytecodes = getDiffPositions(
+  // A single AuxdataDiff has 3 keys: real, diffStart (int offset of diff), diff
```
kuzdogan, line 111: Let's put the AuxdataDiff object here in comments:

{ real: `a264....0033`, diffStart: 10, diff: `dcec...52b6`}

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
```diff
@@ -88,56 +88,75 @@ function getAuxdatasDiff(originalAuxdatas: string[], editedAuxdatas: string[]) {
...
+  // original: a2646970667358221220123af2412fd4b1b89740f76c6ab76a6412bbde98fab732fb97eb012abe7123ff64736f6c63430007000033
+  //           └────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+  //               REAL
+  // edited:   a2646970667358221220dceca8706b29e917dacf25fceef95acac8d90d765ac926663ce4096195952b6164736f6c63430007000033
+  //           └──────────────────┘└──────────────────────────────────────────────────────────────┘
+  //               DIFFSTART           DIFF
+  const auxdataDiffs = getAuxdatasDiff(originalAuxdatas, editedAuxdatas);
```
kuzdogan, line 118: In line with the suggestion above to change param names
```suggestion
  const compilerAuxdataDiffs = getAuxdatasDiff(originalAuxdatas, editedAuxdatas);
```

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
```diff
@@ -88,56 +88,75 @@ function getAuxdatasDiff(originalAuxdatas: string[], editedAuxdatas: string[]) {
...
-          offset:
-            (diffPosition -
-              auxdataDiffObjects[auxdataDiffIndex].diffStart -
-              2) /
-            2,
-          value: `0x${auxdataDiffObjects[auxdataDiffIndex].real}`,
+
+    // For each bytecodeDiffPosition, test every auxdataDiff that we haven't already mapped
+    for (let idx = 0; idx < auxdataDiffs.length; idx++) {
+      // The first CompiledContractCborAuxdata element key is '1'
```
kuzdogan, line 139: I don't understand this line.

Do CBORAuxdata's always have keys '1', '2', '3'...?

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
```diff
@@ -88,56 +88,75 @@ function getAuxdatasDiff(originalAuxdatas: string[], editedAuxdatas: string[]) {
...
+      const resultIndex = idx + 1;
+
+      // If we already mapped this auxdata, skip
+      if (result[resultIndex] !== undefined) continue;
+
+      const diff = auxdataDiffs[idx];
+      // If in position `pos` of the originalBytecode we find the `auxdataDiff.real` value
```
kuzdogan, line 146: Just came to my mind, should we also double check if `editedBytecode` has `diff`?

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
```diff
@@ -88,56 +88,75 @@ function getAuxdatasDiff(originalAuxdatas: string[], editedAuxdatas: string[]) {
...
+      if (bytecodeIncludesAuxdataDiffAt(originalBytecode, diff, pos)) {
+        // Store the CompiledContractCborAuxdata element
+        result[resultIndex] = {
+          offset: (pos - diff.diffStart - 2) / 2,
+          value: `0x${diff.real}`,
         };
+        // One auxdata for each bytecodeDiffPosition. See https://github.com/ethereum/sourcify/issues/1980
```
kuzdogan, line 153: I think we mean first here

```suggestion
        // Match the first auxdata for each bytecodeDiffPosition. If there are multiple identical cborAuxdata, without the break it will always match the last one. With first, it will be "mapped" in the `result[resultIndex]` See https://github.com/ethereum/sourcify/issues/1980
```

kuzdogan (commented):
Some more comments. Sorry this is taking longer but I think it will be helpful longer term to make this function easy to understand. Everytime I read this I'm confused

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
marcocastignoli: Yes, we use those values as `id` in the Transformation

```
{
  type: 'replace',
  reason: 'cborAuxdata',
  offset,
  id,
}
```

marcocastignoli:
No need to be sorry. You are totally right, it's a very important function and we should make it as clear as possible 👍

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
marcocastignoli: Do you have a case in mind in which this double check would actually be helpful? I cannot think how the current approach can fail

packages/lib-sourcify/src/Compilation/auxdataUtils.ts
```diff
@@ -54,90 +57,124 @@ function getDiffPositions(original: string, modified: string): number[] {
...
+      auxdatasFromCompiler[i],
+      editedAuxdatasFromCompiler[i],
     );
-    auxdataDiffs.push({
-      real: originalAuxdatas[i],
+    compilerAuxdataDiffs.push({
+      real: auxdatasFromCompiler[i],
       diffStart: diffPositions[0],
-      diff: originalAuxdatas[i].substring(
+      diff: auxdatasFromCompiler[i].substring(
```
kuzdogan, line 87: I see the `diff` field used to be from `originalAuxdatas` but below in the diagram we show it on the edited part. Either one of them is false

kuzdogan: The `diffStart` also seems to be wrong in that case

packages/lib-sourcify/test/Verification/Verification.spec.ts
```diff
@@ -607,7 +607,7 @@ describe('Verification Class Tests', () => {
     // The function `findAuxdataPositions` used to fail to assign the proper offset values resulting in a null creation match
     // Read more here: https://github.com/ethereum/sourcify/issues/1980
     // Fixed by PR: https://github.com/ethereum/sourcify/pull/2159
-    it('should verify a contract with multiple equal auxdatas', async () => {
+    it.only('should verify a contract with multiple equal auxdatas', async () => {
```
kuzdogan: Remove only pls

kuzdogan: @marcocastignoli

PR #2162, opened 2025-05-20
Add `errorData` to job endpoint

marcocastignoli:
This PR:
- Adds a shared `OutputError` type for both `SolidityOutput` and `VyperOutput`
- Adds a new `CompilerError` type to keep track of compiler errors
- Adds `errorData` to job endpoint
- Integrate `errorData` in tests
- Adds a new test to validate Vyper compilation errors in verification APIs
- Updates the `apiv2.yaml` to display an example of `errorData` in a job containing errors


packages/lib-sourcify/src/SourcifyLibError.ts
```diff
@@ -13,6 +14,8 @@ interface SourcifyLibErrorDataRequired {
   missingSources: string[];
   invalidSources: string[];
   compilationTargets: string[];
+  compilerErrorMessage: string;
+  compilerErrors: OutputError[];
```
manuelwedler, line 18: You might need to mark them as optional here

services/server/src/apiv2.yaml
```diff
@@ -386,25 +386,33 @@ components:
...
+                  errorId: "f319b7fc-0286-4c23-bf1f-75996419a2ce"
+                  onchainRuntimeCode: "0x6080604052348015600f57600080fd5b506004361060325760003560e01c80632e64cec11460375780636057361d146051575b600080fd5b603d6069565b6040516048919060c2565b60405180910390f35b6067600480360381019060639190608f565b6072565b005b60008054905090565b8060008190555050565b60008135905060898160e5565b92915050565b60006020828403121560a057600080fd5b600060ac84828501607c565b91505092915050565b60bc8160db565b82525050565b600060208201905060d5600083018460b5565b92915050565b6000819050919050565b60ec8160db565b811460f657600080fd5b5056fea264697066735822122005183dd5df276b396683ae62d0c96c3a406d6f9dad1ad0923daf492c531124b164736f6c63430008040033"
+                  creationTransactionHash: "0xdd419b049534d2ead35de292ce73f3a9d9ecad33cdc1da1cae1d9fa3a86c74e1"
+                  errorData:
```
manuelwedler, line 404: You need to add the new field to the `MatchingErrorResponse` schema

services/server/test/integration/apiv2/verification/verify.json.spec.ts
```diff
@@ -195,6 +200,84 @@ describe("POST /v2/verify/:chainId/:address", function () {
...
+        chainFixture.localSigner,
+        vyperArtifact.abi,
+        vyperArtifact.bytecode,
+      );
+
+    const verifyRes = await chai
+      .request(serverFixture.server.app)
+      .post(`/v2/verify/${chainFixture.chainId}/${contractAddress}`)
+      .send({
+        language: "Vyper",
```
manuelwedler, line 234: language does not exist anymore. we only have it inside the stdJsonInput

manuelwedler (changes requested):

services/server/test/integration/apiv2/verification/verify.json.spec.ts
manuelwedler: I see.. it's also still in the other test, but could be removed

services/server/src/apiv2.yaml
marcocastignoli: The custom code is wrong in the `MatchingErrorResponse`. It should be `no_match`, instead of `no_matching_bytecodes`

packages/compilers-types/src/CompilationTypes.ts
```diff
@@ -70,6 +70,17 @@ export interface Userdoc {
   version?: number;
 }
 
+// Shared OutputError type between Solidity and Vyper
+export interface OutputError {
```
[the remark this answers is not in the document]
marcocastignoli: It wasn't there before, should we include it in this PR?

services/server/src/apiv2.yaml
marcocastignoli: Also in the case of `no_match` there is no `errorData`

packages/compilers-types/src/CompilationTypes.ts
```diff
@@ -70,6 +70,17 @@ export interface Userdoc {
...
+  component: string;
+  formattedMessage: string;
```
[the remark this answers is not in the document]
manuelwedler: sorry for not being clear. I meant the `formattedMessage` property. You already got it I saw

packages/lib-sourcify/README.md
```diff
@@ -213,6 +213,8 @@ setLibSourcifyLogger({
 
 The `lib-sourcify` library does not come with compilers as dependencies. Instead, you need to provide a class that implements either the `ISolidityCompiler` or `IVyperCompiler` interface and pass it to any function that requires a compiler. We suggest you to use the official [`@ethereum-sourcify/compilers`](https://github.com/ethereum/sourcify/tree/staging/packages/compilers) package.
 
+In the case of an error in the `compile` function, Sourcify's server expect the throwned error to have a public `errors` property containing the compilation output's `errors` property.
```
manuelwedler, line 216:
```suggestion
In the case of an error in the `compile` function, `lib-sourcify` expects the thrown error to have a public `errors` property containing the compilation output's `errors` property.
```

services/server/src/apiv2.yaml
manuelwedler: The `MatchingErrorResponse` still needs the `errorData` under `allOf`. I don't mean the example. The example can stay as it is.


I think we shouldn't define every possible property for the `errorData` in the schema, as we might add more quickly. Just give it the Object type without further definition.

manuelwedler: > Also, I cannot understand where MatchingErrorResponse is displayed in the docs UI

It's a bit hidden in Swagger (it was more easily visible in StopLight). Just go to the 200 response for the job endpoint and click on Schema (instead of example value). Then expand the error property.

manuelwedler (changes requested):

services/server/src/apiv2.yaml
marcocastignoli: Ah now it's clear :) Thanks, I didn't get this

manuelwedler (approved):
Looks good!

PR #2179, opened 2025-05-27
Handle invalid json body error, throw if unable to fetch the compiler

kuzdogan:
Handles the case when the `bodyParser.json` throws in the server before reaching other validations.

Also throws when compilers can't be fetched (e.g. an invalid compiler version)

My only question would be if we should log these internally. I guess since these are "expected" or "handled" errors it's ok not to. But I think I'd want to see these errors when debugging so I added a generic log.debug statement in `genericErrorHandler`

### DIFF
--- a/packages/compilers/src/lib/vyperCompiler.ts
+++ b/packages/compilers/src/lib/vyperCompiler.ts
@@ -52,12 +52,6 @@ export async function useVyperCompiler(
     version,
   );
 
-  if (!vyperPath) {
-    throw new Error(
-      'Vyper path not found. Maybe an incorrect version was provided.',
-    );
-  }
-
   let compiled: string | undefined;
   const inputStringified = JSON.stringify(vyperJsonInput);
   const startCompilation = Date.now();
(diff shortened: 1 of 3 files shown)


services/server/src/common/errors/GenericErrorHandler.ts
```diff
@@ -17,6 +17,8 @@ export default function genericErrorHandler(
       logger.error(`Unexpected server error: ${err.message}`, { error: err });
     }
 
+    logger.debug("❌ Error in genericErrorHandler", { error: err });
```
manuelwedler, line 20: Can you log another message? The current is ambigous because we have the same log in the catch. There it means an error was thrown inside the genericErrorHandler.

kuzdogan:
Also added a catch block for when the solc binary fetching fails. In that case we'll fallback to solc-js. If that also fails then it should throw

services/server/src/common/errors/GenericErrorHandler.ts
manuelwedler: Yes, it's a last resort. I think at the moment it should not be thrown, but we introduced it because we had some error in the past that was not handled and made the server crash.

packages/compilers/src/lib/vyperCompiler.ts
```diff
@@ -192,8 +192,14 @@ async function fetchAndSaveVyper(
...
+      platform,
+      vyperPath,
+      githubVyperURI,
+    });
+    throw new Error(
+      `Failed fetching vyper ${version} for platform ${platform}. Please check if the version is valid.`,
+    );
   }
-
-  return false;
```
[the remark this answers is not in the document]
kuzdogan: Good idea.

manuelwedler (approved):
Looks good! thanks :)

PR #2183, opened 2025-06-02
V2 address lookup on all chains endpoint

### LINKED ISSUE #2114: New v2 endpoint to check contract deployment chainIds by address

kuzdogan:
Closes #2114 

Adding a new enpoint under `/v2/contract/allChains/:address`. Let me know if you have feedback regarding the endpoint naming choice.

The endpoint returns an array of VerifiedContractMinimal under the `results` field on the response body. If there are no contracts at this address on any chain, it returns `404` with an results array that is empty `[]`. Idk if this is good practice and we should return `200`?

- Bumps the APIv2 version to v2.1.0
- Implements the queries, handlers, and the endpoint.
- Add tests

I appreciate a detailed look on the DB query. 

Todo:
- [ ] We need to add "address" as a new index on the `contract_deployments` table. Both on VerA and Sourcify IMO. We should see how we incorporate this as part of a migration etc.

### DIFF
--- a/services/server/src/apiv2.yaml
+++ b/services/server/src/apiv2.yaml
@@ -147,7 +147,7 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerErrorResponse'
       security: []
-  '/v2/contract/allChains/{address}':
+  '/v2/contract/all-chains/{address}':
     get:
       tags:
         - Contract Lookup
--- a/services/server/src/apiv2.yaml
+++ b/services/server/src/apiv2.yaml
@@ -147,6 +147,68 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerErrorResponse'
       security: []
+  '/v2/contract/allChains/{address}':
+    get:
+      tags:
+        - Contract Lookup
+      summary: Get verified contract at an address on all chains
+      description: |-
+        Returns all verified deployments at an address on all Sourcify chains (including deprecated ones).
+
+        Success returns an array of VerifiedContractMinimal objects under `results` field.
+
+        If not verified on any chain, the `results` array will be empty.
+      operationId: get-contract-all-chains
+      parameters:
+        - $ref: '#/components/parameters/address'
+      responses:
+        '200':
+          description: Returns all verified deployments at an address on all Sourcify chains
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  results:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/VerifiedContractMinimal'
+              examples:
+                Example 1:
+                  value:
+                    results:
+                      - match: exact_match
+                        creationMatch: match
+                        runtimeMatch: exact_match
+                        chainId: '1'
+                        address: '0xDFEBAd708F803af22e81044aD228Ff77C83C935c'
+                        verifiedAt: '2024-07-24T12:00:00Z'
+                      - match: match
+                        creationMatch: match
+                        runtimeMatch: match
+                        chainId: '11155111'
+                        address: '0xDFEBAd708F803af22e81044aD228Ff77C83C935c'
+                        verifiedAt: '2025-02-10T14:45:00Z'
+        '400':
+          $ref: '#/components/responses/BadRequestResponse'
+        '404':
+          description: Not Found
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  results:
+                    type: array
+                    items: {}
+              examples:
+                Example 1:
+                  value:
+                    results: []                            
+        '429':
+          $ref: '#/components/responses/TooManyRequestsResponse'
+        '500':
+          $ref: '#/components/responses/InternalServerErrorResponse'
   '/v2/contracts/{chainId}':
     get:
       tags:
@@ -884,7 +946,10 @@ components:
                 message: You are sending too many requests
                 errorId: 1ac6b91a-0605-4459-93dc-18f210a70192
     ListVerifiedContracts:
-      description: ''
+      description: |-
+        Returns an array of VerifiedContractMinimal objects under `results` field.
+
+        If not verified on any chain, the `results` array will be empty.
       content:
         application/json:
           schema:
--- a/services/server/src/apiv2.yaml
+++ b/services/server/src/apiv2.yaml
@@ -172,6 +172,7 @@ paths:
                   results:
                     type: array
                     items:
+                      # Not using ref:ListVerifiedContracts because its example has different addresses in results. Here we want to have the same address in results. The definitions are the same.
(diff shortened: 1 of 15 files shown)


services/server/src/server/apiv2/lookup/lookup.routes.ts
```diff
@@ -3,12 +3,20 @@ import {
...
+  getContractEndpoint,
+  listContractsEndpoint,
+} from "./lookup.handlers";
 
 import { Router } from "express";
 
 const router = Router();
 
+router
+  .route("/contract/allChains/:address")
```
manuelwedler: How about a simpler `/contract/all/:address`? Seems in-line to me with the other endpoints since the chain is always expected to be the first parameter.

services/server/test/integration/apiv2/lookup.spec.ts
```diff
@@ -14,9 +14,93 @@ import fs from "fs";
...
 import Sinon from "sinon";
 import * as proxyContractUtil from "../../../src/server/services/utils/proxy-contract-util";
+import {
+  SourcifyChain,
+  SourcifyChainMap,
+} from "@ethereum-sourcify/lib-sourcify";
 
 chai.use(chaiHttp);
 
+describe("GET /v2/contract/allChains/:address", function () {
```
manuelwedler: How about splitting this file into one file for each endpoint? I think this file is growing too big. See the verification endpoints folder for consistency about the naming.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -735,6 +735,26 @@ export class SourcifyDatabaseService
...
+    const result = await this.database.getSourcifyMatchesAllChains(
+      bytesFromString(address),
+    );
+
+    const results: VerifiedContractMinimal[] = result.rows.map((row) => ({
+      match: getTotalMatchLevel(row.creation_match, row.runtime_match),
+      creationMatch: toMatchLevel(row.creation_match),
+      runtimeMatch: toMatchLevel(row.runtime_match),
+      matchId: row.id,
+      chainId: row.chain_id.toString(),
```
manuelwedler: nitpick: chain_id is already a string

services/server/test/integration/apiv2/lookup.spec.ts
```diff
@@ -14,9 +14,93 @@ import fs from "fs";
...
+        rpcWithoutApiKeys: [`http://localhost:${TEST_CHAIN_PORTS[index]}`],
+      });
+      return acc;
+    }, {} as SourcifyChainMap),
+  });
+  const sandbox = Sinon.createSandbox();
+
+  afterEach(() => {
+    sandbox.restore();
+  });
```
manuelwedler: I don't see the sandbox being used in the tests. So, no need to have it.

services/server/test/integration/apiv2/lookup.spec.ts
```diff
@@ -14,9 +14,93 @@ import fs from "fs";
...
+    );
+    const responseMatches = res.body.results.map((result: any) => result.match);
+    const responseAddresses = res.body.results.map(
+      (result: any) => result.address,
+    );
+    TEST_CHAIN_IDs.forEach((chainId, index) => {
+      chai.expect(responseChainIds).to.include(chainId.toString());
+      chai.expect(responseMatches[index]).to.equal("exact_match");
+      chai.expect(responseAddresses[index]).to.equal(addresses[0]);
+    });
```
manuelwedler: IMO, you can assert all fields of the individual objects here.

manuelwedler: Maybe just sort the `res.body.results` by chainId and iterate only once over it.

services/server/src/apiv2.yaml
```diff
@@ -147,6 +147,71 @@ paths:
...
+                        address: '0xDFEBAd708F803af22e81044aD228Ff77C83C935c'
+                        verifiedAt: '2024-07-24T12:00:00Z'
+                        matchId: '421'
+                      - match: match
+                        creationMatch: match
+                        runtimeMatch: match
+                        chainId: '11155111'
+                        address: '0xDFEBAd708F803af22e81044aD228Ff77C83C935c'
+                        verifiedAt: '2025-02-10T14:45:00Z'
+                        matchId: '5531'
```
manuelwedler, line 194:
```suggestion
        '200':
          $ref: '#/components/responses/ListVerifiedContracts'
```

Shouldn't this be enough?

services/server/src/apiv2.yaml
```diff
@@ -147,6 +147,71 @@ paths:
...
+        '400':
+          $ref: '#/components/responses/BadRequestResponse'
+        '404':
+          description: Not Found
```
manuelwedler:
```suggestion
          description: No verified contract found on any chain for the provided address
```

Maybe worth to be more explicit here.

manuelwedler (changes requested):
Looks mostly great :) just a few minor things. Nicely tested with the setup of multiple chains.

> The endpoint returns an array of VerifiedContractMinimal under the results field on the response body. If there are no contracts at this address on any chain, it returns 404 with an results array that is empty []. Idk if this is good practice and we should return 200?

It seems fine as it is to me. It's consistent, since we are also returning a 404 on the endpoint for fetching a single contract. 

The db query looks also great by the way, I tested it.

services/server/src/server/apiv2/lookup/lookup.routes.ts
kuzdogan: I'm fine with that too but I like being more explicit in general, as long as it doesn't get too long. What deos @marcocastignoli think?

marcocastignoli: I really don't have an opinion here. Both options are nice: `allChains` works because it's clear; `all` works because it's clean (fine if documented)

manuelwedler: Okay, I'm also fine with "all chains". But I would avoid camel case in the path. Let's rather make it `all-chains`

services/server/src/apiv2.yaml
kuzdogan: See the comment above 

`# Not using ref:ListVerifiedContracts because its example has different addresses in results. Here we want to have the same address in results. The definitions are the same.`

services/server/test/integration/apiv2/lookup.spec.ts
kuzdogan: Just learned `to.include.members()` that removes the need to sort.

services/server/src/apiv2.yaml
manuelwedler: Ah didn't read it :+1:

PR #2188, opened 2025-06-10
Update server and database docs

### LINKED ISSUE #1983: Review setting up Sourcify DB docs
> As a general comment, I had some difficulty setting up the Sourcify database backend. As someone who isn't all that familiar with Postgres, it took me a while to figure out I had to set up the user, database, and change the authentication for the public schema myself. Some steps for doing this setup before running the migration scripts would be nice to have in the docs (though this was for a fresh instance and didn't do any actual migration).

Quoting from the feedback from the public chat, we should review how to set up the DB from scratch, and potentially modify the migrations to set everything up.

kuzdogan:
Fixes #1983


README.md
```diff
@@ -27,7 +27,10 @@ Sourcify mainly consists of:
 - Packages:
   - [@ethereum-sourcify/lib-sourcify](/packages/lib-sourcify/): The core library for Sourcify. It contains the logic to verify contracts.
   - [@ethereum-sourcify/bytecode-utils](/packages/bytecode-utils/): A library to extract and parse the CBOR encoded metadata from the bytecode.
+  - [@ethereum-sourcify/compilers](/packages/compilers/): A wrapper around Solidity and Vyper compilers to download the right version and invoke the compilation with a common interface.
+  - [@ethereum-sourcify/compilers-types](/packages/compilers-types/): TypeScript types for the compilers.
 - [Sourcify UI](https://github.com/sourcifyeth/ui) - a web UI to interact with the server, lookup, and verify contracts
+- [repo.sourcify.dev](https://github.com/sourcifyeth/repo.sourcify.dev) - a web UI to browse and display verified contract information.
```
manuelwedler, line 34: Should we mention also the database here?

services/server/.env.dev
```diff
@@ -1,5 +1,17 @@
...
+# Sourcify's database postgres connection
+SOURCIFY_POSTGRES_HOST="localhost"
+SOURCIFY_POSTGRES_DB="sourcify"
+SOURCIFY_POSTGRES_USER="sourcify"
+SOURCIFY_POSTGRES_PASSWORD="sourcify"
+SOURCIFY_POSTGRES_PORT=5432
+# SOURCIFY_POSTGRES_SCHEMA=custom-schema # by default "public" is used
+
 ###########################
 ### Public env variables ##
```
manuelwedler: Maybe give a note that this section is optional

services/server/.env.dev
```diff
@@ -27,14 +39,6 @@ CONCURRENT_VERIFICATIONS_PER_WORKER=
 ## Private env variables ##
```
manuelwedler: Also here, it could be more clear that these are optional

services/server/README.md
```diff
@@ -1,54 +1,135 @@
...
 
 ### Prerequisites
 
-- [Node.js](https://nodejs.org/en/) (v22)
-- (Optional) Postgres (See [Choosing the storage backend](#choosing-the-storage-backend))
-- (Optional) Docker and docker-compose for tests
+- [Node.js](https://nodejs.org/en/) (recommended v22)
+- Postgres 16 (or Docker)
 
 ## Quick Start
```
manuelwedler: It is not clear to me whether this section targets the development or production case. To my understanding, it tries to do both. This seems fine to me, but we should be more clear about the options for the database (see my other comment).

Anyway, how about renaming this section to something like "Running for development or production"? I think it would be nicer to have the Docker compose setup as the "Quick Start" option.

services/server/README.md
```diff
@@ -1,54 +1,135 @@
...
 
+### 10. Test the server
+
+```bash
+curl http://localhost:5555/health
+```
+
+You should see `Alive and kicking!` in the response.
+
+### Instant Run with Docker Compose
```
manuelwedler: What do you think about moving this section to the top and calling it "Quick Start"?

It's really the easiest way to try out running Sourcify, and the other option doesn't feel very quick.

services/server/README.md
```diff
@@ -1,54 +1,135 @@
...
+
+There is a docker compose file which makes running the server locally easy.
+
+Keep in mind this is not recommended for production use. You should run a production instance of a Postgres database, add your user, run the migrations, and then run the server.
+
+You should change the chains you want to support in the `sourcify-chains-example.json` file (see [Chains Config](#chains-config)) and the server's `.env.docker` file with the required and optional values (see `.env.dev` file).
+
```
manuelwedler: Maybe worth to add a note that this pulls the public image, and therefore is not suitable for development.

services/server/README.md
manuelwedler: Also, can we add the "Prerequisites" from above as a subsection of this section?

services/server/README.md
```diff
@@ -1,54 +1,135 @@
...
 
-module.exports = {
-  storage: {
-    read: RWStorageIdentifiers.RepositoryV1,
-    writeOrWarn: [],
-    writeOrErr: [RWStorageIdentifiers.RepositoryV1],
-  },
-};
+### 3. Spin up a PostgreSQL database
+
```
manuelwedler, line 62: How about giving two options here as subsubsections: docker can be used for development, dedicated postgres instance for production

services/server/README.md
```diff
@@ -1,54 +1,135 @@
...
 
-3. Build the monorepo's packages
+Copy the `.env.template` file into a file named `.env`. Change values if they are different for your Postgres instance or use those defaults.
 
 ```bash
-npx lerna run build
+cp .env.template .env
+```
+
+Run Postgres with docker compose (note that this will not have a `postgres` root user. For production, you should run a Postgres instance and then add your own user).
```
manuelwedler: In the linked issue (https://github.com/ethereum/sourcify/issues/1983), the user specifically asked to give instructions of how to set up postgres, the database, the user and give read and write access to the user. Maybe we should give a few commands to explain how to do this.

services/server/README.md
```diff
@@ -1,54 +1,135 @@
...
+cd services/database
 ```
 
-3. Build the monorepo's packages
+Copy the `.env.template` file into a file named `.env`. Change values if they are different for your Postgres instance or use those defaults.
 
 ```bash
-npx lerna run build
+cp .env.template .env
+```
```
manuelwedler: I would move this part to the `Run the migrations` section

services/server/README.md
```diff
@@ -178,14 +262,9 @@ A full example of a chain entry is as follows:
         ]
       },
       {
-        "type": "APIKeyRPC", // Alchemy RPCs
+        "type": "APIKeyRPC", // Alchemy RPCs, or any other RPC that requires an API key. ${API_KEY} will be replaced with the value of the env var.
```
manuelwedler, line 285: It is `{API_KEY}` without `$`

kuzdogan:
Incorporated the requested changes, can you have a look @manuelwedler

Issue #2166, opened 2025-05-21
New re-verify endpoint specifications


marcocastignoli:
We often need to fix contracts data due to misaligned data. Currently, we lack a generic method to replace an existing misaligned contract in our database, so we’ve been forced to create custom scripts to address specific issues (#1974).

In this issue, I propose introducing a new private endpoint—callable only by the Sourcify instance maintainer—that allows replacing an existing contract. To improve the performance of this endpoint, we can optimize it by skipping compilation and RPC-based on-chain data fetching, instead relying on information already stored in the database. Two properties, getCompilationFromDatabase and getOnchainDataFromDatabase, can be used to specify that the data should be retrieved from the database.

### Parameters
- **chainId** 
- **address** 
- **creation_txhash**
- **getCompilationFromDatabase: bool**. If set to true it skips compilation and queries compilation information from the database
- **[if(getCompilationFromDatabase==false)] jsonInput**: optionally pass jsonInput, if it's passed it uses the passed one. if it's not passed it uses the jsonInput stored in the database (but it will recompile to get the jsonOutput)
- **getOnchainDataFromDatabase: bool**. If set to true it skips calling rpc to find onchain information and queries onchain information from the database

### Process
1. fetch onchain information from the database if specified in the request's parameter, otherwise call the rpc
2. fetch compiled information from the database if specified in the request's parameter, otherwise compile the contract
3. call lib-sourcify verification flow
4. if successful  
  4.1. delete existing `sourcify_match` and all linked rows in foreign tables, only if those rows are dangling. If there are non-dangling rows then return an error and we manually have to evaluate the case.
  4.2. insert the new verification
5. if not successful, return an error

### lib-sourcify refactoring

[Suggestion](https://github.com/ethereum/sourcify/issues/1990#issuecomment-2835310003) from @manuelwedler: we could implement a new class in lib-sourcify `DatabaseCompilation` that follows the `AbstractCompilation` interface and gets the compiled bytecode and compilation artifacts as constructor arguments. That could be useful for skipping the compilation and avoiding the [monkey patching we do at the current `upgradeContract`](https://github.com/ethereum/sourcify/blob/a682fc6842ff603bc347309c8a1898a4b5da1fdc/services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts#L298-L316) endpoint.

We an potentially do something similar for `SourcifyChain`.

manuelwedler:
[...]

PR #2189, opened 2025-06-10
Add `replace-contract` endpoint

### LINKED ISSUE #1990: Contract's onchain creation bytecode is equal to onchain runtime bytecode
Looking for contracts for the UI I again encountered a weird response. In this case, the contract's "onchain creation bytecode" is the same as its "onchain runtime bytecode". I first thought this was a UI error but looking into the API response and the DB I could confirm it:

https://sourcify.dev/server/v2/contract/8453/0xB1A868BE8ABCB386400820355B3D6944AAD53224?fields=all

```
creationBytecode	
- onchainBytecode	"0x6080604052348015610010…664736f6c63430008140033" <-----
- recompiledBytecode	"0x6101206040523480156200…664736f6c63430008140033"

runtimeBytecode	
- onchainBytecode	"0x6080604052348015610010…664736f6c63430008140033" <----
- recompiledBytecode	"0x6080604052348015610010…664736f6c63430008140033"
```

The hashes are the same from the DB response

<details>
<summary>Query</summary>

```sql

-- Get runtime and creation bytecodes by address
SELECT 
    cd.chain_id,
    cd.address,
    -- rec_runtime.code AS rec_runtime_bytecode,
    rec_runtime.code_hash as rec_runtime_hash,
    -- onc_runtime.code as onc_runtime_bytecode,
    onc_runtime.code_hash as onc_runtime_hash,
    -- onc_creation.code as onc_creation_bytecode,
    onc_creation.code_hash as onc_creation_hash,
    -- rec_creation.code AS rec_creation_bytecode,
    rec_creation.code_hash as rec_creation_hash
FROM contract_deployments cd
JOIN verified_contracts vc ON cd.id = vc.deployment_id
JOIN compiled_contracts cc ON vc.compilation_id = cc.id
join contracts on cd.contract_id = contracts.id 
JOIN code rec_creation ON cc.creation_code_hash = rec_creation.code_hash
JOIN code rec_runtime ON cc.runtime_code_hash = rec_runtime.code_hash
join code onc_creation on contracts.creation_code_hash = onc_creation.code_hash 
join code onc_runtime on contracts.runtime_code_hash = onc_runtime.code_hash
where cd.address=decode('B1A868BE8ABCB386400820355B3D6944AAD53224', 'hex');
```
</details>

|chain_id|address|rec_runtime_hash|onc_runtime_hash|onc_creation_hash|rec_creation_hash|
|--------|-------|----------------|----------------|-----------------|-----------------|
|8453| B1A868BE8ABCB386400820355B3D6944AAD53224 | 523108545387D3A34F2AF40A0121021492926C217EBBA28F46C4BDFDDE7CB85C | A1487C7C0FE446096BA42CDD8092CC7DA67F6AE87BC4A33D173CE1E8F48C0F58 | A1487C7C0FE446096BA42CDD8092CC7DA67F6AE87BC4A33D173CE1E8F48C0F58 | 549F17E7C5247EC58C1E05FB4818E888F1CD21756C2325257130101AF2B6D152 |

I wasn't able to reproduce this after importing from Etherscan as can be seen from the staging response: https://staging.sourcify.dev/server/v2/contract/8453/0xB1A868BE8ABCB386400820355B3D6944AAD53224?fields=all

We should further investigate this as I can see there are more contracts that have their "onchain creation code" == "onchain runtime code":

<details>
<summary>Query</summary>

```sql
SELECT COUNT(*) AS matching_contracts_count
FROM contracts
WHERE creation_code_hash = runtime_code_hash;
```
</details>

Result: 76609 `contracts` entries

Grouping by chain ie. contract_deployments entries:

<details>
<summary>Query</summary>

```sql
SELECT cd.chain_id, COUNT(*) AS contract_count
FROM contracts c
JOIN contract_deployments cd ON c.id = cd.contract_id
WHERE c.creation_code_hash = c.runtime_code_hash
GROUP BY cd.chain_id
ORDER BY contract_count DESC;
```
</details>

|chain_id|contract_count|
|--------|--------------|
|8453|333492|
|80001|207305|
|5|152758|
|4|124177|
|3|76503|
|84531|49830|
|421613|20584|
|11155111|14917|
|69|11221|
|1|5813|
|420|5238|
|17000|4927|
|421611|4212|
|100|1545|
|77|1149|
|28|414|
|40|145|
|41|42|
|8217|28|
|420666|28|
|10200|19|
|99|15|
|167005|14|
|1101|11|
|534|11|
|167006|7|
|356256156|6|
|78430|4|
|78431|3|
|369|1|
|103090|1|

marcocastignoli:
The new `/private/replace-contract`endpoint was originally conceived to fix #1990, then specified in #2166.

This PR:
- Adds `PreRunCompilation` and `DatabaseCompilation`
- Adds the new `/private/replace-contract endpoint`
- Adds tests for the new `/private/replace-contract endpoint`


packages/lib-sourcify/src/Compilation/PreRunCompilation.ts
```diff
@@ -0,0 +1,158 @@
...
+    language: 'solidity' | 'vyper',
+    public compilerVersion: string,
+    jsonInput: SolidityJsonInput | VyperJsonInput,
+    jsonOutput: SolidityOutput | VyperOutput,
+    public compilationTarget: CompilationTarget,
+    public _creationBytecodeCborAuxdata: CompiledContractCborAuxdata,
+    public _runtimeBytecodeCborAuxdata: CompiledContractCborAuxdata,
+  ) {
+    super(jsonInput);
+    this.compilerOutput = jsonOutput as SolidityOutput;
```
manuelwedler:
```suggestion
    this.compilerOutput = jsonOutput;
```

type assertion should not be needed, no?

packages/lib-sourcify/src/Compilation/PreRunCompilation.ts
```diff
@@ -0,0 +1,158 @@
...
+
+export class PreRunCompilation extends AbstractCompilation {
+  public auxdataStyle: AuxdataStyle;
+  // Vyper version is not semver compliant, so we need to handle it differently
+  public compilerVersionCompatibleWithSemver?: string;
+  public language: CompilationLanguage;
+
+  public constructor(
+    public compiler: ISolidityCompiler | IVyperCompiler,
+    language: 'solidity' | 'vyper',
```
manuelwedler: I think we can remove the language parameter and use the language property of the json input.

packages/lib-sourcify/src/Compilation/PreRunCompilation.ts
```diff
@@ -0,0 +1,158 @@
...
+            this.compilerVersionCompatibleWithSemver = `${this.compilerVersion
+              .split('+')[0]
+              .replace(
+                /(b\d+|rc\d+)$/,
+                '',
+              )}+${this.compilerVersion.split('+')[1]}`;
+          } else {
+            throw new CompilationError({ code: 'invalid_compiler_version' });
+          }
+        }
```
manuelwedler: This is duplicated from VyperCompilation. I think it would be better to move the logic in a common function.

packages/lib-sourcify/src/Compilation/PreRunCompilation.ts
```diff
@@ -0,0 +1,158 @@
...
+        // Vyper version support for auxdata is different for each version
+        if (semver.lt(this.compilerVersionCompatibleWithSemver, '0.3.5')) {
+          this.auxdataStyle = AuxdataStyle.VYPER_LT_0_3_5;
+        } else if (
+          semver.lt(this.compilerVersionCompatibleWithSemver, '0.3.10')
+        ) {
+          this.auxdataStyle = AuxdataStyle.VYPER_LT_0_3_10;
+        } else {
+          this.auxdataStyle = AuxdataStyle.VYPER;
+        }
```
manuelwedler: Also this logic could be moved to a common function.

services/server/src/server/services/utils/Database.ts
```diff
@@ -919,4 +993,118 @@ ${
...
+        WHERE cd.chain_id = $1
+          AND cd.address   = $2
+          AND cd.transaction_hash = $3
+        LIMIT 1;
+        `,
+      [chainId, addressBytes, txHashBytes],
+    );
+
+    if (rows.length === 0) {
+      throw new NotFoundError("No existing verified contract found to delete");
```
manuelwedler: I think we should not throw API v1 errors from this class. If for any reason we want to use this method in a v2 style API, the error would be wrongly formatted.

services/server/src/server/apiv1/verification/private/stateless/private.stateless.paths.yaml
```diff
@@ -149,33 +151,84 @@ paths:
...
+                  example: "1"
+                address:
+                  type: string
+                  format: address
+                  description: The contract address
+                  example: "0x00000000219ab540356cBB839Cbe05303d7705Fa"
+                transactionHash:
+                  type: string
+                  description: The creation transaction hash
+                  example: "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"
```
manuelwedler, line 175: Biggest question here... Why do we base the API on the chainId + address and not on the sourcifyMatchId?

Why do we need the transaction hash?

services/server/src/server/apiv1/verification/private/stateless/private.stateless.paths.yaml
```diff
@@ -149,33 +151,84 @@ paths:
...
+                  description: If true, fetches the contract's information from the RPC
+                  example: false
+                jsonInput:
+                  type: object
+                  description: Required if forceCompilation is true. The JSON input to use for recompilation
+                compilerVersion:
+                  type: string
+                  description: Required if forceCompilation is true. The compiler version to use for recompilation
+                  example: "0.8.19+commit.7dd6d404"
+                compilationTarget:
```
manuelwedler, line 191: I have a tendency to say that we should use the same format for compilationTarget accross our APIs. In API v2, we use contractIdentifier which is {name}:{path}. Do you think it makes sense to also apply the format here? (We have an utility function `splitFullyQualifiedName` to extract name and path from it)

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -135,212 +144,126 @@ export async function verifyDeprecated(
...
- * Then we verify the contract again with the new mocked compilation and sourcifyChain objects.
- * Finally we upgrade the misaligned contract by UPDATING the verified_contracts and sourcify_matches creation fields.
+ * @param {string|number} req.body.chainId - The chain ID of the contract.
+ * @param {string} req.body.address - The contract address.
+ * @param {string} req.body.transactionHash - The creation transaction hash.
+ * @param {boolean} req.body.forceCompilation - If true, recompiles the contract using the provided jsonInput, compilerVersion and compilationTarget.
+ * @param {boolean} req.body.forceRPCRequest - If true, fetches the contract's information from the RPC.
+ * @param {object} [req.body.jsonInput] - If forceCompilation is true, provide jsonInput to use for recompilation.
+ * @param {string} [req.body.compilerVersion] - If forceCompilation is true, provide the compiler version to use for recompilation.
+ * @param {object} [req.body.compilationTarget] - If forceCompilation is true, provide the compilation target to use for recompilation.
```
manuelwedler: Not sure if we should duplicate the parameter descriptions here. It should be sufficient to have them in the openapi definition.

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -135,212 +144,126 @@ export async function verifyDeprecated(
...
-
+  // Get the connection pool from SourcifyDatabaseService
+  const sourcifyDatabaseService = services.storage.rwServices[
+    "SourcifyDatabase"
+  ] as SourcifyDatabaseService;
   if (!sourcifyDatabaseService) {
     return res
       .status(StatusCodes.INTERNAL_SERVER_ERROR)
       .send({ error: "Database service not available" });
   }
```
manuelwedler: To be consistent, it would be better to throw an `InternalServerError` here.

manuelwedler: and not send the status manually

services/server/src/server/services/utils/database-util.ts
```diff
@@ -792,3 +794,26 @@ export function prepareCompilerSettingsFromVerification(
...
+    await client.query("BEGIN");
+    const result = await callback(client);
+    await client.query("COMMIT");
+    return result;
+  } catch (error) {
+    await client.query("ROLLBACK");
+    if (errorCallback) {
+      errorCallback(error);
+    }
+    throw error;
```
manuelwedler: I think we can remove the errorCallback. The error is already thrown. We can just catch it outside this function

services/server/src/server/services/utils/database-util.ts
```diff
@@ -792,3 +794,26 @@ export function prepareCompilerSettingsFromVerification(
     verification.compilation.jsonInput.settings;
   return restSettings;
 }
+
+// Transaction wrapper utility
+export async function withTransaction<T>(
+  database: Database,
```
manuelwedler: If I'm not mistaken, this function could simply be a method of the `AbstractDatabaseService`. Then, we would not need to provide the `database` parameter to it.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -864,73 +866,122 @@ export class SourcifyDatabaseService
...
+  }
+
+  // Override this method to include the SourcifyMatch
+  async storeVerification(
+    verification: VerificationExport,
+    jobData?: {
+      verificationId: VerificationJobId;
+      finishTime: Date;
+    },
+    alternativePoolClient?: PoolClient,
```
manuelwedler: Instead of providing the pool client as parameter here, just call the `storeVerificationWithPoolClient` inside the replaceContract handler directly.

services/server/src/server/services/utils/Database.ts
```diff
@@ -252,13 +258,77 @@ ${
...
+        JOIN compiled_contracts ON compiled_contracts.id = verified_contracts.compilation_id
+        JOIN sourcify_matches ON sourcify_matches.verified_contract_id = verified_contracts.id
+        JOIN code compiled_creation_code ON compiled_contracts.creation_code_hash = compiled_creation_code.code_hash 
+        JOIN code compiled_runtime_code ON compiled_contracts.runtime_code_hash = compiled_runtime_code.code_hash
+        JOIN contract_deployments ON contract_deployments.id = verified_contracts.deployment_id
+        WHERE 1=1
+        AND contract_deployments.chain_id = $1
+        AND contract_deployments.address = $2
+        AND contract_deployments.transaction_hash = $3`,
+      [chainId, address, transactionHash],
```
manuelwedler: Why is the transaction hash part of the query? Theoretically, there should only be one sourcify match per chain + address. The join on the sourcify_matches table would eliminate the second row if there are two entries for the same chain + address but with different transaction hashes.

services/server/src/server/services/utils/Database.ts
```diff
@@ -252,13 +258,77 @@ ${
...
-    runtime_match,
-    creation_match,
-    metadata,
-  }: Omit<Tables.SourcifyMatch, "created_at" | "id">) {
-    await this.pool.query(
+  async getVerifiedContractFromDeployment(
+    chainId: number,
+    address: Bytes,
+    transactionHash: Bytes,
+  ) {
```
manuelwedler: Can we have a return type?

services/server/src/server/services/utils/Database.ts
```diff
@@ -252,13 +258,77 @@ ${
...
+        AND contract_deployments.transaction_hash = $3`,
+      [chainId, address, transactionHash],
+    );
+  }
+
+  async getContractDeploymentInfo(
+    chainId: number,
+    address: Bytes,
+    transactionHash: Bytes,
+  ) {
```
manuelwedler: also here a return type would be nice

services/server/src/server/services/utils/Database.ts
```diff
@@ -919,4 +993,118 @@ ${
...
+          ctr.runtime_code_hash   AS contract_runtime_code_hash,
+          cc.creation_code_hash   AS compilation_creation_code_hash,
+          cc.runtime_code_hash    AS compilation_runtime_code_hash
+        FROM ${this.schema}.verified_contracts vc
+        JOIN ${this.schema}.contract_deployments cd ON cd.id = vc.deployment_id
+        JOIN ${this.schema}.contracts ctr          ON ctr.id = cd.contract_id
+        JOIN ${this.schema}.compiled_contracts cc  ON cc.id = vc.compilation_id
+        WHERE cd.chain_id = $1
+          AND cd.address   = $2
+          AND cd.transaction_hash = $3
```
manuelwedler: contract_id is also part of the unique constraint on contract_deployments. Wouldn't it be possible that this query returns multiple rows for that reason?

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -351,69 +274,53 @@ export async function upgradeContract(
...
-        creationMetadataMatch,
-        verifiedContract.id,
-      ],
-    );
+    // If the new verification leads to a non-match, we can't replace the contract
+    if (!runtimeMatch || !creationMatch) {
+      throw new BadRequestError(
+        "The contract is not verified or the verification is not perfect",
+      );
+    }
```
manuelwedler: I worry a bit that if the forceRPCRequest is used and we don't check that the new verification also was able to fetch creation information, we might store worse matches than before here. Should we add some checks for that too?

manuelwedler: So maybe it would be nice to add some checks here

manuelwedler: Also the error message is a bit confusing here. We don't expect the verification to be perfect for this endpoint, no?

services/server/src/server/services/utils/Database.ts
```diff
@@ -919,4 +993,118 @@ ${
...
+      throw new NotFoundError("No existing verified contract found to delete");
+    }
+
+    const info = rows[0];
+
+    // 2. Child-first deletions relying on FK safety
+    await poolClient.query(
+      `DELETE FROM ${this.schema}.sourcify_matches WHERE verified_contract_id = $1`,
+      [info.verified_contract_id],
+    );
```
manuelwedler, line 1008: Is it possible to keep the sourcify match? We could rather update it's properties. I somehow feel like it would be good to keep the same match id.

services/server/src/server/services/utils/Database.ts
```diff
@@ -919,4 +993,118 @@ ${
...
+    await poolClient.query(
+      `DELETE FROM ${this.schema}.verification_jobs WHERE verified_contract_id = $1`,
+      [info.verified_contract_id],
+    );
```
manuelwedler: I still think it would be better to not delete the job. We can change the constraint, or set the verified_contract_id on the job to null

services/server/src/server/apiv1/verification/private/stateless/private.stateless.paths.yaml
```diff
@@ -149,33 +151,84 @@ paths:
...
+                  example: "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"
+                forceCompilation:
+                  type: boolean
+                  description: If true, recompiles the contract using the provided jsonInput, compilerVersion and compilationTarget
+                  example: false
+                forceRPCRequest:
+                  type: boolean
+                  description: If true, fetches the contract's information from the RPC
+                  example: false
+                jsonInput:
```
manuelwedler, line 184: I'm also wondering regarding the recompilation functionality. What's the use case for providing new std json input? Where would we get it from? Would it also be a use case to recompile using the json input that we already have in the db?

services/server/src/server/services/utils/DatabaseCompilation.ts
```diff
@@ -0,0 +1,175 @@
...
+  SolidityJsonInput,
+  SolidityOutput,
+  VyperJsonInput,
+  VyperOutput,
+} from "@ethereum-sourcify/lib-sourcify";
+import { bytesFromString } from "./database-util";
+import { Database } from "./Database";
+import logger from "../../../common/logger";
+
+export class DatabaseCompilation {
```
manuelwedler: I don't like the name `DatabaseCompilation` too much. Actually, I think it would be better to just have a function for this. There is no need to keep state. You create a DatabaseCompilation instance, call one function on it to generate the PreRunCompilation and never use the DatabaseCompilation instane again. This is typically a sign that a function would better fit the use case. I'd say just make a function `extractCompilationFromDatabase`.

services/server/src/server/services/utils/DatabaseCompilation.ts
```diff
@@ -0,0 +1,175 @@
...
+      logger.error("DatabaseCompilation: database pool not initialized");
+      throw new Error("DatabaseCompilation: database pool not initialized");
+    }
+  }
+
+  async extractCompilationProperties() {
+    try {
+      // Fetch compilation data from the database
+      const verifiedContractResult =
+        await this.database.getVerifiedContractFromDeployment(
```
manuelwedler: You could also use `STORED_PROPERTIES_TO_SELECTORS["std_json_input"]` and `STORED_PROPERTIES_TO_SELECTORS["std_json_output"]` from database-util in order to get the data directly from the db and not generate it here.

manuelwedler: Also for all the other properties. This would reduce the code there to a minimum.

services/server/src/server/services/utils/Database.ts
```diff
@@ -919,4 +993,118 @@ ${
...
+             AND NOT EXISTS (
+               SELECT 1 FROM ${this.schema}.compiled_contracts_sources WHERE source_hash = $1
+             )`,
+        [source_hash],
+      );
+    }
+
+    // 4. Deployment side clean-up
+    await poolClient.query(
+      `DELETE FROM ${this.schema}.contract_deployments WHERE id = $1`,
```
manuelwedler, line 1041: Isn't it also possible that a contract deployment is referenced by more than one verified_contract?

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -135,212 +144,126 @@ export async function verifyDeprecated(
...
+    let sourcifyChain: SourcifyChain;
+    if (!forceRPCRequest) {
+      // Create a SourcifyChainMock object filled with data from the database
+      sourcifyChain = new SourcifyChainMock(
+        sourcifyDatabaseService.database,
+        chainId,
+        address,
+        transactionHash,
+      );
+      await (sourcifyChain as SourcifyChainMock).init();
```
manuelwedler: instead of first creating an instance, then calling init, you can just have a single factory method `async createSourcifyChainMock` that returns an instance of the class.

services/server/test/integration/apiv1/verification-handlers/private.stateless.spec.ts
```diff
@@ -0,0 +1,574 @@
...
+  it("should replace contract using existing database compilation (forceCompilation: false) and restore creation_match", async () => {
+    // First, verify with perfect match
+    const res = await chai
+      .request(serverFixture.server.app)
+      .post("/")
+      .field("address", chainFixture.defaultContractAddress)
+      .field("chain", chainFixture.chainId)
+      .field("creatorTxHash", chainFixture.defaultContractCreatorTx)
+      .attach("files", chainFixture.defaultContractMetadata, "metadata.json")
+      .attach("files", chainFixture.defaultContractSource);
```
manuelwedler: Why not use the `verifyContract` utility from helpers.ts here and in the other tests?

services/server/src/server/services/utils/database-util.ts
marcocastignoli: that's a good idea :)

services/server/src/server/services/utils/Database.ts
marcocastignoli: We decided that it's easier to DELETE and INSERT instead of UPDATING because we would have conflict in the `storeVerification` function if a match is already present in the table.

services/server/src/server/apiv1/verification/private/stateless/private.stateless.paths.yaml
marcocastignoli: `stdJsonInput` is stored in the database split into multiple tables, the field needs to be passed if we need to fix a case in which the stdJsonInput fields were stored wrongly in the database

services/server/src/server/apiv1/verification/private/stateless/private.stateless.paths.yaml
marcocastignoli: We decided to use chainid + address to follow the rest of our APIs. We dropped transaction_hash to be mandatory

services/server/src/server/services/utils/Database.ts
marcocastignoli: If a delete query fails, the entire upgrade fails. like defined in #2166

> 4.1. delete existing sourcify_match and all linked rows in foreign tables, only if those rows are dangling. If there are non-dangling rows then return an error and we manually have to evaluate the case.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -864,73 +865,112 @@ export class SourcifyDatabaseService
...
+          verification,
+          jobData,
+        );
+      });
+    } catch (error: any) {
+      logger.error("Error storing verification", {
+        error: error,
       });
+      throw error;
     }
```
manuelwedler:
```suggestion
    await this.withTransaction(async (transactionPoolClient) => {
      await this.storeVerificationWithPoolClient(
        transactionPoolClient,
        verification,
        jobData,
      );
    });
```

The log is duplicated from `storeVerificationWithPoolClient`

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -128,219 +142,137 @@ export async function verifyDeprecated(
...
-      sources[source.path] = { content: source.content };
+    let sourcifyChain: SourcifyChain;
+    if (!forceRPCRequest) {
+      // Create a SourcifyChainMock object filled with data from the database
+      sourcifyChain = await SourcifyChainMock.create(
+        sourcifyDatabaseService.database,
+        chainId,
+        address,
+      );
+      let transactionHashFromDatabase = (sourcifyChain as SourcifyChainMock)
```
manuelwedler:
```suggestion
      const transactionHashFromDatabase = (sourcifyChain as SourcifyChainMock)
```

I'm getting a linter error here

services/server/src/server/services/utils/DatabaseCompilation.ts
```diff
@@ -0,0 +1,114 @@
...
+      await database.getSourcifyMatchByChainAddressWithProperties(
+        chainId,
+        bytesFromString(address),
+        [
+          "std_json_input",
+          "std_json_output",
+          "runtime_cbor_auxdata",
+          "creation_cbor_auxdata",
+          "metadata",
+          "compiler",
```
manuelwedler: I think you need to query `version` to get the compiler version. `compiler` should just be "solc" or "vyper".

services/server/src/server/services/utils/DatabaseCompilation.ts
```diff
@@ -0,0 +1,114 @@
+import {
```
manuelwedler: Can we also rename this file? Or just move the function to `database-util.ts`?

services/server/src/server/services/utils/DatabaseCompilation.ts
```diff
@@ -0,0 +1,114 @@
...
+    const verifiedContractResult =
+      await database.getSourcifyMatchByChainAddressWithProperties(
+        chainId,
+        bytesFromString(address),
+        [
+          "std_json_input",
+          "std_json_output",
+          "runtime_cbor_auxdata",
+          "creation_cbor_auxdata",
+          "metadata",
```
manuelwedler: instead of querying metadata, just get the contract identifier by querying `fully_qualified_name`

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -351,69 +283,52 @@ export async function upgradeContract(
...
-        message: "Contract upgrade successful",
-        verifiedContractId: verifiedContract.id,
-        sourcifyMatchId: sourcifyMatchId,
-      },
+      replaced: true,
+      address: address,
+      chainId: chainId,
+      transactionHash: transactionHash,
+      newStatus: verificationStatus,
+      rpcFailedFetchingCreationBytecode:
```
manuelwedler, line 328: this still needs to be added in the spec

manuelwedler: I mean the parameter rpcFailedFetchingCreationBytecode

services/server/src/server/services/utils/DatabaseCompilation.ts
marcocastignoli: Good catch!

kuzdogan:
One question, does this endpoint also deletes verified contracts? I see a "deleteMatch" function but from the API spec I understood as this is only for updating matches.

If not, we can open another issue for an endpoint to delete a contract, for the future.

Issue #2206, opened 2025-07-01
Investigation of Sourcify downtimes


manuelwedler:
We recently saw a lot of short-lived downtimes. These happened while we were verifying a lot of contracts (~300,000 over one weekend) via API v2. I am trying to summarise here what we learned so far.

### Unresponsive instances

The downtimes seem to happen due to some unresponsive instances. I filtered the logs for one failing instance:
https://console.cloud.google.com/logs/query;query=resource.type%20%3D%20%22cloud_run_revision%22%0Aresource.labels.service_name%20%3D%20%22sourcify-production-server%22%0Aresource.labels.location%20%3D%20%22europe-west1%22%0Aseverity%3E%3DDEFAULT%0Alabels.instanceId%3D%220069c7a98845de1865402b6c676cb1f45a8246a022afd16d8227c0def453a23db5a663f1e53be28c58746e822f35875fb602777dd1e10e25d434430379e6614e356538c8546fb0388365419b1d32af85b431%22;storageScope=project;cursorTimestamp=2025-06-23T11:00:59.056733Z;startTime=2025-06-23T11:00:09.036Z;endTime=2025-06-23T11:01:24.717Z?inv=1&invt=Ab1mbg&project=sourcify-project&pli=1

You can see that the instance behaves normally at beginning but at some point, any request just fails because of:
`The request failed because either the HTTP response was malformed or connection to the instance had an error. Additional troubleshooting documentation can be found at: https://cloud.google.com/run/docs/troubleshooting#malformed-response-or-connection-error`

In the end, it is killed by GCP because of being unresponsive.

To me it looks like the instance runs into a memory issue or maybe with the CPU. The metrics also show quite high CPU usage. The memory is around 50 % which should actually be fine. The bigger question is why GCP doesn't scale the instances up. To me the CPU utilization of above 80% looks like it should actually. According to the docs, GCP tries to target 60 % across instances. Also I wonder why these instances are still served with requests, there should also be some with less load.

One cause for the unresponsiveness could be the `/v2/contracts/` endpoint. Before the crash, this instance received a lot of /contracts/ requests and the very first one with 503 (takes +30sec to respond) is also a /contracts/

### High request latencies

While my batch verification script was running, we saw quite high request latencies:

![Image](https://github.com/user-attachments/assets/29a8d21e-81a4-4ab8-8b74-69533181b117)

### Dangling jobs

[...]

PR #2216, opened 2025-07-08
Verify via SolidityCompilation directly inside /verify/solc-json endpoint

manuelwedler:
See #2206 

The breaking request of #2206 fails on the unzip method as far as I can tell from the logs. The unzip is called with Metadata files that we get from recompiling the contract a first time. The first recompilation via the provided std json works. Since the processing via Metadata is a workaround from legacy code, we can remove this step and directly call the actual compilation for the std json, and verify from there. This should hopefully fix the server for this request, otherwise I added logging to check what the provided parameters and then we can debug further.

### DIFF
--- a/services/server/src/server/apiv1/verification/solc-json/stateless/solc-json.stateless.handlers.ts
+++ b/services/server/src/server/apiv1/verification/solc-json/stateless/solc-json.stateless.handlers.ts
@@ -2,13 +2,15 @@ import { Response, Request } from "express";
 import { extractFiles } from "../../verification.common";
 import {
   ISolidityCompiler,
-  SolidityMetadataContract,
-  createMetadataContractsFromFiles,
+  SolidityCompilation,
+  SolidityJsonInput,
 } from "@ethereum-sourcify/lib-sourcify";
 import { BadRequestError } from "../../../../../common/errors";
 import { Services } from "../../../../services/services";
 import { ChainRepository } from "../../../../../sourcify-chain-repository";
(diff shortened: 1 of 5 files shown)


marcocastignoli (approved):
lgtm 👍

services/server/src/server/apiv1/verification/solc-json/stateless/solc-json.stateless.handlers.ts
```diff
@@ -34,40 +36,38 @@ export async function verifySolcJsonEndpoint(req: Request, res: Response) {
...
-  const metadataContracts = await createMetadataContractsFromFiles(
-    metadataAndSourcesPathBuffers,
-  );
+  logger.debug("Request to /verify/solc-json", {
+    chainId: chain,
+    address: address,
+    contractName: contractName,
+    compilerVersion: compilerVersion,
+    creatorTxHash: creatorTxHash,
+    solcJson: solcJson,
```
kuzdogan, line 47: Just a note, I find it not so useful to print large inputs like the solcJson even in the debug mode. It makes harder to scroll the terminal and make sense of what's going on. If needed in debug mode we can always use breakpoints with the same input.

manuelwedler: By this log, I actually wanted to find the exact json input that caused the server downtimes. Turns out, it was so long that it was truncated by GCP in the end. So indeed not very helpful :/

We can remove the json input again from the log.

PR #2228, opened 2025-07-14
Remove metadata only outputSettings in etherscan-utils

kuzdogan:
This is an old behavior from metadata-based verification times. We used to first compile the etherscan result to get the metadata. Now we generate std-json in these functions and the outputSelection field is always overridden by the SolidityCompilation and VyperCompilation:

https://github.com/ethereum/sourcify/blob/5e362a8707ff10af76673c9ae1c05250dffcb0ff/packages/lib-sourcify/src/Compilation/SolidityCompilation.ts#L48-L68

https://github.com/ethereum/sourcify/blob/5e362a8707ff10af76673c9ae1c05250dffcb0ff/packages/lib-sourcify/src/Compilation/VyperCompilation.ts#L84-L99

### DIFF
--- a/packages/compilers-types/src/SolidityTypes.ts
+++ b/packages/compilers-types/src/SolidityTypes.ts
@@ -88,7 +88,7 @@ export interface SoliditySettings {
   debug?: Debug;
   metadata?: SettingsMetadata;
   libraries?: Libraries;
-  outputSelection: OutputSelection;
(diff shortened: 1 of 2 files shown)


manuelwedler (approved):
thanks!

manuelwedler:
I missed the failing CI, tests need to be adapted too

manuelwedler (approved):

PR #2231, opened 2025-07-15
Add examples to the std-json input + log API key's first letters

kuzdogan:
I found myself having to check the compiler version format and where we put the language field. So adding the examples in the API docs for reference.

Also logs the first letters of Etehrscan API keys to be able to understand which key is in use

### DIFF
--- a/services/server/src/server/services/utils/etherscan-util.ts
+++ b/services/server/src/server/services/utils/etherscan-util.ts
@@ -205,13 +205,15 @@ export const fetchFromEtherscan = async (
       : new BadRequestError(errorMessage);
   }
 
-  const url = `https://api.etherscan.io/v2/api?chainid=${sourcifyChain.chainId}&module=contract&action=getsourcecode&address=${address}&apikey=`;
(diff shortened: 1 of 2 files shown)


services/server/src/server/services/utils/etherscan-util.ts
```diff
@@ -205,13 +205,15 @@ export const fetchFromEtherscan = async (
...
     userApiKey ||
     process.env[sourcifyChain.etherscanApi.apiKeyEnvName || ""] ||
     process.env.ETHERSCAN_API_KEY ||
     "";
   const secretUrl = url + apiKey;
+
+  url = url + apiKey.slice(0, 6) + "...";
   let response;
   logger.debug("Fetching from Etherscan", {
     url,
```
manuelwedler: For being able to check how frequently users provide their own key, we should also change the log level to info.

kuzdogan: Agreed

manuelwedler (approved):

PR #2237, opened 2025-07-17
Disable failing chains, update chains.json

kuzdogan:


kuzdogan:
Ah I always forget them, thanks again.

I tested these locally and they all fail even with no timeout. There has been one chain that eventually succeeds and I didn't turn that off.

kuzdogan:
Actually should I just add a check in the tests to not run if the chain is not supported?

manuelwedler:
> Actually should I just add a check in the tests to not run if the chain is not supported?

Good idea, would be much easier :)

kuzdogan:
Quick look to skipping please @manuelwedler

services/server/test/chains/chain-tests.spec.ts
```diff
@@ -1950,6 +1950,22 @@ describe("Test Supported Chains", function () {
...
         .end(async (err, res) => {
+          // Check if the chain is not supported
+          if (
+            res &&
+            res.body &&
+            res.body.error &&
+            res.body.error.includes(
+              `Chain ${chainId} not supported for verification!`,
+            )
+          ) {
```
manuelwedler: Why don't we check against the `sourcify-chains-default.json` directly and even before sending the request? I think checking against an error message isn't great since they might change.

kuzdogan: My concern there was the `default.json` is only the default and can change e.g. you can pass `--chains sourcifyChains.json` for your custom implementaiton.

A thing of both would be to fetch the `/chains` endpoint first and keep that list.

manuelwedler: True, best would be the response from `/chains`. Just had a look at the test file, and saw we actually fetch it already. So should be easy to reuse it here.

kuzdogan:
Done!

manuelwedler (approved):
Thanks!

PR #2246, opened 2025-07-22
Log client identification headers for verify.sourcify.dev UI requests

kuzdogan:
Turns out GCP does not log custom headers and they need to be logged inside the service

See https://github.com/sourcifyeth/verify.sourcify.dev/commit/0ae11add1df284f487594f40bb0963fc06da845e

### DIFF
--- a/services/server/src/server/server.ts
+++ b/services/server/src/server/server.ts
@@ -210,7 +210,7 @@ export class Server {
       const clientType = req.headers["x-client-type"] as string;
 
       if (clientSource) {
-        logger.info("verify.sourcify.dev UI client request", {
+        logger.info("Client request via header", {
           method: req.method,
           path: req.path,
           clientSource,


services/server/src/server/server.ts
```diff
@@ -203,6 +203,24 @@ export class Server {
       });
     });
 
+    // Log verify.sourcify.dev UI client identification headers if present
+    this.app.use((req, res, next) => {
+      const clientSource = req.headers["x-client-source"] as string;
+      const clientVersion = req.headers["x-client-version"] as string;
+      const clientType = req.headers["x-client-type"] as string;
+
+      if (clientSource) {
```
manuelwedler, line 212:
```suggestion
      if (clientSource === "sourcify-ui") {
```
Shouldn't this be the check? otherwise the log message is not precise.

kuzdogan: Yes. I think the check is correct but the log message is wrong as you said. I'd change the log message wording to just "client request".

services/server/src/server/server.ts
```diff
@@ -203,6 +203,24 @@ export class Server {
...
+        logger.info("verify.sourcify.dev UI client request", {
```
kuzdogan:
```suggestion
        logger.info("Client request via header", {
```

kuzdogan: @manuelwedler feel free to commit if ok

Issue #2123, opened 2025-04-29
Bring back migration scripts to Sourcify database


marcocastignoli:
We used to have migration files to keep track of every change in the sourcify database (see below). We decided that it wasn't that helpful to keep track of every change and that it was better to use the migration to highlight the differences between [VerA schema](https://github.com/verifier-alliance/database-specs/blob/master/database.sql) and [Sourcify schema](https://github.com/ethereum/sourcify/blob/9db1250d2b1126a4f6b645b9681cb54b1c68bdc9/services/database/migrations/20231109160023-sourcify.js).

```
// migration files before our decision to drop migrations
20231109160022-create-alliance-schema.js
20231109160023-sourcify.js
20240624124246-add-metadata.js
20240627084509-code-hash-becomes-sha256.js
20240710103541-add-metadata_match.js
20240715152144-add-transformations-constraint.js
20240715154744-add-timestamp-fields.js
20240716160322-metadata_match-nullable.js
20240729112258-compiled_contracts-constraints.js
```

[Recently we received complains](https://github.com/ethereum/sourcify/pull/1881#issuecomment-2804772648) that now maintaining forks of sourcify became more difficult. We should start again using migrations while keeping in mind our niche use-case in which we have to extend the VerA database.

manuelwedler:
> A nice solution could be starting to use migrations with the same format for both VerA and Sourcify

This is what I actually wanted to propose yesterday 👍 

We could check via a CI test that for any change to the main database.sql the result of the migrations is the same db schema. Or we discontinue to have one main database.sql script, but in my opinion it is quite useful to also have it in one file.
And I think migrations in the VerA repo should be in plain SQL (to not be opinionated about the tooling).

marcocastignoli:
If @kuzdogan doesn't have anything against this solution I would consider this decided :)

kuzdogan:
I agree that the migrations should be in plain SQL for Vera. Maybe even for Sourcify we should do that if it's not too difficult. 

Just, is this urgent or why did we add this to this sprint? Or just for discussion?

marcocastignoli:
We added the discussion to this sprint, the implementation itself is not urgent

PR #2250, opened 2025-07-23
Add new migrations format

manuelwedler:
See #2123 and https://github.com/verifier-alliance/database-specs/issues/31

Changes:

- Adds dbmate as the migration tool as we also will have it in VerA
- Adds npm scripts to run dbmate to set it up depending on the command (we need to merge the two migration folders of sourcify and vera)
- Adds documentation about how to make new schema changes
- Removes db-migrate
- Adds a CI workflow to validate that the migrations work as expected


.circleci/new_branch.yml
```diff
@@ -2,14 +2,23 @@ version: 2.1
...
   node-build:
     jobs:
       - node-v22
+  database-schema-validation:
+    when: << pipeline.parameters.run-validate-database-schema >>
+    jobs:
+      - validate-database-schema
   # Has to run always. Can't regex filter on add-chain-{chainId} branch names see: https://stackoverflow.com/questions/55839004/circleci-regex-filtering-match-within-string
-  test-new-chain:
+  new-chain-tests:
```
marcocastignoli, line 21: Isn't this key the same used below at line 60?

.circleci/config.yml
```diff
@@ -103,9 +103,12 @@ workflows:
...
     jobs:
-      - continuation/continue:
-          name: continue-new-branch
-          configuration_path: .circleci/new_branch.yml
+      - path-filtering/filter:
+          name: check-updated-modules-new-branch
+          mapping: |
+            services/database/.* run-validate-database-schema true
+          base-revision: staging  
+          config-path: .circleci/new_branch.yml
```
marcocastignoli, line 111: Why are we replacing the continue-new-branch job?

marcocastignoli (changes requested):
A couple questions while I continue reviewing

.circleci/config.yml
manuelwedler: We are also continuing from the job here, but can set a flag in case `services/database/` was changed. We only need to run the db validation in that case. CI should still continue in any case on the new_branch workflows.

.circleci/new_branch.yml
manuelwedler: No, in line 60 we define the job. This is the name of the workflow. I just thought it's better to not use the same name for a worflow and a job, because it's confusing as your question already shows.

manuelwedler:
> Seems like `ln -r` is not supported on mac

Oh no, that's an issue...

manuelwedler:
> Seems like `ln -r` is not supported on mac

added the fix

marcocastignoli (approved):
LGTM! Great, I like how migrations are structured now! :)

PR #2257, opened 2025-07-28
Add custom replace method for contract replacement

marcocastignoli:


kuzdogan:
Not sure if this is ready to review but these functions should definitely be tested IMO.

marcocastignoli:
Nono, I just wanted Manu to have a look after our call. I forgot to mention this @manuelwedler

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -183,6 +184,19 @@ export async function replaceContract(
...
+    case "replace-creation-information":
+      customReplaceMethod = replaceCreationInformation;
+      break;
+    case undefined: // Default to the standard replacement method
+      break;
+    default:
+      throw new BadRequestError(
+        `Unknown customReplaceMethod: ${req.body.customReplaceMethod}`,
+      );
+  }
```
manuelwedler: You could create an object like the following inside `customReplaceMethods.ts`:

```ts
export const REPLACE_METHODS = {
  "replace-creation-information": replaceCreationInformation
}
```

Then just use it here like this:

```ts
const customReplaceMethod = REPLACE_METHODS[req.body.customReplaceMethod];
```

A bit easier to extend, IMO.

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
```diff
@@ -0,0 +1,146 @@
+import { VerificationExport } from "@ethereum-sourcify/lib-sourcify";
+import { PoolClient } from "pg";
+import {
+  bytesFromString,
+  getDatabaseColumnsFromVerification,
+} from "../../../../services/utils/database-util";
+
+export async function replaceCreationInformation(
+  poolClient: PoolClient,
+  verification: VerificationExport,
```
manuelwedler, line 10: Maybe it makes sense to check if the VerificationExport successfully fetched creation information before calling this method?

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
```diff
@@ -0,0 +1,146 @@
...
+      [newCreationCodeHash, runtimeCodeHash],
+    );
+    newContractId = existingContractResult.rows[0].id;
+  } else {
+    newContractId = newContractInsertResult.rows[0].id;
+  }
+
+  // Update contract deployment with new creation fields and new contract_id
+  await poolClient.query(
+    `UPDATE contract_deployments
```
manuelwedler: Do we need to set the updated_at column manually or is it done automatically?

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
```diff
@@ -295,22 +309,36 @@ export async function replaceContract(
...
+            await sourcifyDatabaseService.database.deleteMatch(
+              transactionPoolClient,
+              chainId,
+              address,
+            );
+
+            // Insert the new verification information into the database
+            await sourcifyDatabaseService.storeVerificationWithPoolClient(
+              transactionPoolClient,
+              verificationExport,
```
manuelwedler: Imo, this part is completely useless since we will always need to provide a custom methods. The amount of non-dangling rows is just too big. Any reason why you would want to keep it?

manuelwedler (commented):
I like the update approach much better to be honest. Deleting never felt very safe to me, especially since in most cases we only need to fix a few columns. So this approach now looks great to me. 

Potentially, if we now follow the same approach on the compiled_contracts table we could have a "updateAllTables" function (handling sources might be a bit more complicated though). But for now this seems totally enough for our problems.

Obviously tests are still missing.

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
marcocastignoli: It's actually up to discussion. Maybe this can become an additional `customReplaceMethod`, and be the default one? I would keep it, I think it could be helpful to fix standalone contracts.

manuelwedler: I don't believe that we will ever encounter an issue that solely appears on standalone contracts, such that we won't run the deletion logic anyway.  Overall, I think it's much better to have a well functioning update logic (as you started to implement it) than deleting. In the future, we could add a function for updating the compiled_contracts side of the db schema, and then we should be good to fix all data of a contract via updating and not deleting anything.

Maybe we should also get the opinion of @kuzdogan on the general approach of this endpoint. You should at least be aware of this: In order to fix #1990, we now don't delete anything in the db, and instead update the related information on the tables.

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
marcocastignoli: It's done automatically

services/server/src/server/apiv1/verification/private/stateless/private.stateless.paths.yaml
```diff
@@ -191,6 +191,10 @@ paths:
                   type: string
                   description: Required if forceCompilation is true. The compilation target to use for recompilation
                   example: "project:/contracts/Storage.sol:Storage"
+                customReplaceMethod:
+                  type: string
+                  description: Optional. The custom replace method to use for the contract replacement
+                  example: "replace-creation-information"
```
kuzdogan: It took a while for me to understand what a "custom replace method" is. I think this needs to be documented and explained more explicitly here

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
```diff
@@ -0,0 +1,164 @@
...
+      // Code already exists, get the hash
+      const existingCodeResult = await poolClient.query(
+        `SELECT code_hash FROM code WHERE code_hash = digest($1::bytea, 'sha256')`,
+        [databaseColumns.onchainCreationCode.bytecode],
+      );
+      newCreationCodeHash = existingCodeResult.rows[0].code_hash;
+    } else {
+      newCreationCodeHash = creationCodeInsertResult.rows[0].code_hash;
+    }
+  }
```
manuelwedler: If we would pass the `sourcifyDatabaseService` instead of the `poolClient`, we could call `sourcifyDatabaseService.database.insertCode()` directly. (would also need to move the withTransaction call)

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
```diff
@@ -0,0 +1,164 @@
...
+  if (newContractInsertResult.rows.length === 0) {
+    // Contract already exists, get the id
+    const existingContractResult = await poolClient.query(
+      `SELECT id FROM contracts WHERE creation_code_hash = $1 AND runtime_code_hash = $2`,
+      [newCreationCodeHash, runtimeCodeHash],
+    );
+    newContractId = existingContractResult.rows[0].id;
+  } else {
+    newContractId = newContractInsertResult.rows[0].id;
+  }
```
manuelwedler: Same here, we could reuse `insertContract()`.

services/server/src/server/apiv1/verification/private/stateless/private.stateless.handlers.ts
kuzdogan: I don't have the full context. If this fixes #1990 it's fine. I just in general think it's good to have a way to delete a verified contract as I explained here https://github.com/ethereum/sourcify/issues/1990#issuecomment-3131161875 but this is not urgent or necessary. My main use is being able to delete contracts locally in the dev environment. In stagin/prod maybe deleting a contract and going through the full verification flow to reproduce somehting

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
```diff
@@ -0,0 +1,164 @@
...
+  // Get database columns from verification
+  const databaseColumns =
+    await getDatabaseColumnsFromVerification(verification);
+
+  // Get existing verified contract to find deployment_id
+  const existingVerifiedContractQuery = `
+    SELECT vc.id, vc.deployment_id, cd.chain_id, cd.address, cd.contract_id
+    FROM verified_contracts vc
+    JOIN contract_deployments cd ON cd.id = vc.deployment_id
+    LEFT JOIN sourcify_matches sm ON sm.verified_contract_id = vc.id
```
manuelwedler: Since you use a left join here, am I right that you could get multiple verified_contracts for the same deployment here? Shouldn't we just care about the one referenced by a sourcify_match?

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
```diff
@@ -0,0 +1,164 @@
...
+  verification: VerificationExport,
+) => {
+  if (
+    verification.status.creationMatch !== "perfect" &&
+    verification.status.creationMatch !== "partial"
+  ) {
+    throw new Error(
+      "Creation match is null, cannot replace creation information",
+    );
+  }
```
manuelwedler: With my other comment, I meant to check if the `VerificationExport` contains creation bytecode. I think for the contracts affected by #1990, it is still possible that there is no creation match.

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
```diff
@@ -0,0 +1,164 @@
...
+       creation_metadata_match = $5
+     WHERE id = $1`,
+    [
+      existingVerifiedContract.id,
+      databaseColumns.verifiedContract.creation_match,
+      databaseColumns.verifiedContract.creation_values,
+      databaseColumns.verifiedContract.creation_transformations
+        ? JSON.stringify(
+            databaseColumns.verifiedContract.creation_transformations,
+          )
```
manuelwedler: vscode shows me an eslint error here. I just found that our CI doesn't run the lint check because it is configured wrongly... Will open a pr to fix

manuelwedler (changes requested):

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
marcocastignoli: My vscode doesn't

manuelwedler: Do you have the eslint extension installed?

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
marcocastignoli: Good catch, I actually wanted to use an INNER JOIN

services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
marcocastignoli: lol I forgot to install it after I moved back to vscode :)

PR #2268, opened 2025-07-31
Add massive replace script with configuration and API integration

marcocastignoli:
I just thought of creating a generic massive-replace script that calls the new `/private/replace-contract` endpoint efficiently. The script is configurable by passing a config file so we can re-use it to run other fixes on sourcify.

```js
// Configuration for replace-creation-information use case
// This configuration targets contracts where creation_code_hash equals runtime_code_hash
// and updates their creation information

module.exports = {
  query: async (sourcePool, sourcifySchema, currentVerifiedContract, n) => {
    return await sourcePool.query(
      `
      SELECT 
          [...]
          AND vc.id >= $1
          [...]
      LIMIT $2
    `,
      [currentVerifiedContract, n],
    );
  },
  buildRequestBody: (contract) => {
    return {
      chainId: contract.chain_id.toString(),
      address: `0x${contract.address.toString("hex")}`,
      forceCompilation: false,
      forceRPCRequest: true,
      customReplaceMethod: "replace-creation-information",
    };
  },
  description:
    "Updates existing verified contracts with creation bytecode information for contracts where creation and runtime bytecode are identical",
};

```

### DIFF
--- a/services/database/.env.template
+++ b/services/database/.env.template
@@ -20,3 +20,17 @@ DATABASE_URL=postgres://username:password@127.0.0.1:5432/database_name?sslmode=d
 
 # Keep this for updating the schema dump after adding a new migration
 DBMATE_SCHEMA_FILE=./sourcify-database.sql
+
+
+##### Massive replace script configuration
+# Database configuration
+SOURCIFY_HOST=localhost
+SOURCIFY_DB=sourcify
+SOURCIFY_USER=sourcify
+SOURCIFY_PASSWORD=sourcify
+SOURCIFY_PORT=5432
+SOURCIFY_SCHEMA=public
+
+# API configuration
+API_BASE_URL=https://staging.sourcify.dev/server
+API_AUTH_TOKEN=token_here
--- a/services/database/massive-replace-script/config-replace-creation-information.js
+++ b/services/database/massive-replace-script/config-replace-creation-information.js
@@ -1,0 +1,47 @@
+// Configuration for replace-creation-information use case
+// This configuration targets contracts where creation_code_hash equals runtime_code_hash
+// and updates their creation information
+
+module.exports = {
+  query: async (sourcePool, sourcifySchema, currentVerifiedContract, n) => {
+    return await sourcePool.query(
+      `
+      SELECT 
+          cd.chain_id,
+          cd.address,
+          cd.transaction_hash,
+          c.creation_code_hash,
+          c.runtime_code_hash,
+          vc.runtime_match,
+          vc.creation_match,
+          sm.runtime_match as sourcify_runtime_match,
+          sm.creation_match as sourcify_creation_match,
+          cd.created_at,
+          vc.id as verified_contract_id
+      FROM ${sourcifySchema}.contract_deployments cd
+      JOIN ${sourcifySchema}.contracts c ON cd.contract_id = c.id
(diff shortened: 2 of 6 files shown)


services/database/.env.template
```diff
@@ -20,3 +20,17 @@ DATABASE_URL=postgres://username:password@127.0.0.1:5432/database_name?sslmode=d
 
 # Keep this for updating the schema dump after adding a new migration
 DBMATE_SCHEMA_FILE=./sourcify-database.sql
+
+
+##### Massive replace script configuration
```
manuelwedler, line 25: Should we reuse the `POSTGRES_HOST` and other env variables from above? I feel like we shouldn't duplicate them for every functionality. For dbmate I only needed to create a new variable because it only supports connecting via the url.

services/database/massive-replace-script/config-replace-creation-information.js
```diff
@@ -0,0 +1,46 @@
...
+          vc.runtime_match,
+          vc.creation_match,
+          sm.runtime_match as sourcify_runtime_match,
+          sm.creation_match as sourcify_creation_match,
+          cd.created_at,
+          vc.id as verified_contract_id
+      FROM ${sourcifySchema}.contract_deployments cd
+      JOIN ${sourcifySchema}.contracts c ON cd.contract_id = c.id
+      JOIN ${sourcifySchema}.verified_contracts vc ON vc.deployment_id = cd.id
+      LEFT JOIN ${sourcifySchema}.sourcify_matches sm ON sm.verified_contract_id = vc.id
```
manuelwedler: why a left join here?

services/database/massive-replace-script/config-replace-creation-information.js
```diff
@@ -0,0 +1,46 @@
...
+          cd.address,
+          cd.transaction_hash,
+          c.creation_code_hash,
+          c.runtime_code_hash,
+          vc.runtime_match,
+          vc.creation_match,
+          sm.runtime_match as sourcify_runtime_match,
+          sm.creation_match as sourcify_creation_match,
+          cd.created_at,
+          vc.id as verified_contract_id
```
manuelwedler: I wonder if we should rather use the sourcify_match id for the CURRENT_VERIFIED_CONTRACT counter as it's the top level entrypoint to the db.

services/database/massive-replace-script/config-replace-creation-information.js
marcocastignoli: Uuups :D

services/database/massive-replace-script/config-replace-creation-information.js
```diff
@@ -0,0 +1,46 @@
...
+      FROM ${sourcifySchema}.contract_deployments cd
+      JOIN ${sourcifySchema}.contracts c ON cd.contract_id = c.id
+      JOIN ${sourcifySchema}.verified_contracts vc ON vc.deployment_id = cd.id
+      INNER JOIN ${sourcifySchema}.sourcify_matches sm ON sm.verified_contract_id = vc.id
+      WHERE c.creation_code_hash = c.runtime_code_hash
+          AND c.creation_code_hash IS NOT NULL
+          AND c.runtime_code_hash IS NOT NULL
+          AND sm.id >= $1
```
manuelwedler, line 28: you changed the `sm.id >= $1` clause to use the sourcify_matches id but the other parts still use `vc.id`

Issue #2208, opened 2025-07-02
"Duplicate key value violates unique constraint" error during verification


marcocastignoli:
Address: `0xdAC17F958D2ee523a2206206994597C13D831ec7`
ChainId: 1

I used staging.sourcify.dev to "Import from Etherscan" the contract and I received the following error:

![Image](https://github.com/user-attachments/assets/5bca2d67-6d7e-4468-a3b7-114e3886ffe4)

```
cannot update verified_contract address=����.�#� b�E��=�� chainId=1
error: duplicate key value violates unique constraint "verified_contracts_pseudo_pkey"
```

~~Bonus error: the address is not formatted to string~~ (not relevant anymore with the new UI)

kuzdogan:
This also looks somewhat important, moving to candidates

manuelwedler:
The good thing about this contract is you can have a comparison between production and staging. On production we have a creation match, on staging not:

https://repo.sourcify.dev/1/0xdAC17F958D2ee523a2206206994597C13D831ec7
https://repo.staging.sourcify.dev/1/0xdAC17F958D2ee523a2206206994597C13D831ec7

The issue seems to be that an old version of Sourcify didn't apply the constuctor arguments transformation correctly and therefore didn't get a creation match on staging. The current version is able to match creation code correctly but the onchain bytecode didn't change, also the unique keys of the compiled_contract did not changed. Therefore, we get the key violation error.

I think this is just another occurrence of https://github.com/ethereum/sourcify/issues/2086 that we didn't catch yet.

manuelwedler:
I think we can fix it with the `/private/replace-contract` endpoint. The problem is that I am struggeling to find an efficient query to get all of these contracts. We basically need to find contracts where the onchain creation code matches the recompiled creation code excluding constructor arguments, auxdata and libraries. Since libraries can be anywhere in the bytecode, getting all instances of this problem by comparing their bytecodes in a query seems infeasible to me.

 Maybe we can just run the endpoint on all contracts with `creationMatch = false` and creation information not null? Since we would not use rpc data nor recompile, it should be quick, right? The only issue I see is that I wouldn't want to replace contracts (meaning inserting a new sourcify_match) if the reverification doesn't lead to a better match, but for that we can modify the endpoint.

kuzdogan:
[...]

PR #2282, opened 2025-08-07
Replace script config for fixing missing constructor args transformations

manuelwedler:
See #2208

This adds a new configuration for the massive replace script to fix the contracts of the linked issue. We found one problem in the `replace-contract` endpoint that led to writing an empty transaction hash when replacing. This is also addressed by correctly assigning the transaction hash which is retrieved from the database. Additionally, this adds smaller adjustments to the massive replace script like an option to store info about contracts which failed replacing.


marcocastignoli (approved):
LGTM 👍 I like the new method to store IDs of failed contracts

PR #2283, opened 2025-08-07
Rpcs housekeeping

kuzdogan:
Removes unstable RPCs

### DIFF
--- a/services/server/src/server/services/utils/contract-creation-util.ts
+++ b/services/server/src/server/services/utils/contract-creation-util.ts
@@ -243,6 +243,7 @@ async function getCreatorTxUsingFetcher(
     }
   } catch (e: any) {
     logger.warn("Error while getting creation transaction", {
+      url: fetcher?.url,
       error: e.message,
     });
     return null;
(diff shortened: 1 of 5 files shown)


services/server/src/server/services/utils/contract-creation-util.ts
```diff
@@ -243,6 +243,7 @@ async function getCreatorTxUsingFetcher(
     }
   } catch (e: any) {
     logger.warn("Error while getting creation transaction", {
+      url: fetcher?.url,
```
manuelwedler: This will log API keys (currently Etherscan and VeChain). This is especially bad since CI logs are public. Let's either prune the keys before logging, or not log the URL.

manuelwedler (changes requested):

services/server/src/server/services/utils/contract-creation-util.ts
manuelwedler: Ah after looking at the code I saw that we actually skip these tests for that reason. I still think it's not great that we log the keys in GCP, but not blocking for this PR.

kuzdogan:
Ok I've created new entries in `SourcifyChain.maskedUrl` and creation tx's `fetcher.maskedUrl` to be used in log statements. Final review pls

manuelwedler (approved):
Nice, thanks for the fix :)

I guess we can close this one then?
https://github.com/sourcifyeth/private-issues/issues/28

kuzdogan:
Nice yes thanks for remembering

Issue #2227, opened 2025-07-12
Verified address has incorrect metadata or source


sealer3:
See `0x000000000022D473030F116dDEE9F6B43aC78BA3` on chain 1.

The metadata.json has at least one source (`src/AllowanceTransfer.sol`) whose hash in the metadata differs from the source returned by the `sources` field response.

```diff
diff --git a/./AllowanceTransfer.sol b/tmp/set2.sol
index ddf3781..84edcde 100644
--- a/./AllowanceTransfer.sol
+++ b/tmp/set2.sol
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.17;
 
-import {ERC20} from "solmate/src/tokens/ERC20.sol";
-import {SafeTransferLib} from "solmate/src/utils/SafeTransferLib.sol";
+import {ERC20} from "solmate/tokens/ERC20.sol";
+import {SafeTransferLib} from "solmate/utils/SafeTransferLib.sol";
 import {PermitHash} from "./libraries/PermitHash.sol";
 import {SignatureVerification} from "./libraries/SignatureVerification.sol";
 import {EIP712} from "./EIP712.sol";
```

(the "plus" is the file whose hash is in the metadata object, which I retrieved via IPFS, and the "minus" is what Sourcify returns in the sources field)

manuelwedler:
My guess is that this problem originates from this: https://github.com/ethereum/sourcify/pull/1475
We originally stored the metadata which was provided by the user. We fixed the code, but not the data in the database for this issue.

I am not sure when we deployed the fix above. The release that included it was [sourcify-server@2.1.0](https://github.com/ethereum/sourcify/releases/tag/sourcify-server%402.1.0) published on Jul 25, 2024. But the contract above was verified on 2024-08-08. I think it might be that we were still running an old version before we switched our server deployment.

From the GCP deployment I can see we deployed on 2024-08-12 but I cannot find out which version this was because the image is not tagged in ghcr. The version deployed on 2024-08-29 should definitely have the fix because it relates to the [sourcify-server@2.2.0](https://github.com/ethereum/sourcify/releases/tag/sourcify-server%402.2.0) release.
So in my opinion it would be best to check the metadata for all contracts verified before that date which are not perfect matches:

```sql
select count(*) from sourcify_matches sm where sm.created_at < '2024-08-29 08:58:57 +0200' and runtime_match ='partial';
```

-> 3,605,591 contracts

[...]

PR #2285, opened 2025-08-07
Replace script config for fixing mismatched metadata

manuelwedler:
See #2227

- adds the configuration to the massive replace script for fixing the above issue
- adds an option to exclude contracts from submission to the `replace-contract` endpoint in order to only update contracts with wrong metadata
- adds a new custom replace method to only update metadata on the submitted contract


marcocastignoli (approved):
lgtm

PR #2288, opened 2025-08-11
Release new package versions

manuelwedler:


services/database/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog for `sourcify-database`
 
+## sourcify-database@2.7.0 - 2025-08-11
```
marcocastignoli, line 3: if there are no new features we can just bump for patch, it's still ok. Just saying for the next time

marcocastignoli (approved):

services/database/CHANGELOG.md
manuelwedler: I was a bit unsure if this is a patch or a minor update. It actually contains minor new features for the massive replace script, so I went with minor.

PR #2291, opened 2025-08-11
Handle vyper compiler warnings in asyncExec function

marcocastignoli:
For some reason Vyper outputs warnings in `stderr`. With this PR we are handling this case.


kuzdogan:
Maybe it's better we handle it all in `vyperCompiler.ts` as we do in the SolidityCompiler?

https://github.com/ethereum/sourcify/blob/0a417e6de521ee23a8e3d19b9192780d36c2ba7e/packages/compilers/src/lib/vyperCompiler.ts#L80-L89

I'm asking because I'm wondering if we throw when we see something in the stdErr, are we going to see these as structured and meaningful errors in a verificationJob? Or just an unstructured string that is coming from the stdErr?

From what I can see it won't because you are checking `startsWith` which suggests this is a simple string. 

Also what happens if there's a Warning in the beginning and errors come later?

marcocastignoli:
> just an unstructured string that is coming from the stdErr

The one I see was just an unstructured string, so I don't think an advanced parsing is needed.

kuzdogan:
Ok but to my point, when we throw it on the stdErr output, we don't get the structured errors like this one https://verify.sourcify.dev/jobs/65f64b95-2e1b-4a8f-9db3-66985b1c15bd

and also this: "Also what happens if there's a Warning in the beginning and errors come later?"

marcocastignoli:
> Also what happens if there's a Warning in the beginning and errors come later?

Ah! I lost this one, it actually makes sense. Then unfortunately we need to dig deeper into this and create some tests to explore all the cases :(

kuzdogan:
How about ignoring the stderr and looking for compiler errors in the std JSON output? I wonder if this line of checking stdErr does ever get triggered in Solidity because this `asyncExec` is a shared function. We need to check how the changes affect both solidity and vyper compilers

marcocastignoli:
I'll move forward with this by creating some tests with different errors / warnings for both solidity and vyper. Then we'll adapt the fix so that it doesn't break any test.

marcocastignoli:
I discovered that Vyper < 0.4.0 outputs warnings in the stderr, while vyper >=0.4.0 includes them correctly in the errors object.

@kuzdogan I think the approach of this PR is correct, I added the tests for Vyper to make sure we handle cases (< 0.4.0 AND >= 0.4.0), what do you think?

packages/compilers/test/vyperCompiler.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    );
+    expect(compiledJSON?.contracts?.['test.vy']?.test).to.not.equal(undefined);
+    // 0.3.10 doesn't include the warnings in the standardJson output
+    expect(compiledJSON?.errors).to.be.undefined;
+  });
+
+  it('Should handle vyper 0.4.1 warnings', async function () {
+    const compiledJSON = await useVyperCompiler(
+      vyperRepoPath,
+      '0.4.0+commit.e9db8d9f',
```
manuelwedler: the test description uses a different vyper version than the test itself
```suggestion
  it('Should handle vyper 0.4.0 warnings', async function () {
    const compiledJSON = await useVyperCompiler(
      vyperRepoPath,
      '0.4.0+commit.e9db8d9f',
```

packages/compilers/src/lib/common.ts
```diff
@@ -74,9 +74,15 @@ export function asyncExec(
         if (error) {
           reject(error);
         } else if (stderr) {
-          reject(
-            new Error(`Compiler process returned with errors:\n ${stderr}`),
-          );
+          // Vyper compilers <0.4.0 outputs warnings to stderr
+          // we handle this by checking if the stderr starts with "Warning:"
+          if (stderr.startsWith('Warning:')) {
```
manuelwedler, line 79: Isn't stderr the complete output over multiple lines? This just checks the first line, but is there a possibility that the first line is a warning and the second an error?

marcocastignoli: Good idea, let me check

marcocastignoli: > is there a possibility that the first line is a warning and the second an error

This answer to your question, the warning is in the stderr, while the errors are in the error property of the object. I'm adding a test for this.

#### command

```bash
> ./vyper.0.3.10 --standard-json << 'EOF'
{
  "language": "Vyper",
  "sources": {
    "test.vy": {
      "content": "@exteral\n@view\ndef foo(x: uint256, y: int128) -> uint256:\n return shift(x, y)"
    }
  },
  "settings": {
    "outputSelection": {
      "*": ["*"]
    }
  }
}
EOF
```

#### result
```
Warning: `shift()` is deprecated! Please use the << or >> operator instead.
{"compiler": "vyper-0.3.10", "errors": [{"component": "compiler", "formattedMessage": "Unknown decorator: exteral\n  contract \"test.vy:1\", function \"foo\", line 1:1 \n  ---> 1 @exteral\n  --------^\n       2 @view\n", "message": "Unknown decorator: exteral", "severity": "error", "sourceLocation": {"file": "test.vy"}, "type": "FunctionDeclarationException"}]}
```

The formatted error is 
```json
{
  "compiler": "vyper-0.3.10",
  "errors": [
    {
      "component": "compiler",
      "formattedMessage": "Unknown decorator: exteral\n  contract \"test.vy:1\", function \"foo\", line 1:1 \n  ---> 1 @exteral\n  --------^\n       2 @view\n",
      "message": "Unknown decorator: exteral",
      "severity": "error",
      "sourceLocation": { "file": "test.vy" },
      "type": "FunctionDeclarationException"
    }
  ]
}
```

manuelwedler: ah okay that makes sense :+1:

manuelwedler (approved):

PR #2297, opened 2025-08-12
Move Etherscan utils to lib-sourcify

marcocastignoli:


packages/lib-sourcify/README.md
```diff
@@ -488,14 +499,45 @@ setLibSourcifyLogger({
...
+### Fetching Contract Data from Etherscan
+
+```typescript
+import { EtherscanUtils } from '@ethereum-sourcify/lib-sourcify';
+
+// Fetch contract source code and metadata from Etherscan
+const etherscanResult = await EtherscanUtils.fetchFromEtherscan(
+  1, // chainId
+  '0x6B175474E89094C44Da98b954EedeAC495271d0F', // contract address
+  'YOUR_API_KEY', // optional API key
```
kuzdogan: This should be required, no?

packages/lib-sourcify/src/utils/etherscan/etherscan-util.ts
```diff
@@ -0,0 +1,416 @@
...
+export const fetchFromEtherscan = async (
+  chainId: number | string,
+  address: string,
+  apiKey: string = '',
+): Promise<EtherscanResult> => {
+  const url = `https://api.etherscan.io/v2/api?chainid=${chainId}&module=contract&action=getsourcecode&address=${address}&apikey=`;
+  const secretUrl = url + apiKey;
+  const maskedUrl = url + (apiKey ? apiKey.slice(0, 6) + '...' : '');
+
+  let response: any;
```
kuzdogan: There's the `Response` type we can use instead of any

```suggestion
  let response: Response;
```

services/server/src/server/services/utils/etherscan-util.ts
```diff
@@ -1,488 +1,110 @@
...
-  const secretUrl = url + apiKey;
-
-  url = url + apiKey.slice(0, 6) + "...";
-  let response;
-  logger.info("Fetching from Etherscan", {
-    url,
-    chainId: sourcifyChain.chainId,
-    address,
-  });
+  throwV2Errors = false,
```
kuzdogan, line 67: What was the use of this parameter? I don't see it being used anywhere

packages/lib-sourcify/package.json
```diff
@@ -45,6 +45,7 @@
     "@ethersproject/bignumber": "5.8.0",
     "@ethersproject/bytes": "5.8.0",
     "@fairdatasociety/bmt-js": "2.1.0",
+    "@solidity-parser/parser": "0.20.2",
```
kuzdogan, line 47: Let's remove this from the server

packages/lib-sourcify/test/utils/etherscan-util.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+  MULTIPLE_CONTRACT_RESPONSE,
+  RATE_LIMIT_REACHED_RESPONSE,
+  SINGLE_CONTRACT_RESPONSE,
+  STANDARD_JSON_CONTRACT_RESPONSE,
+  UNVERIFIED_CONTRACT_RESPONSE,
+  VYPER_SINGLE_CONTRACT_RESPONSE,
+  VYPER_STANDARD_JSON_CONTRACT_RESPONSE,
+  mockEtherscanApi,
+} from '../../../../services/server/test/helpers/etherscanResponseMocks';
+import { EtherscanImportError } from '../../src/utils/etherscan/EtherscanTypes';
```
kuzdogan, line 27: Why are these coming from the server? I think `etherscanResponseMocks.ts` should also have been in lib-sourcify

kuzdogan (changes requested):
Overall lgtm. Small things 

I just renamed the `url` in the etherscan-utils to `maskedUrl` following the commit 99048366667789b14965198d990251d9d614548a that happened while you were working on this

services/server/src/server/services/utils/etherscan-util.ts
marcocastignoli: For example in `services/server/src/server/apiv2/verification/verification.handlers.ts:132` you can see it being used. It's needed because we need to send different error responses depending on v1 vs v2

packages/lib-sourcify/test/utils/etherscan-util.spec.ts
marcocastignoli: I also put a comment, you don't agree with that reasoning?

> Here we import the mock responses directly from server to avoid duplication and ensure consistency across lib-sourcify and server tests.

I think it makes no sense to duplicate this, since we have a monorepo we can take advantage of this, at the end it's just for tests

packages/lib-sourcify/src/utils/etherscan/etherscan-util.ts
kuzdogan: You're still assigning `as any` below

marcocastignoli: sorry I didn't see it

packages/lib-sourcify/test/utils/etherscan-util.spec.ts
kuzdogan: I'm not suggesting to duplicate it but it's better if it's coming from lib-sourcify as a rule of thumb instead of server to lib-sourcify. But shouldn't matter too much as you said

marcocastignoli: Ah, yeah either lib-sourcify or server is the same for me.

marcocastignoli:
@kuzdogan I had to rebase it again because of conflicts with staging. Now it should be ready.

kuzdogan (approved):

Issue #2247, opened 2025-07-22
Input validation preventing from verifying Vyper versions lower than 0.3.2


manuelwedler:
While pushing the data from my script, I noticed that I get a lot of `invalid_parameter` errors for Vyper contracts:

```json
{"customCode":"invalid_parameter","message":"request/body/compilerVersion must match pattern "\d+\.\d+\.\d+(-nightly\.\d{4}\.\d+\.\d+)?\+commit\.[a-f0-9]{8}"","errorId":"e0358a83-5c52-4cb1-9091-298e5c0d9440"}
```

The issue is that our input validation requires the commit hash part to be exactly 8 characters. As you can see in this [list](https://vyper-releases-mirror.hardhat.org/list.json), Vyper versions 0.3.0 and lower used 7 characters for the commit hash. Version 0.3.1 skips it even completely.

kuzdogan:
I can do a quick fix but I feel like we need to standardize this. So in our DB there are entries in format `0.4.0+commit.e9db8d9f` but the 0.3.1 is saved as `0.3.1`. I assume these were all created with apiv1 or apiv2 Etherscan?

<details>
<summary>Example Sourcify DB Vyper Contracts</summary>

|version|created_at|chain_id|address|compiler|
|-------|----------|--------|-------|--------|
|0.3.10+commit.91361694|2025-02-21 16:55:52.219 +0100|421614|b91d7c61c369390c365717689fa5ced07212118c|vyper|
|0.3.10+commit.91361694|2025-01-10 22:44:22.105 +0100|8453|99efc97ee98e72efa3331e7cbc114e0079c11e78|vyper|
|0.2.12+commit.2c6842c|2025-01-14 09:15:15.494 +0100|1|042815d6fe33152c123059c0e2be77404769b9c8|vyper|
|0.3.10+commit.91361694|2025-01-21 14:39:13.153 +0100|8453|cf0cb198c4bf8cc3db61a852c6b70a2503b772af|vyper|
|0.4.0+commit.e9db8d9f|2025-05-24 14:42:56.244 +0200|42161|b5fd541d9b77ce5cedc35bb3118dad05858431ed|vyper|
|0.4.0+commit.e9db8d9f|2025-05-24 15:11:26.250 +0200|42161|d7000cd9ec8cc95e88005ec75597f595806054fa|vyper|
[...]

PR #2303, opened 2025-08-13
Compiler version validation

kuzdogan:
See https://github.com/ethereum/sourcify/issues/2247#issuecomment-3155878775

### DIFF
--- a/services/server/src/apiv2.yaml
+++ b/services/server/src/apiv2.yaml
@@ -312,7 +312,7 @@ components:
 
                   Must include the `language` field inside the stdJsonInput object. Currently supports Solidity and Vyper. 
               compilerVersion:
-                $ref: '#/components/schemas/SolidityCompilerVersion'
+                $ref: '#/components/schemas/CompilerVersion'
               contractIdentifier:
                 type: string
                 description: The fully qualified file path and contract name to indicate which contract to verify.
@@ -581,7 +581,8 @@ components:
                         type: string
                         example: 'solc'
                       compilerVersion:
-                        $ref: '#/components/schemas/SolidityCompilerVersion'
+                        $ref: '#/components/schemas/CompilerVersion'
+                        description: 'The server accepts v prefixed but should only respond without the v prefix'
                       compilerSettings:
                         type: object
                       name:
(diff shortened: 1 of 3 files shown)


services/server/src/server/apiv2/middlewares.ts
```diff
@@ -105,6 +105,39 @@ export function validateFieldsAndOmit(
...
+  next: NextFunction,
+) {
+  let compilerVersion = req.body.compilerVersion;
+  if (!compilerVersion) {
+    throw new InvalidParametersError("Compiler version is required.");
+  }
+
+  if (compilerVersion.startsWith("v")) {
+    compilerVersion = compilerVersion.slice(1);
+  }
```
manuelwedler, line 120: Note that I added something similar to `AbstractCompilation` yesterday:

https://github.com/ethereum/sourcify/blob/d11e3969aff8508bd901220952539b00331d1de8/packages/lib-sourcify/src/Compilation/AbstractCompilation.ts#L24-L27

Probably we can keep both.

services/server/src/apiv2.yaml
```diff
@@ -1119,6 +1120,15 @@ components:
       title: SolidityCompilerVersion
       pattern: '\d+\.\d+\.\d+(-nightly\.\d{4}\.\d+\.\d+)?\+commit\.[a-f0-9]{8}'
       example: '0.8.7+commit.e28d00a7'
```
manuelwedler: Can we remove the `SolidityCompilerVersion` component? It's still used in the `VerifiedContract` response but I think it would also be more accurate to use the `CompilerVersion` there.

kuzdogan: Good catch, my bad here

kuzdogan: Good catch, it's actually not used anywhere

services/server/src/server/apiv2/middlewares.ts
kuzdogan: I'd say keep both, this is for server

manuelwedler (approved):
nice, good testing

PR #2308, opened 2025-08-18
Refactor constructor arguments transformation to use ABI instead of Metadata

marcocastignoli:


kuzdogan:
Can you explain the reason for the change? Why did `JsonFragment[]`  (metadata.output.abi's type) not work?

marcocastignoli:
> Can you explain the reason for the change?

We discussed about this in the last call, we want to make processes less dependent on Metadata. This is I think the only usage of Metadata in the verification class

> Why did JsonFragment[] (metadata.output.abi's type) not work?

It's actually using the same type.  This is the declaration of `InterfaceAbi`: `export type InterfaceAbi = string | ReadonlyArray<Fragment | JsonFragment | string>;`

Issue #2239, opened 2025-07-17
Use "ContractFileName" field in Etherscan API directly


kuzdogan:
Event though not documented, it seems Etherscan is retuning a new field "ContractFileName" to indicate the path of the contract. 

We should use this field but also keep falling back to our previous parsing approach if the field does not exist e.g. for older Etherscan API versions

kuzdogan:
It seems the API doesn't return the correct value always #2290 so closing this for now

marcocastignoli:
> It seems the API doesn't return the correct value always [#2290](https://github.com/ethereum/sourcify/issues/2290) so closing this for now

Basically the UI code is using `ContractFileName` in combination with `ContractName`. While `ContractFileName` is correct, the `ContractName` is not. 

In the PR #2312 I implemented the usage of `ContractFileName` only for Vyper contracts, in which the contract name is just the `ContractFileName` without the ".vy" extension.

[See the docs](https://docs.vyperlang.org/en/stable/compiling-a-contract.html) for more info 

```js
// This contains the contract-level outputs. Can be limited/filtered by the outputSelection settings.
    "contracts": {
        "source_file.vy": {
            // The contract name will always be the file name without a suffix
            "source_file": {
```

kuzdogan:
I don't get it. How is `ContractName` not correct? You mean the Etherscan is returning the wrong `ContractName`? 

Also why are you using `ContractFileName` only for Vyper contracts? Isn't it returned in Solidity?

manuelwedler:
> Also why are you using `ContractFileName` only for Vyper contracts? Isn't it returned in Solidity?

Yes, if the `ContractFileName` is working consistently, we should remove the Solidity parser and just rely on that field.

kuzdogan:
Ok so from what I can see:
- Etherscan returns `ContractFileName` for std-json and multi-file for both Solidity and Vyper.
- For single file Solidity and Vyper, Etherscan returns empty `ContractFileName`. In that case we need to create ourselves for the std-json.
- Etherscan can return `ContractName` in Vyper with whitespaces like `My Token Contract`.
- `ContractName` and `ContractFileName` do not always have to match for Vyper. e.g. here name is `USDaf Flash Zapper` and path is `src/flash_zapper.vy`. 
  - https://api.etherscan.io/v2/api?chainid=1&module=contract&action=getsourcecode&address=0xf97f41F003a5b51aBd62Aadec12445cE47c96C8C&apikey=<API_KEY>
[...]

PR #2312, opened 2025-08-19
Add support for `ContractFileName` for "Import from Etherscan"

marcocastignoli:
See #2239

This PR removes the solidity-parser from lib-sourcify replacing it with `ContractFileName`. The solidity-parser is still needed for the /v1/solc-json endpoint so I moved it back to server (with its tests)

Now instead of using the solidity-parser, the contract's target path is extracted from the `ContractFileName` for all cases except "single-file import from Etherscan", in which:
- for Vyper the contract path is constructed from the normalized contract name + ".sol" (because the `ContractName` property in Etherscan can return a contract name with whitespaces)
- for Solidity the contract path is constructed from the contract name + ".sol"


packages/lib-sourcify/src/utils/etherscan/etherscan-util.ts
```diff
@@ -138,20 +137,6 @@ export const getSolcJsonInputFromEtherscanResult = (
...
 };
 
-export const getContractPathFromSourcesOrThrow = (
-  contractName: string,
-  sources: Sources,
-): string => {
-  const contractPath = getContractPathFromSources(contractName, sources);
-  if (contractPath === undefined) {
-    throw new EtherscanImportError({
-      code: 'etherscan_missing_contract_definition',
```
manuelwedler, line 148: This error code can also be removed completely I think.

manuelwedler: I mean from the EtherscanTypes

packages/lib-sourcify/src/utils/etherscan/etherscan-util.ts
```diff
@@ -333,18 +330,8 @@ export const processVyperResultFromEtherscan = async (
...
-      if (!contractPath) {
-        throw new EtherscanImportError({
-          code: 'etherscan_missing_contract_in_json',
-          contractName: contractResultJson.ContractName,
-        });
-      }
-    }
+    contractPath =
+      getContractFileNameFromEtherscanResultOrThrow(contractResultJson);
     contractName = contractPath.split('/').pop()!.split('.')[0];
```
kuzdogan: I think this needs to split from the last `.` but take the part before the last `.`. Ie. if the contract name has `.` 

path: `path/to/my.weird.contract.vy` contract name will be `my`, it should be `my.weird.contract`

manuelwedler (approved):
looks great

PR #2313, opened 2025-08-19
Skip schema validation in CI

manuelwedler:
pg_dump had some changes in the recent postgres release and doesn't produce a consistent result anymore. For that reason we get a failing CI here: https://app.circleci.com/pipelines/github/ethereum/sourcify/9290/workflows/f03100ba-b7d8-4894-a2b6-b77021fd89b7/jobs/54501

This skips the CI schema validation:

For  reference, pg_dump adds these lines since 16.10:

```
\restrict wzpNUSKP7T1c6qgyqXCflEagKcM9dkwutqIx5cIHPVh7KHWPDu4pd6WdTgxRfqu

-- Dumped from database version 16.10 (Ubuntu 16.10-1.pgdg24.04+1)
-- Dumped by pg_dump version 16.10 (Ubuntu 16.10-1.pgdg24.04+1)
```

for the restrict command there is actually an option to use a specific key (a new security feature), but dbmate would need to add support, or we could work around it by running the dump ourselves.
But we would still have the issue of the version comments which don't have an option to opt out.

### DIFF
--- a/.circleci/new_branch.yml
+++ b/.circleci/new_branch.yml
@@ -129,33 +129,35 @@ jobs:
           command: |
             cd services/database
             npm install
-      - run:
-          name: Validate schema is in sync with migrations
-          command: |
-            cd services/database
-            # Initialize git submodules
-            git submodule update --init
-            # Apply migrations using npm script
-            npm run migrate:up
-            # Force dump in case problems were ignored
-            npx dbmate dump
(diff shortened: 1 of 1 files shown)


kuzdogan:
Can you create a backlog issue for this? We'll probably remove this eventually

manuelwedler:
Sure, here it is: #2314

PR #2330, opened 2025-08-25
Initialize git submodules before running down migration test in CI

manuelwedler:
Missed that when I commented out the schema validation test (#2313)


marcocastignoli (approved):
is this what was failing in the CI?

manuelwedler:
yes exactly

PR #2334, opened 2025-08-27
Replace github repository occurences of "ethereum" with "argotorg"

marcocastignoli:


packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -282,7 +282,7 @@ export class Verification {
   handleSolidityExtraFileInputBug(): SolidityBugType {
     // Case when extra unused files in compiler input cause different bytecode
     // See issues:
-    //   https://github.com/ethereum/sourcify/issues/618
+    //   https://github.com/argotorg/sourcify/issues/618
     //   https://github.com/ethereum/solidity/issues/14250
     //   https://github.com/ethereum/solidity/issues/14494
```
manuelwedler: seeing this randomly. We should also update the links to solidity, no?
```suggestion
    //   https://github.com/argotorg/solidity/issues/14250
    //   https://github.com/argotorg/solidity/issues/14494
```

marcocastignoli: good catch! :)

PR #2335, opened 2025-08-27
Fix CircleCI missing base_revision variable

marcocastignoli:


manuelwedler:
For documentation, our new CircleCi org is integrated via GitHub as a GitHub App, not via GitHub OAuth. For this reason `pipeline.git.base_revision` is not available (see https://circleci.com/docs/guides/orchestrate/pipeline-variables/). 

Since we use `HEAD~1` as base revision now for deciding which images need to be rebuild after merging to staging, we cannot use GitHub merge by rebase anymore. I disabled the rebase via a new branch rule. @kuzdogan

PR #2339, opened 2025-08-27
Test to check CircleCI run on external PRs

marcocastignoli:


manuelwedler:
These CI checks are just displayed here because I enabled them to be required when I changed the branch rules. They actually don't run...

marcocastignoli:
Yeah I also noticed it

PR #2340, opened 2025-08-27
Restore `./circleci` configurations for our new CircleCI OAUTH organization

marcocastignoli:
This morning we created a new CircleCI organization using a different auth method than OAUTH. This caused several problems:
- there is no `pipeline.git.base_version` support
- the organization slug is `circleci/2bWbncbTSPHJDSWS2B9PoE` instead of `argotorg`, causing the codecov CircleCI orb to fail
- **Most importantly**: we cannot enable CircleCI for external PRs

This PR rolls back some changing we had to do in the `./circleci` configurations so we can support the new CircleCI OAUTH account we'll create soon.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ workflows:
             services/server/.* run-build-server true
             packages/.* run-build-packages true
           # Compare against the last build of the branch not the default "main" branch
-          base-revision: HEAD~1
+          base-revision: << pipeline.git.base_revision >>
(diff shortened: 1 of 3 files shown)


marcocastignoli:
@manuelwedler please double check that I'm not missing anything :)

manuelwedler:
I think it depends how our integration will look like in the end. If we just enable OAuth and have both integration types, I'm not sure whether we still need the slug

PR #2344, opened 2025-08-28
Add migrations for signature tables

### LINKED ISSUE #2316: 4bytes: Migrations for the 4bytes tables

manuelwedler:
Closes #2316

### DIFF
--- a/services/database/migrations/20250828092603_add_signature_tables.sql
+++ b/services/database/migrations/20250828092603_add_signature_tables.sql
@@ -1,0 +1,57 @@
+-- migrate:up
+
+/*
+    The `signatures` table stores signature information for compiled_contracts.
+    It includes each signature in text and its keccak hash.
+*/
+CREATE TABLE signatures (
+  /* 32 bytes, full keccak of the signature */
+  signature_hash_32 BYTEA PRIMARY KEY,
+
+  /* 4 bytes */
+  signature_hash_4 BYTEA GENERATED ALWAYS AS (SUBSTRING(signature_hash_32 FROM 1 FOR 4)) STORED,
+
+  /* the signature text, e.g. 'transfer(address,uint256)' */
+  signature VARCHAR NOT NULL,
+
+  /* type of signature: function, event, error */
+  signature_type VARCHAR NOT NULL CHECK (signature_type IN ('function','event','error')),
+
+  /* timestamps */
+  created_at timestamptz NOT NULL DEFAULT NOW(),
+  updated_at timestamptz NOT NULL DEFAULT NOW(),
+
+  CONSTRAINT signatures_pseudo_pkey UNIQUE (signature, signature_type)
+);
+
+CREATE INDEX signatures_hash_4_type_idx ON signatures (signature_hash_4, signature_type);
+
+CREATE TRIGGER update_set_updated_at
+  BEFORE UPDATE ON signatures
+  FOR EACH ROW
+  EXECUTE FUNCTION trigger_set_updated_at();
+
+/*
+    The `compiled_contracts_signatures` table links a compiled_contract to its associated signatures.
+    It forms a many-to-many relationship.
+*/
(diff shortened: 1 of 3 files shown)


services/server/package.json
```diff
@@ -15,7 +15,7 @@
     "start": "node ./dist/server/cli.js",
     "postgres-test:start": "docker-compose -f test/docker-compose.yml up -d",
     "postgres-test:stop": "docker-compose -f test/docker-compose.yml rm -fsv",
-    "postgres-test:migrate": "cd ../database && git submodule update --init && export DATABASE_URL=postgres://sourcify:sourcify@127.0.0.1:${DOCKER_HOST_POSTGRES_TEST_PORT}/sourcify?sslmode=disable && export DBMATE_SCHEMA_FILE=./sourcify-database.sql && npm run migrate:load-schema",
+    "postgres-test:migrate": "cd ../database && git submodule update --init && export DATABASE_URL=postgres://sourcify:sourcify@127.0.0.1:${DOCKER_HOST_POSTGRES_TEST_PORT}/sourcify?sslmode=disable && export DBMATE_SCHEMA_FILE=./sourcify-database.sql && npm run migrate:up",
```
marcocastignoli, line 18: Why this change?

marcocastignoli (approved):
just a question, but approved

services/server/package.json
manuelwedler: the CI failed on the first commit because it wasn't using the newest postgres 16 version in CI but I updated the schema using postgres 16.10. The schema has the `\restrict` commands in the schema but postgres 16.9 running in CI does not know about them yet.

manuelwedler: loading the schema via running the migrations prevents running into this.

services/database/migrations/20250828092603_add_signature_tables.sql
```diff
@@ -0,0 +1,57 @@
...
+/*
+    The `compiled_contracts_signatures` table links a compiled_contract to its associated signatures.
+    It forms a many-to-many relationship.
+*/
+CREATE TABLE compiled_contracts_signatures (
+  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+
+  /* the specific compilation and the specific signature */
+  compilation_id UUID NOT NULL REFERENCES compiled_contracts(id),
+  signature_hash_32 BYTEA NOT NULL REFERENCES signatures(signature_hash_32),
```
kuzdogan, line 43: Something I thought now when looking at it shortly.

What happens if a contract has identical `signature`s for an event AND e.g. a function. Like 

`function transfer(uint256,address)`
`event transfer(uint256,address)`

Maybe we should check if this is handled on the compiler level for both Vyper and Solidity. If not maybe the identifier should be `signature_hash_32` + `type`? With this definition we can have the above example. 

This is mainly due to `signature_hash_32` being the primary key

manuelwedler: I tried it and solc prevents having the same identifier twice even with different number of arguments. You can add this to a contract:
```solidity
event transfer(address);
function transfer(address to, uint value) external {
    emit transfer(to);
}
```
And you will get a `DeclarationError: Identifier already declared.`.

Same for Vyper:

```vyper
event Transfer:
    sender: address

@external
def Transfer(receiver: address):
    log Transfer(msg.sender, receiver)
 ```
 -> `NamespaceCollision:multiple functions or events named 'Transfer'!`

Btw, I also tried what happens if you write two functions with the same function selector (i.e. the 4 bytes) into a contract. If you add these:
```solidity
    function OwnerTransferV7b711143(uint256 test) public view returns (uint256)
    function withdraw(uint256 test) public view returns (uint256)
```

solc returns this error: `TypeError: Function signature hash collision for withdraw(uint256)`

kuzdogan: Great, thanks for checking

PR #2355, opened 2025-09-03
Move `signature_type` to `compiled_contracts_signatures` table in signature table migration

manuelwedler:
Related to #2316 

We decided to move `signature_type` to the `compiled_contracts_signatures` table, because it is a compilation attribute and it would otherwise not be possible to add the same signature for different types to the database.

I removed `compiled_contracts_signatures_compilation_idx` because the pseudo pkey constraint already creates an index with it as the leftmost column.


kuzdogan (approved):
LGTM

Issue #2240, opened 2025-07-21
4-bytes DB API Design


kuzdogan:
Let's have a call for the planned 4-bytes DB. We should have a look at the existing openchain.xyz spec, gather use cases here and discuss how the API should look here, and have a call in this sprint.

kuzdogan:
I had a look at what we have and summarizing here everything.

Requirements:
- Search by 4byte signature `0x70a08231`
  - Should they search like `?event=0xabd..cd` or just for all `0xab..cd`
- Search signature by human readable name (e.g. `transfer(address,uint256)`
- Addtionally, search by signature and get all contracts implementing that function/event/error. 
- (Optional) Searcy by chainId+address
  - For searching by chainid-address include the variable names too like `transfer(address _to, uint256 _amount)`. Do decoding tools allow variable names for that?
  - Probably not needed because you can get the abi
- For the first two above, the results need to be weighted by occurance as there is a lot of spam
- Include Event signatures
- Include Error signatures
- We should have a simple playground UI for this like https://www.etherface.io/text
  - Make it directly linkable with query params like `/text?=transfer*`
- Dump the database publicly
- At the end check if we cover all existing OpenChain signatures

IMO we can directly adopt the OpenChain API and extend it in a backwards compatible way. That way we can make the current url directly point to our servers and take over the service.

OpenChain docs: https://docs.openchain.xyz/

- Search by 4byte signature `0x70a08231`
  - Openchain: [/signature-database/v1/lookup](https://docs.openchain.xyz/#/default/get_signature_database_v1_lookup) e.g. `?function=0x227066c0,0x6891eb6f&event=0x009679b540aed4db019a586536beb8dc741acde4470efe562cd7dc340520c549`
  - We can add a `weight` field to indicate how likely one signature is the one compared to other results
- Search signature by human readable name (e.g. `transfer(address,uint256)`
  - Openchain: [/signature-database/v1/search](https://docs.openchain.xyz/#/default/get_signature_database_v1_search): e.g. `?query=transferFrom*` 
  - Openchain's search is pretty basic. We could add some advanced search e.g. in [4byte.directory API](https://www.4byte.directory/docs/): 
    - `/?text_signature=transfer(address,uint256)`=exact str match
    - `/?text_signature__icontains=Transf`= case insensitive search

Other:
- Maybe include a new APIv2 field for the signatures?

- How is the filtering done?

manuelwedler:
Draft for the database from our call:

[...]

PR #2356, opened 2025-09-04
Remove constructor type from signature tables migration

manuelwedler:
Related to #2316 
We decided to remove the constructor type again because it's not a callable signature


kuzdogan:
Related #2240

PR #2357, opened 2025-09-04
Add logic for writing signatures to db inside `SourcifyDatabaseService`

### LINKED ISSUE #2317: Write logic for the 4bytes in SourcifyDatabase

manuelwedler:
Closes #2317

- adds a new `signature-util.ts` to extract signatures from the ABI via ethers
- writes the signatures to the new table in a separate db transaction, such that verification job will still succeed even though there is an error while writing the signatures
- `storeSignatures` is also called when a match is updated, because it cannot write any duplicates due to the primary key, nor raise errors due to the `ON CONFLICT` clause


services/server/src/server/services/utils/signature-util.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const iface = new Interface(abi);
+
+  iface.fragments.forEach((fragment) => {
+    switch (fragment.type) {
+      case "function":
+      case "event":
+      case "error":
+        signatures.push(getSignatureData(fragment));
+    }
```
kuzdogan, line 26: We should wrapt this in a try/catch because fragment.format() will fail for some library contracts that have an invalid "type" field. Library ABI items are allowed to have custom types because their ABIs are not intended to be publicly callable. However this breaks with ethers when assembling the signature. See the latest messages in the Solidity private chat.

kuzdogan: Let's also add a test case for the above issue

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -892,10 +946,10 @@ export class SourcifyDatabaseService
       verificationId: VerificationJobId;
       finishTime: Date;
     },
-  ): Promise<void> {
+  ): Promise<{ verifiedContractId: Tables.VerifiedContract["id"] }> {
     try {
       const { type, verifiedContractId, oldVerifiedContractId } =
-        await super.insertOrUpdateVerification(verification, poolClient);
+        await this.insertOrUpdateVerification(verification, poolClient);
```
kuzdogan: Why the change? Does it have an effect on semantics?

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -884,6 +887,57 @@ export class SourcifyDatabaseService
...
+      const signatureData = extractSignaturesFromAbi(abi);
+      const signatureColumns = signatureData.map((sig) => ({
+        signature_hash_32: bytesFromString<BytesKeccak>(sig.signatureHash32),
+        signature: sig.signature,
+        signature_type: sig.signatureType,
+      }));
+
+      await this.database.insertSignatures(signatureColumns, poolClient);
+      await this.database.insertCompiledContractSignatures(
+        compilationId,
```
kuzdogan, line 922: Small thing but we are passing the same `signatureColumns` to both `insertSignatures` and `insertCompiledContractSignatures` functions. Even though the parameter `signatures` in both functions have different types. This works because the `signatureColumns` is a union of both types. 

I don't have a strong opinion here but just wanted to point out.

services/server/src/server/services/VerificationService.ts
```diff
@@ -345,14 +345,14 @@ export class VerificationService {
...
-  private async verifyViaWorker(
+  private verifyViaWorker(
     verificationId: VerificationJobId,
     functionName: string,
     input:
       | VerifyFromJsonInput
       | VerifyFromMetadataInput
       | VerifyFromEtherscanInput,
-  ): Promise<void> {
+  ): void {
```
kuzdogan, line 355: Good catch, how did you recognize this? 

Actually this makes me realize now this function is difficult to read with .then()s. It's ok though.

services/server/test/integration/storageServices/SourcifyDatabaseService.spec.ts
```diff
@@ -46,4 +65,119 @@ describe("SourcifyDatabaseService", function () {
...
+    );
+    const storeSignature = signatures.find(
+      (s) => s.signature === "store(uint256)",
+    );
+
+    expect(retrieveSignature).to.exist;
+    expect(storeSignature).to.exist;
+
+    const expectedRetrieveSignatureHash32 = bytesFromString(
+      keccak256str(retrieveSignature!.signature),
```
kuzdogan: Isn't this self referencing? This should be calculated from teh MockVerificationExport or raw strings above?

kuzdogan (commented):
Nice and simple. Just a few comments

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
manuelwedler: good catch. `super` can actually stay. I think it's a leftover from Claude which implemented the feature in a different way at first.

services/server/src/server/services/VerificationService.ts
manuelwedler: I just read over the function and wondered why the function is async. 

We mainly have the `then()` style here because we need to keep a reference to the promise object. We could move the inner logic to another function to use `await`, but I think it's fine.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
manuelwedler: Yes, I did this on purpose to keep the code shorter. IMO, when you accept an object typed via an interface, you should also expect the object to possibly have more properties, because TypeScript doesn't guarantee the absence of other properties ("duck typing").

services/server/src/server/services/utils/signature-util.ts
manuelwedler: Thanks for pointing this out. I made the function throwing in this case now, so we would not store any of the signatures if only one custom type was found. Or was your intention rather to ignore the fragments with custom types?

services/server/src/server/services/utils/signature-util.ts
```diff
@@ -9,16 +9,21 @@ export interface SignatureData {
...
+        case "error":
+          signatures.push(getSignatureData(fragment));
+      }
     }
-  });
+  } catch (error) {
+    throw new Error(
+      "Failed to extract signatures from ABI due to an invalid fragment",
+    );
+  }
```
kuzdogan: This should not throw, or if it does I don't see it being caught anywhere. If there's an invalid type e.g. when it's a library contract ABI we should simply skip. Right now this throws here https://github.com/argotorg/sourcify/blob/6bfbf5f848e869451a294475797efd1017f1fb8b/services/server/src/server/services/storageServices/SourcifyDatabaseService.ts#L913-L914

kuzdogan (changes requested):

services/server/src/server/services/utils/signature-util.ts
manuelwedler: I still didn't understand precisely how you want to handle this case. If it is a library contract, do you want to skip storing all signatures of the ABI, or do you want to just skip the ABI fragments with custom types and still store the other signatures? Both options seem reasonable to me.

The `storeSignatures` function actually has a try-catch which is wrapping the lines you referenced here.

kuzdogan: Hmm I'd say just skip the ABI fragments that are invalid ie. not an "error", "function", or "event". This would also include the constructors. So if a library has valid signatures we will save them. 

From my understanding, as is, this code would break at the invalid signature and won't save the rest of the valid signatures. Ie if a contract has signatures of type:

1. `function`
2. `InvalidType`
3. `function`

~~it will break at 2. and won't save 3., no?~~

Edit: actually it will break at 2. and won't save any of the signatures. I think we should save the valid ones

kuzdogan: We can just silently skip processing if the `fragment.type` is not one of the valid ones and won't enter it to the `getSignatureData` which will throw via ethersjs.

manuelwedler: Alright sounds good to me. I just think we should ignore inside the `getSignatureData` function and not throw from it in any case then.

PR #2364, opened 2025-09-09
Add BigQuery integration with API endpoint for executing queries

marcocastignoli:


marcocastignoli:
Instead of including the bigquery endpoint in server, we decided to implement it in a separate service: github.com/sourcifyeth/bigquery-api

PR #2368, opened 2025-09-11
Remove `update_at` column from signature table migration

manuelwedler:
We decided that it's not necessary to have it on the signature table as a signature should never be updated.


manuelwedler:
already did so :+1:

PR #2372, opened 2025-09-15
Replace Blast API with Alchemy, add Celo Sepolia

kuzdogan:


manuelwedler (commented):
We also have blast api on soucify server on Gnosis chain / 100.

kuzdogan:
Good catch, replaced them

PR #2373, opened 2025-09-16
4byte read endpoints

### LINKED ISSUE #2318: 4bytes: Read logic and endpoints
...

Also include the https://github.com/openchainxyz/canonical-signatures for filtering

manuelwedler:
Closes #2318 

I tried the endpoints to mimic the current behavior of the OpenChain API. The only addition I made is introducing `error` properties to the results to also return error signatures (OpenChain doesn't have error signatures).

Notes:

- OpenChain returns a 500 for any error it encounters. Therefore, the validation functions here also return 500 in case of a user error. It would actually be more correct to return a 4XX status code. I prioritized being as close to OpenChain API as possible.
- I added a new index to the migrations because all queries rely on the `signature_type` property. With the new index, the queries were much faster in my trials.
- I added the canonical signatures from https://github.com/openchainxyz/canonical-signatures to a json file. Potentially, we can manually add more signatures we want to filter for there. I also checked staging db for duplicate signatures, and there were a couple of them. So this might be needed actually.


services/server/src/server/signature-api/openchain.validation.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const hashes = [
+    ...(functionQuery?.split(",") || []),
+    ...(eventQuery?.split(",") || []),
+    ...(errorQuery?.split(",") || []),
+  ];
+
+  for (const hash of hashes) {
+    if (!validateHash(hash)) {
+      sendSignatureApiFailure(res, "failed to load signatures");
```
kuzdogan: I see the reason we don't throw a BadRequest here is to mimic the Openchain API but at least we can return a string meaningful in this context, eg. "Invalid has strings"

services/server/src/server/signature-api/routes.ts
```diff
@@ -0,0 +1,37 @@
...
+    );
+    return;
+  }
+  next();
+});
+
+// Routes to mimic the OpenChain API: https://github.com/openchainxyz/openchain-monorepo
+router.route("/v1/lookup").get(validateHashQueries, lookupSignatures);
+router.route("/v1/search").get(validateSearchQuery, searchSignatures);
+router.route("/v1/stats").get(getSignaturesStats);
```
kuzdogan, line 35: Does this route run in reasonable time or do we need some materialized view etc?

services/server/src/server/signature-api/openchain.validation.ts
```diff
@@ -0,0 +1,67 @@
...
+  } = req.query;
+
+  const hashes = [
+    ...(functionQuery?.split(",") || []),
+    ...(eventQuery?.split(",") || []),
+    ...(errorQuery?.split(",") || []),
+  ];
+
+  for (const hash of hashes) {
+    if (!validateHash(hash)) {
```
kuzdogan, line 45: Should we separately check if the values under `function` queryParam are actually 4 bytes? This also accepts 32byte queries. How is it in openchain?

services/database/migrations/20250828092603_add_signature_tables.sql
```diff
@@ -49,6 +49,7 @@ CREATE TABLE compiled_contracts_signatures (
 );
 
 CREATE INDEX compiled_contracts_signatures_signature_idx ON compiled_contracts_signatures (signature_hash_32);
+CREATE INDEX compiled_contracts_signatures_type_signature_idx ON compiled_contracts_signatures (signature_type, signature_hash_32);
```
kuzdogan, line 46: We should not forget to run this for staging DB.

kuzdogan: We need `type` as an index on its own for /stats right?

services/server/src/server/signature-api/openchain.validation.ts
kuzdogan: Also to check if errors are 4 byte

services/server/src/signature-database.yaml
```diff
@@ -0,0 +1,149 @@
...
+                properties:
+                  ok:
+                    type: boolean
+                  result:
+                    $ref: "#/components/schemas/SignatureResponse"
+
+  /signature-database/v1/search:
+    get:
+      summary: Search signatures
+      description: Search signatures by name with wildcards, limited to the first 100 unordered results
```
kuzdogan: Let's add that it's case insensitive

services/server/src/server/services/utils/Database.ts
```diff
@@ -852,37 +933,6 @@ ${
     return verifiedContractsInsertResult;
   }
 
-  async updateContractDeployment(
```
kuzdogan, line 855: Why are we removing this?

services/server/src/server/routes.ts
```diff
@@ -56,5 +57,6 @@ router.get("/chains", (_req, res) => {
 
 router.use("/", apiV1Routes);
 router.use("/v2", apiV2Routes);
+router.use("/signature-database", signatureApiRoutes);
```
kuzdogan, line 60: I'd suggest we name it `/signatures` to make it shorter or `/openchain` to make it descriptive of the openchain API

services/server/src/server/signature-api/routes.ts
manuelwedler: I tried it on the current staging db and it took roughly 5 seconds to respond. I would still try how long it takes in production when we have all the data there. If we think it's too slow we can implement a materialized view.

services/server/src/server/signature-api/openchain.validation.ts
manuelwedler: Yes, this is true. It is possible to lookup each type via 32 or 4 byte hash. In OpenChain, functions can only be looked up by 4 byte hashes and events by 32 byte hashes. OpenChain actually stores them in separate tables and only one of the hashes for each type. Therefore, it has this limitation. 

See these queries for reference: https://github.com/openchainxyz/openchain-monorepo/blob/a348474d05524982e2d370755d3be0b24a9c71d2/services/signature-database-srv/database/database.go#L24C1-L25C112
They are separated by signature type.

I thought it's fine to be more permissive here. If you disagree, we can also implement the checks.

services/server/src/server/services/utils/Database.ts
manuelwedler: This is an unused function, just to clean up here.

services/server/src/server/signature-api/openchain.validation.ts
kuzdogan: Yeah actually it's fine this way either

services/server/src/server/routes.ts
manuelwedler: It's already `signature-database` in OpenChain, e.g. `https://api.openchain.xyz/signature-database/v1/lookup`. If we want people to only have to replace the base url, we should keep the `signature-database` path.

services/database/migrations/20250828092603_add_signature_tables.sql
manuelwedler: I actually already added it to test performance of it. 

> We need `type` as an index on its own for /stats right?

No, composite indexes also improve performance of queries where only the leftmost column of the composite is used.

manuelwedler: Ah but forgot to update sourcify-database.sql in this PR

services/server/src/server/routes.ts
kuzdogan: Ah ok I wasn't aware of it. All good then

PR #2377, opened 2025-09-17
Add fields and selectors for querying signatures via contract lookup API

### LINKED ISSUE #2320: 4bytes: Add an additional `?fields` parameter for getting signatures of this contract

manuelwedler:
Closes #2320 

When requested via `?fields=signatures` (or `?fields=all`), the signatures are returned like in this example:

```json
"signatures": {
  "function": [
    {
      "signature": "name()",
      "signatureHash32": "0x06fdde0383f15d582d1a74511486c9ddf862a882fb7904b3d9fe9b8b8e58a796",
      "signatureHash4": "0x06fdde03"
    },
    {
      "signature": "deprecate(address)",
      "signatureHash32": "0x0753c30c37785fe67c057e37b49f6f8d5a438a94f4d86e2f327454f9c52bf927",
      "signatureHash4": "0x0753c30c"
    }
  ],
  "event": [
    {
      "signature": "AddedBlackList(address)",
      "signatureHash32": "0x42e160154868087d6bfdc0ca23d96a1c1cfa32f1b72ba9ba27b69b98a0d819dc",
      "signatureHash4": "0x42e16015"
    }
  ],
  "error": []
}
```

It is possible to query for subfields like `?fields=signatures.event`.

I made the response intentionally more verbose by including the 32 and 4 byte hashes, even if the later could just be calculated on the requester's end. The reason is that this feature is mainly for making things easier and all of this could also be generated from the `abi` field.

### DIFF
--- a/services/server/src/apiv2.yaml
+++ b/services/server/src/apiv2.yaml
@@ -916,7 +916,10 @@ components:
                     - signature: "AddedBlackList(address)"
                       signatureHash32: "0x42e160154868087d6bfdc0ca23d96a1c1cfa32f1b72ba9ba27b69b98a0d819dc"
                       signatureHash4: "0x42e16015"
-                  error: []
+                  error: 
+                    - signature: "ERC20InsufficientAllowance()"
+                      signatureHash32: "0x890e98490493b2eeaa713ba179e7b5e82994498e2a4fd1cb19df5f91a38c1532"
+                      signatureHash4: "0x890e9849"
                 proxyResolution:
                   isProxy: false
                   proxyType: null
--- a/services/server/src/server/types.ts
+++ b/services/server/src/server/types.ts
@@ -21,7 +21,7 @@ import { Response } from "express";
 import { JsonFragment } from "ethers";
 import { ProxyDetectionResult } from "./services/utils/proxy-contract-util";
 import { GenericErrorResponse, MatchingErrorResponse } from "./apiv2/errors";
-import { SignatureData, SignatureType } from "./services/utils/signature-util";
+import { SignatureType } from "./services/utils/signature-util";
 
 // Types used internally by the server.
 


services/server/src/apiv2.yaml
```diff
@@ -885,6 +904,19 @@ components:
...
+                      signatureHash32: "0x06fdde0383f15d582d1a74511486c9ddf862a882fb7904b3d9fe9b8b8e58a796"
+                      signatureHash4: "0x06fdde03"
+                    - signature: "deprecate(address)"
+                      signatureHash32: "0x0753c30c37785fe67c057e37b49f6f8d5a438a94f4d86e2f327454f9c52bf927"
+                      signatureHash4: "0x0753c30c"
+                  event:
+                    - signature: "AddedBlackList(address)"
+                      signatureHash32: "0x42e160154868087d6bfdc0ca23d96a1c1cfa32f1b72ba9ba27b69b98a0d819dc"
+                      signatureHash4: "0x42e16015"
+                  error: []
```
kuzdogan: Why left empty?

services/server/src/server/types.ts
```diff
@@ -21,6 +21,7 @@ import { Response } from "express";
 import { JsonFragment } from "ethers";
 import { ProxyDetectionResult } from "./services/utils/proxy-contract-util";
 import { GenericErrorResponse, MatchingErrorResponse } from "./apiv2/errors";
+import { SignatureData, SignatureType } from "./services/utils/signature-util";
```
kuzdogan: `SignatureData` is unused

kuzdogan (approved):
Nice and clean. Just two comments

services/server/src/apiv2.yaml
manuelwedler: I just didn't have a contract at hand with error signatures. I'll find one and add the example

kuzdogan (approved):

PR #2383, opened 2025-09-22
Optimize signature search and materialized view for /stats

kuzdogan:
- Adds GIN index to `signatures.signature` for fast `LIKE` search
- Enables `pg_trgm` for efficient trigram search
- Handles `_` escape
- Adds and underscore function name test and updates other tests to be independent of the order of inserted functions.
- Creates a materialized view `sourcify_stats` for the `/stats` endpoint to use. Adds when this materialized view was last updated and adds that field to the /stats response.
- Adds `pg_cron` to update the stats daily

TODO:
- [x] Need to enable `pg_cron` on GCP postgres instances
- [x] Apply the migrations on live DBs

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -262,7 +262,7 @@ jobs:
   tests-node-v22:
     docker:
       - image: cimg/node:22.5.1
-      - image: postgres:15.13-alpine
+      - image: postgres:15-alpine
         environment:
           POSTGRES_DB: sourcify
           POSTGRES_USER: sourcify
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -262,7 +262,7 @@ jobs:
   tests-node-v22:
     docker:
       - image: cimg/node:22.5.1
-      - image: postgres:16-alpine
+      - image: postgres:15.13-alpine
         environment:
           POSTGRES_DB: sourcify
           POSTGRES_USER: sourcify
--- a/services/database/README.md
+++ b/services/database/README.md
@@ -9,6 +9,8 @@ A complete dump of the Sourcify database schema can be found in `./sourcify-data
 
 ## Running the database
 
+We use PostgreSQL 15.13 for the database. Higher versions should also work but are not tested.
+
 ### Run with Docker
 
 For convenience, you can run the Postgres container in `docker-compose.yml` with
--- a/services/database/README.md
+++ b/services/database/README.md
@@ -81,7 +81,7 @@ Please follow these steps:
 
 1. Create a new migration file: `npm run migrate:new <migration_name>`
 2. Add the required SQL for the schema change to the generated migration file (e.g., `./migrations/20250717103432_<migration_name>.sql`).
-3. Apply the new migration to a local database: `npm run migrate:up`. `dbmate` automatically generates the updated `sourcify-database.sql` dump. The there will be an error because the dump is not yet generated. This step will be skipped. You can run `dbmate dump` to generate the dump manually or see the errors.
+3. Apply the new migration to a local database: `npm run migrate:up`. `dbmate` automatically generates the updated `sourcify-database.sql` dump. There won't be an error if the dump cannot be generated. You can run `dbmate dump` to generate the dump manually or see the errors.
 4. Commit both the new migration file and the updated `sourcify-database.sql` to the repository.
 
 Important: Since the schema dump should be committed, ensure that the connected database does not contain any custom schema changes that are not part of the migrations.
--- a/services/database/README.md
+++ b/services/database/README.md
@@ -37,6 +37,8 @@ Please initialize the Verifier Alliance [database-specs](https://github.com/veri
 git submodule update --init
 ```
 
+**Extension**: pg_cron is used to schedule the refresh of the signature stats materialized view.
+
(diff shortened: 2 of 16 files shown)


.circleci/continue_config.yml
```diff
@@ -256,7 +256,7 @@ jobs:
   tests-node-v22:
     docker:
       - image: cimg/node:22.5.1
-      - image: postgres:16-alpine
+      - image: postgres:15.13-alpine
```
manuelwedler: Not sure if we should be specific about the minor version. Google Cloud SQL updates minor versions automatically.

services/database/docker-compose.yml
```diff
@@ -2,7 +2,14 @@ version: "3.8"
...
   db:
-    image: postgres:16-alpine
+    build:
+      context: .
+      dockerfile_inline: |
+        FROM postgres:15.13-bookworm
+        RUN apt-get update && \
+            apt-get -y install postgresql-15-cron && \
+            apt-get clean && \
+            rm -rf /var/lib/apt/lists/*
```
manuelwedler, line 12: Just for my understanding, why do we need to install pg_cron here and not in CircleCI?

services/database/migrations/20250922141802_signature_stats_materialized_view.sql
```diff
@@ -0,0 +1,67 @@
...
+
+-- Remove scheduled job (gracefully handle if pg_cron not available)
+DO $$
+BEGIN
+    PERFORM cron.unschedule('refresh-signature-stats');
+    RAISE NOTICE 'Unscheduled signature stats refresh job';
+EXCEPTION WHEN OTHERS THEN
+    RAISE NOTICE 'pg_cron not available or job not found, continuing with cleanup';
+END
+$$;
```
manuelwedler, line 63: We should also remove pg_cron here, no?

services/database/README.md
```diff
@@ -37,6 +39,8 @@ Please initialize the Verifier Alliance [database-specs](https://github.com/veri
 git submodule update --init
 ```
 
+**Extension**: pg_cron is used to schedule the refresh of the signature stats materialized view. If the pg_cron extension is not available, the migration will be skipped.
```
manuelwedler: As far as I can tell, it only skips setting up the scheduled refresh. So this seems not accurate.

services/database/README.md
```diff
@@ -77,7 +81,7 @@ Please follow these steps:
 
 1. Create a new migration file: `npm run migrate:new <migration_name>`
 2. Add the required SQL for the schema change to the generated migration file (e.g., `./migrations/20250717103432_<migration_name>.sql`).
-3. Apply the new migration to a local database to generate the updated `sourcify-database.sql` dump: `npm run migrate:up`
+3. Apply the new migration to a local database: `npm run migrate:up`. `dbmate` automatically generates the updated `sourcify-database.sql` dump. The there will be an error because the dump is not yet generated. This step will be skipped. You can run `dbmate dump` to generate the dump manually or see the errors.
```
manuelwedler:
```suggestion
3. Apply the new migration to a local database: `npm run migrate:up`. `dbmate` automatically generates the updated `sourcify-database.sql` dump. There won't be an error if the dump cannot be generated. You can run `dbmate dump` to generate the dump manually or see the errors.
```

services/database/docker-compose.yml
kuzdogan: We could get away with not installing here too. I used this to set up my local Postgres and kept it. After that, I couldn't really get this same method to work on the CI so I decided to handle it gracefully in the migration.

services/database/README.md
kuzdogan: wdym? The migration skips if the pg_cron is not installed

services/database/migrations/20250922141802_signature_stats_materialized_view.sql
kuzdogan: Claude tells me not to:

  Why not to drop pg_cron extension:

  1. Shared resource: Extensions are database-wide, not migration-specific. Other migrations or applications might be using pg_cron
  2. Permission issues: Dropping extensions often requires superuser privileges
  3. Data loss risk: pg_cron might have other scheduled jobs that would be lost
  4. Deployment complexity: Extension management is typically handled at the infrastructure level

.circleci/continue_config.yml
kuzdogan: By habit I kept the minor version but 15 should be fine too, changing

services/database/README.md
manuelwedler: The whole migration? I thought it would only skip installing pg_cron and the cron job

services/database/docker-compose.yml
manuelwedler: Okay that makes sense to me. We also don't need to run the cron job in CI.

services/database/README.md
kuzdogan: In services/database/migrations/20250922141802_signature_stats_materialized_view.sql it will skip adding pg_cron to the database and skip adding the cron job.

kuzdogan: So yes not the whole migration haha

manuelwedler (approved):
Looks good, thanks :)

PR #2385, opened 2025-09-23
Properly handle filter query parameter in OpenChain handlers

### LINKED ISSUE #2382: `filter=true` query parameter on `/signature-database` not working
I noticed that the filter parameter has no effect currently.

https://sourcify.dev/server/signature-database/v1/lookup?function=0xa9059cbb&filter=true

Returns:
```json
{
  "ok": true,
  "result": {
  "function": {
    "0xa9059cbb": [
      {
        "name": "transfer(address,uint256)",
        "filtered": false
      },
      {
        "name": "_____$_$__$___$$$___$$___$__$$(address,uint256)",
        "filtered": true
      }
    ]
  },
  "event": {},
  "error": {}
  }
}
```

It should actually not return the second signature.

Besides this, I noticed that the filter function only runs on the function signatures:
https://github.com/argotorg/sourcify/blob/db0469e96458091d7b4be9206a814b417af2b9ea/services/server/src/server/signature-api/openchain.handlers.ts#L40

manuelwedler:
Fixes #2382

### DIFF
--- a/services/database/package.json
+++ b/services/database/package.json
@@ -8,6 +8,7 @@
     "sourcify:database": "node scripts.mjs ",
     "migrate:setup": "rm -rf migrations-temp && mkdir -p migrations-temp && ln -sf $PWD/database-specs/migrations/* $PWD/migrations-temp/ && ln -sf $PWD/migrations/* $PWD/migrations-temp/",
     "migrate:up": "npm run migrate:setup && DBMATE_MIGRATIONS_DIR=./migrations-temp dbmate up",
+    "migrate:up-no-dump": "npm run migrate:setup && DBMATE_MIGRATIONS_DIR=./migrations-temp dbmate --no-dump-schema up",
     "migrate:rollback": "npm run migrate:setup && DBMATE_MIGRATIONS_DIR=./migrations-temp dbmate rollback",
     "migrate:new": "DBMATE_MIGRATIONS_DIR=./migrations dbmate new",
     "migrate:create-db": "dbmate create",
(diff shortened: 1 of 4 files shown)


manuelwedler:
> Can you add a test for filter=false? I'm a bit unsure if the query value filter=false might end up being true because it's not empty string.

There is already this test:
https://github.com/argotorg/sourcify/pull/2385/files#diff-a84c8cad9e93ed89d902b736af67ba8d1cab613b5005cf6b3f35000c69eb1904L218-R244

kuzdogan:
From what I see the test doesn't really test what we want. The test only checks if there's a `filtered` field in the canonical response. We need to check that a spam signature is not included in the response, no?

See
https://api.openchain.xyz/signature-database/v1/lookup?function=0xa9059cbb&filter=false
https://api.openchain.xyz/signature-database/v1/lookup?function=0xa9059cbb&filter=true

manuelwedler:
> The test only checks if there's a `filtered` field in the canonical response.

Are you looking at the right version of the code? This was the old version of the test which is indeed wrong / not sufficient. The new code does:

```ts
      chai.expect(res.body.result.function).to.have.property(hash4);
      chai.expect(res.body.result.function[hash4]).to.be.an("array");
      chai.expect(res.body.result.function[hash4]).to.have.deep.members([
        { name: "transfer(address,uint256)", filtered: false },
        { name: "transfer(bytes4[9],bytes5[6],int48[11])", filtered: true },
      ]);
```


There is also the test `should lookup function signatures by 4-byte hash and filter by default` which checks for filtering of spam signatures. Do you want another test that explicitly sets `filtered=true`?

kuzdogan (approved):
Ok my bad, was looking at the wrong branch... Thanks

PR #2392, opened 2025-09-25
Make pg max connections configurable and set to 15 default

kuzdogan:
While working on the separate 4bytes DB I came across the code where we create the DB pool with a max connection `max: 5`. 

I tried tracking down why this was set to 5 (default is 10). I found it was set 2 years ago here and kept the same but couldn't see why this was set to a lesser value https://github.com/argotorg/sourcify/blob/9e96c8e6412f7ab1b1aedd74cb1ed9432e67aa8d/src/server/services/alliance-database-util.ts

I believe this was a reason why the responses to the signature queries started getting longer when we had the surge. Our prod. db has a lot more max connections available and if needed this should have been utilized.

image

Therefore I'm setting this to a higher value, and making it configurable.


manuelwedler (approved):
Looks good. Let's see how it performs

PR #2396, opened 2025-09-25
Remove buggy the refresh_materialized_view function, raise the pg notice to log level

kuzdogan:
Turns out the original function we wrote was buggy as you can't really UPDATE a materialized view. That's why the refresh failed last night. We can completely get rid of the function and just schedule a direct refresh of the materialized view, and it will take care of the `refreshed_at` columns itself

TODO:
- [x] Execute fix queries on staging
- [ ] Execute fix queries on prod

### DIFF
--- a/services/database/README.md
+++ b/services/database/README.md
@@ -41,7 +41,7 @@ git submodule update --init
 
 **Extensions**
 
-- `pg_cron`: used to schedule the refresh of the signature stats materialized view. Make sure you [install the `pg_cron` extension](https://github.com/citusdata/pg_cron) abd set the `cron.database_name` variable to the name of the database you are using.
+- `pg_cron`: used to schedule the refresh of the signature stats materialized view. Make sure you [install the `pg_cron` extension](https://github.com/citusdata/pg_cron) and set the `cron.database_name` variable to the name of the database you are using.
   - If the `pg_cron` extension is not available, adding `pg_cron` and creating the cron job will be skipped in the migration.
   - In Google Cloud SQL, you can install the extension by setting the flag `cloudsql.enable_pg_cron` to `true`, and setting the `cron.database_name` flag to the name of the database you are using.
 - `pg_trgm`: used to create the index on the `signatures` table.
--- a/services/database/README.md
+++ b/services/database/README.md
@@ -39,7 +39,13 @@ Please initialize the Verifier Alliance [database-specs](https://github.com/veri
 git submodule update --init
 ```
 
-**Extension**: `pg_cron` is used to schedule the refresh of the signature stats materialized view. If the `pg_cron` extension is not available, adding `pg_cron` and creating the cron job will be skipped in the migration.
+**Extensions**
+
+- `pg_cron`: used to schedule the refresh of the signature stats materialized view. Make sure you [install the `pg_cron` extension](https://github.com/citusdata/pg_cron) abd set the `cron.database_name` variable to the name of the database you are using.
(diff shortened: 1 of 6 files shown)


manuelwedler:
Did you already apply the migration to production db before a release?

Wondering because you are saying we need to execute the fix on production, but I expect that before a release we haven't run the related migration yet.

manuelwedler:
> it will take care of the `refreshed_at` columns itself

How does this work? Is this just default behavior and it expects `refreshed_at` column to be present?

kuzdogan:
> Did you already apply the migration to production db before a release?
> 
> Wondering because you are saying we need to execute the fix on production, but I expect that before a release we haven't run the related migration yet.

I did run the migration on prod after this PR https://github.com/argotorg/sourcify/pull/2383#issue-3441420739 to not forget about it at the release. Maybe it was better to wait for the release but any idea how we could avoid forgetting about it? 



> > it will take care of the `refreshed_at` columns itself
> 
> How does this work? Is this just default behavior and it expects `refreshed_at` column to be present?

It's already defined at how we create the materialized view. The date is set to `now()`

https://github.com/argotorg/sourcify/blob/282d5f27cc8360c33e80f63234850ee2aaab307e/services/database/migrations/20250922141802_signature_stats_materialized_view.sql#L15-L22

manuelwedler:
> It's already defined at how we create the materialized view. The date is set to `now()`

But doesn't that mean that `created_at` will also be updated? In that case, we can remove one of them.



> I did run the migration on prod after this PR [#2383 (comment)](https://github.com/argotorg/sourcify/pull/2383#issue-3441420739) to not forget about it at the release. Maybe it was better to wait for the release but any idea how we could avoid forgetting about it?

Yes, I think we should wait until a release, because we might make fixes to the migrations. It depends on the migration, but on staging it should usually be fine to rollback the latest migration and reapply with a fix, but on production I would avoid this. 
Anyway, we could just add a step in the release script to remind us. It should be the step before merging to main, since this is the point when the production deployment is updated.

I also thought of automating running the migrations, but Marco raised some concerns.

kuzdogan:
> But doesn't that mean that created_at will also be updated? In that case, we can remove one of them.

Yea very good point

kuzdogan:
It was more complicated than I thought as the pg_cron by default runs on the `postgres` database and sourcify user does not have permissions.

I've set the default database to `sourcify` instead of `postgres` via the GCP flag `cron.databasename=sourcify`.

I ran the manual queries to set up the updated materialized view and the cron job. I'll wait until tomorrow for the cron job to run before merging.

services/database/README.md
```diff
@@ -39,7 +39,13 @@ Please initialize the Verifier Alliance [database-specs](https://github.com/veri
 git submodule update --init
 ```
 
-**Extension**: `pg_cron` is used to schedule the refresh of the signature stats materialized view. If the `pg_cron` extension is not available, adding `pg_cron` and creating the cron job will be skipped in the migration.
+**Extensions**
+
+- `pg_cron`: used to schedule the refresh of the signature stats materialized view. Make sure you [install the `pg_cron` extension](https://github.com/citusdata/pg_cron) abd set the `cron.database_name` variable to the name of the database you are using.
```
manuelwedler:
```suggestion
- `pg_cron`: used to schedule the refresh of the signature stats materialized view. Make sure you [install the `pg_cron` extension](https://github.com/citusdata/pg_cron) and set the `cron.database_name` variable to the name of the database you are using.
```

kuzdogan:
Merging. Executing queries after running on staging. Added a card on the board

PR #2398, opened 2025-09-26
Remove old Celo testnets

karlb:
Since Holesky will be shut down end of September, the L2 testnets that use Holesky as L1 will also stop working. Let's clean up and remove the entries for the Holesky-based Celo networks.

### DIFF
--- a/services/server/src/sourcify-chains-default.json
+++ b/services/server/src/sourcify-chains-default.json
@@ -388,29 +388,6 @@
       }
     ]
   },
-  "44787": {
-    "sourcifyName": "Celo Alfajores Testnet",
-    "supported": true,
-    "etherscanApi": {
-      "supported": true,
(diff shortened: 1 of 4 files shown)


services/server/src/sourcify-chains-default.json
```diff
@@ -388,29 +388,6 @@
       }
     ]
   },
-  "44787": {
-    "sourcifyName": "Celo Alfajores Testnet",
-    "supported": true,
```
kuzdogan, line 393: Can you, for both chains, turn `supported: false`, just keep `sourcifyName` in addition, and remove other fields?

kuzdogan (changes requested):
Small change

services/server/src/sourcify-chains-default.json
karlb: Sure, done.

PR #2402, opened 2025-09-30
Create 4byte service

kuzdogan:
Creates a separate service for the 4byte backend.

This also entails a separate Postgres DB for its self-contained tests. That's why you'll see new postgres containers in the CI runs. In the `4byte/test` you'll see the `migrations` folder. Instead of creating the whole Sourcify DB, this only runs the signatures tables migration. Before that we need a stub `compiled_contracts` table because of the existing reference in `compiled_contracts_signatures.compiled_contract`. Then we run the exact migrations in `database/migrations` (via linked files, not copied).

I think I will put this service behind api.4byte.sourcify.dev

### DIFF
--- a/services/4byte/src/cli.ts
+++ b/services/4byte/src/cli.ts
@@ -30,7 +30,18 @@ const server = new FourByteServer({
 // Enable Swagger UI for CLI usage
 const apiSpecPath = path.join(__dirname, "openapi.yaml");
 const openApiSpec = yaml.load(apiSpecPath);
-server.app.use("/api-docs", swaggerUi.serve, swaggerUi.setup(openApiSpec));
+server.app.get("/api-docs/swagger.json", (req, res) => {
+  res.json(openApiSpec);
+});
+server.app.use(
+  "/api-docs",
+  swaggerUi.serve,
+  swaggerUi.setup(openApiSpec, {
+    customSiteTitle:
+      "Sourcify Ethereum Function, Event, and Error Signatures API",
+    customfavIcon: "https://sourcify.dev/favicon.ico",
+  }),
+);
 
 server
   .listen()


marcocastignoli:
From my pov we can create the whole database in the tests. I don't like the idea of managing more migrations. Maybe I'm missing something.

kuzdogan:
@marcocastignoli I don't expect these migrations to change. If we use the same DB for both server and 4byte tests, these will interfere with each other e.g. server test can wipe data for 4byte, and we need to insert full verified contracts for the 4byte tests. Now we can only insert signatures only.

IMO it's better if the 4byte service is fully self contained, read-only, and knows nothing about the verified contracts etc.

services/4byte/src/utils/canonical-signatures.json
```diff
@@ -0,0 +1,137 @@
+{
```
manuelwedler, line 1: Can you also remove this file from the server package?

services/4byte/README.md
```diff
@@ -0,0 +1,109 @@
...
+- **Signature Search**: Search signatures by text patterns
+- **Statistics**: Get signature counts and metadata
+- **Multi-type Support**: Functions, events, and error signatures
+- **Filtering**: Optional filtering for canonical signatures
+
+## API Endpoints
+
+### GET `/signature-database/v1/lookup`
+
+Lookup signatures by hash.xxw
```
manuelwedler:
```suggestion
Lookup signatures by hash.
```

services/4byte/src/utils/signature-util.ts
```diff
@@ -0,0 +1,58 @@
...
+  Error = "error",
+}
+
+export interface SignatureData {
+  signature: string;
+  signatureHash32: string;
+  signatureType: SignatureType;
+}
+
+export function extractSignaturesFromAbi(abi: JsonFragment[]): SignatureData[] {
```
manuelwedler: This function is not used anywhere. We can remove it and the unexported `getSignatureData` too.

services/4byte/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+```bash
+# Install dependencies
+npm install
+
+# Build the service
+npm run build
+
+# Run tests
+npm run test
```
manuelwedler:
```suggestion
npm run test-local
```

services/4byte/test/unit/utils/signature-util.spec.ts
```diff
@@ -0,0 +1,241 @@
+import chai from "chai";
+import {
+  extractSignaturesFromAbi,
+  getCanonicalSignatures,
+} from "../../../src/utils/signature-util";
+import { JsonFragment, id as keccak256str } from "ethers";
+
+describe("4byte signature-util", function () {
+  describe("extractSignaturesFromAbi", function () {
```
manuelwedler: These tests can also be removed

services/4byte/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Environment Variables
+
+- `POSTGRES_HOST` - Database host (default: localhost)
+- `POSTGRES_PORT` - Database port (default: 5432)
+- `POSTGRES_DB` - Database name (default: sourcify)
+- `POSTGRES_USER` - Database user
+- `POSTGRES_PASSWORD` - Database password
+- `POSTGRES_SCHEMA` - Database schema (default: public)
+- `PORT` - Service port (default: 80)
```
manuelwedler: The env variables are prefixed with `FOURBYTES_`. IMO, you can just refer to the `.env.example` file in this section and not list all variables explicitly.

services/4byte/src/FourByteServer.ts
```diff
@@ -0,0 +1,137 @@
...
+  port: string | number;
+  pool: Pool;
+  database: SignatureDatabase;
+  httpServer?: http.Server;
+
+  constructor(options: ServerOptions) {
+    this.app = express();
+    this.port = options.port;
+    logger.info("4Byte Server port set", { port: this.port });
+    this.pool = new Pool(options.databaseConfig);
```
manuelwedler: I think the pool should be created inside the `SignatureDatabase` class and not exposed on the server. I think it would also be cleaner to have all SQL (including the `SELECT 1;` checks) inside the database class. Any reason why it should not be there?

services/4byte/src/FourByteServer.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.database = new SignatureDatabase(this.pool, {
+      schema: options.databaseConfig.schema,
+    });
+
+    // Check database health during initialization
+    this.checkDatabaseHealth();
```
manuelwedler: This returns an unawaited Promise that can lead to an unhandled rejection error. Can we await it? I would recommend a method `static async create(): Promise<FourByteServer>` instead of the constructor to circumvent the sync limitation of the constructor.

services/4byte/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+      ? parseInt(process.env.FOURBYTES_POSTGRES_MAX_CONNECTIONS)
+      : 20,
+    schema: process.env.FOURBYTES_POSTGRES_SCHEMA,
+  },
+});
+
+// Enable Swagger UI for CLI usage
+const apiSpecPath = path.join(__dirname, "openapi.yaml");
+const openApiSpec = yaml.load(apiSpecPath);
+server.app.use("/api-docs", swaggerUi.serve, swaggerUi.setup(openApiSpec));
```
manuelwedler: Can we move these lines inside the server creation method / constructor like all other routes? I think it's an anti-pattern that we do this in the main Sourcify server and we shouldn't follow it.

manuelwedler: Also, is there a reason why we don't serve a swagger.json or swagger.yaml?

services/4byte/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+
+server
+  .listen()
+  .then(() => {
+    logger.info("4byte API server started successfully");
+  })
+  .catch((error) => {
+    logger.error("Failed to start 4byte API server", { error });
```
manuelwedler, line 52: You could call the `shutdown` method of the server here in order to be on the safe side.

services/4byte/src/FourByteServer.ts
```diff
@@ -0,0 +1,137 @@
...
+      handlers.searchSignatures,
+    );
+    this.app.get("/signature-database/v1/stats", handlers.getSignaturesStats);
+
+    this.app.get("/health", async (_req, res) => {
+      await this.pool.query("SELECT 1;");
+      res.status(200).json({ status: "ok", service: "4byte-api" });
+    });
+
+    const shutdown = async (signal: NodeJS.Signals) => {
```
manuelwedler: The `FourByteServer` also has a method `shutdown`. One of them should be removed.

services/4byte/src/openapi.yaml
```diff
@@ -1,4 +1,4 @@
-openapi: 3.0.0
+openapi: 3.1.0
```
manuelwedler, line 1: We can also add the server URLs to this spec.

services/4byte/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+RUN mkdir -p /home/app/services/4byte
+
+WORKDIR /home/app/
+COPY package.json ./package.json
+COPY package-lock.json ./package-lock.json
+COPY lerna.json ./lerna.json
+COPY nx.json ./nx.json
+
+COPY --from=builder /home/app/packages/ ./packages/
```
manuelwedler: I think we don't need to copy from packages here, no?

services/4byte/src/api/handlers.ts
```diff
@@ -0,0 +1,235 @@
...
+interface SignatureResult {
+  function: SignatureHashMapping;
+  event: SignatureHashMapping;
+  error: SignatureHashMapping;
+}
+
+function filterResponse(response: SignatureResult, shouldFilter: boolean) {
+  const canonicalSignatures = getCanonicalSignatures();
+
+  for (const hash in response.function) {
```
manuelwedler: This function should operate on all types of signatures, not only on function. It looks to me like the changes from https://github.com/argotorg/sourcify/pull/2385 are missing

services/4byte/src/FourByteServer.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.app.get(
+      "/signature-database/v1/search",
+      validateSearchQuery,
+      handlers.searchSignatures,
+    );
+    this.app.get("/signature-database/v1/stats", handlers.getSignaturesStats);
+
+    this.app.get("/health", async (_req, res) => {
+      await this.pool.query("SELECT 1;");
+      res.status(200).json({ status: "ok", service: "4byte-api" });
```
manuelwedler: Not sure if we should follow the same pattern as in the other endpoints and return `"ok": true`

services/4byte/src/SignatureDatabase.ts
```diff
@@ -0,0 +1,147 @@
...
+    limit?: number,
+  ): Promise<SignatureSearchRow[]>;
+  getSignatureCounts(): Promise<SignatureStatsRow[]>;
+}
+
+export interface SignatureDatabaseOptions {
+  schema?: string;
+}
+
+export class SignatureDatabase implements SignatureDataProvider {
```
manuelwedler: Can we also remove all related functions from the Sourcify server class `Database`?

services/4byte/src/SignatureDatabase.ts
```diff
@@ -0,0 +1,147 @@
...
+  signature_hash_4: string;
+  signature_hash_32: string;
+}
+
+export interface SignatureStatsRow {
+  signature_type: SignatureType;
+  count: string;
+  created_at: Date;
+  refreshed_at: Date;
+}
```
manuelwedler, line 20:
```suggestion
export interface SignatureStatsRow {
  signature_type: SignatureType;
  count: string;
  refreshed_at: Date;
}
```

services/4byte/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "scripts": {
+    "prebuild": "copyfiles -u 1 src/**/*.yaml src/**/*.json dist",
+    "build": "tsc",
+    "start": "node dist/cli.js",
+    "dev": "ts-node src/cli.ts",
+    "postgres-test:start": "docker-compose -f test/docker-compose.yml up -d",
+    "postgres-test:stop": "docker-compose -f test/docker-compose.yml rm -fsv",
+    "postgres-test:migrate": "export DATABASE_URL=postgres://fourbytes:fourbytes@${FOURBYTES_POSTGRES_HOST:-fourbytes-db}:${FOURBYTES_POSTGRES_PORT:-5432}/fourbytes_test_db?sslmode=disable && export DBMATE_MIGRATIONS_DIR=./test/migrations && npx dbmate up",
+    "test:unit": "mocha test/unit/**/*.spec.ts",
+    "test": "npm run postgres-test:migrate & npx c8 --reporter=none mocha test/unit/**/*.spec.ts test/integration/4byte-api.spec.ts",
```
manuelwedler:
```suggestion
    "test": "npm run postgres-test:migrate & npx c8 --reporter=none mocha test/unit/**/*.spec.ts test/integration/**",
```

The test wasn't running in CI:
`sourcify-4byte: Warning: Cannot find any files matching pattern "test/integration/4byte-api.spec.ts"`

services/4byte/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "main": "dist/cli.js",
+  "types": "dist/cli.d.ts",
+  "scripts": {
+    "prebuild": "copyfiles -u 1 src/**/*.yaml src/**/*.json dist",
+    "build": "tsc",
+    "start": "node dist/cli.js",
+    "dev": "ts-node src/cli.ts",
+    "postgres-test:start": "docker-compose -f test/docker-compose.yml up -d",
+    "postgres-test:stop": "docker-compose -f test/docker-compose.yml rm -fsv",
+    "postgres-test:migrate": "export DATABASE_URL=postgres://fourbytes:fourbytes@${FOURBYTES_POSTGRES_HOST:-fourbytes-db}:${FOURBYTES_POSTGRES_PORT:-5432}/fourbytes_test_db?sslmode=disable && export DBMATE_MIGRATIONS_DIR=./test/migrations && npx dbmate up",
```
manuelwedler:
```suggestion
    "postgres-test:migrate": "export DATABASE_URL=postgres://fourbytes:fourbytes@${FOURBYTES_POSTGRES_HOST:-fourbytes-db}:${FOURBYTES_POSTGRES_PORT:-5432}/fourbytes_test_db?sslmode=disable && export DBMATE_MIGRATIONS_DIR=./test/migrations && npx dbmate --no-dump-schema up",
```

services/4byte/test/migrations/20250925000003_remove_compiled_contracts_dependency.sql
```diff
@@ -0,0 +1,23 @@
+-- migrate:up
+
+-- Remove dependency on compiled_contracts table for 4byte service
+
+-- Drop the foreign key constraint (keep the column for mock data)
+ALTER TABLE compiled_contracts_signatures
+DROP CONSTRAINT IF EXISTS compiled_contracts_signatures_compilation_id_fkey;
```
manuelwedler: If you just run these two lines from the `FourByteServerFixture`, you don't need to add any migrations inside the tests and can just create the whole sourcify database in the usual way. 

In my understanding, the 4byte service isn't self-contained by design, since the signature types are based on compiled contracts in our db. Of course, we should run separate postgres instances for sourcify server and 4bytes tests, but I don't think it's necessary to have custom migrations for this.

services/4byte/src/FourByteServer.ts
kuzdogan: I'd refrain from using a separate create/init function for such a simple service. I can wrap this in a try catch and exit if it fails

services/4byte/test/migrations/20250925000003_remove_compiled_contracts_dependency.sql
kuzdogan: Good idea 👍

services/4byte/src/FourByteServer.ts
kuzdogan: Maybe we should also have a generic error handler?

services/4byte/src/cli.ts
kuzdogan: I missed .json and .yaml so good to add those.

The reason we do it in CLI in the server is because resolving $ref's in the .yaml is async and it's just easier in the CLI. I don't see the use case with the Server module and maybe even better that the server doesn't serve an extra endpoint if you're running an embedded Sourcify server etc. instead of running it from CLI. Why would it be an anti-pattern?

manuelwedler: > Why would it be an anti-pattern?

I think it's unexpected to add any routes outside of the server abstraction. At least, I found it very confusing where this route came from when I was looking at the Sourcify server code at the first time. For the scenario you are describing, I'd prefer to have a parameter to disable it.

Anyway, let's not spend too much time on this. I know about the issue of the async ref parsing. If we need it here too and you don't like the async create mehod, just keep the code in the cli.

services/4byte/src/FourByteServer.ts
manuelwedler: > I'd refrain from using a separate create/init function for such a simple service. I can wrap this in a try catch and exit if it fails

The issue is that you can't wrap it since it's asynchronous. I think an async create method would simplify things further here, since you could also already call `listen` inside it. Anyway, if you don't like it, you could call `catch` on the promise and do a process.exit(1) to handle gracefully. Or just remove the check, I don't think it is required to have.

> Maybe we should also have a generic error handler?

I think the default express handler is enough (https://expressjs.com/en/guide/error-handling.html#the-default-error-handler). But it doesn't help with unhandled rejection errors which just kill the app.

manuelwedler:
@kuzdogan left two comments on the open conversations. Other than that it looks good :)

services/4byte/src/FourByteServer.ts
kuzdogan: I'm adding a `catch`. I think it's needed because you wouldn't know the DB is not there until you call an endpoint. That helps a lot in server.

services/4byte/src/cli.ts
kuzdogan: I added the .json route but I couldn't find where we serve the .yaml in the server. Do we do it on the sourcify-server?

kuzdogan:
Final comments approval @manuelwedler

services/4byte/src/cli.ts
manuelwedler: I think we only serve the json from Sourcify server. I was asking about the yaml because OpenChain serves only the yaml. But not necessary to have the same format probably.

PR #2404, opened 2025-10-01
Introduce Circuit Breaker pattern for failing rpcs on `SourcifyChain` and restructure rpc properties

manuelwedler:
Closes https://github.com/argotorg/sourcify/issues/2163

Tackling this issue turned into a bigger change, but I think the refactoring here is needed and was worth to tackle while working on this.

I think it's not possible to tackle the issue cleanly without modifying the way we store the rpcs and providers inside `SourcifyChain`. So, I took the chance and refactored all the rpc properties and the providers into one array that tracks all information related to an rpc inside a single object. This is a breaking change for anyone using lib-sourcify of course, so I think we should raise the major version for it when we release this.

The circuit breaker pattern works this way that an rpc is not called again after it failed. After a timeout the rpc can be retried. The duration of the timeout depends on how many consecutive failures happened. When calling an rpc method on `SourcifyChain`, we still iterate over the array of rpcs to find a working one but skip any rpc that is currently in blocked state.

### DIFF
--- a/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
+++ b/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
@@ -32,6 +32,8 @@ export function createFetchRequest(rpc: FetchRequestRPC): FetchRequest {
   return ethersFetchReq;
 }
 
+class RpcFailure extends Error {}
+
 export type SourcifyChainMap = {
   [chainId: string]: SourcifyChain;
 };
@@ -206,12 +208,12 @@ export class SourcifyChain {
             maskedUrl: rpc.maskedUrl,
             chainId: this.chainId,
           });
-          this.recordRpcSuccess(rpc);
+          // Don't record success here, as RPC might have been skipped in this case
           continue;
+        } else if (result !== undefined) {
+          this.recordRpcSuccess(rpc);
+          return result as T;
         }
-
-        this.recordRpcSuccess(rpc);
-        return result as T;
       } catch (err) {
         logError('RPC operation threw error', {
           operation: operationName,
@@ -329,9 +331,13 @@ export class SourcifyChain {
       );
     }
 
-    // Try sequentially all RPCs with trace support
-    for (const rpc of this.rpcs) {
-      if (!rpc.traceSupport || !rpc.provider) continue;
+    return this.executeWithCircuitBreaker(async (rpc) => {
+      if (!rpc.provider) {
+        throw new Error('No provider configured for this RPC');
+      }
+      if (!rpc.traceSupport) {
+        return { tryNext: true };
+      }
 
       const { traceSupport: type } = rpc;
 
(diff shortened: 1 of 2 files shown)


services/server/src/sourcify-chains.ts
```diff
@@ -212,18 +226,13 @@ for (const chain of allChains) {
...
 
-    let rpc: (string | FetchRequestRPCWithHeaderEnvName)[] = [];
-    let rpcWithoutApiKeys: string[] = [];
-    let rpcWithApiKeyMasked: string[] = [];
-    let traceSupportedRPCs: TraceSupportedRPC[] | undefined = undefined;
+    let rpcs: SourcifyRpc[] = [];
     if (sourcifyExtension.rpc) {
-      ({ rpc, rpcWithoutApiKeys, rpcWithApiKeyMasked, traceSupportedRPCs } =
-        buildCustomRpcs(sourcifyExtension.rpc));
+      rpcs = buildCustomRpcs(sourcifyExtension.rpc);
```
marcocastignoli, line 264: If I understood correctly how it works, previously we were setting in `buildCustomRpcs` the rpc object coming directly from `sourcifyExtension`. So in the code below (239-247), when we update the headerValue the object was updated both in `sourcifyExtension` and `rpcs` because it is a linked object. 

Now inside `buildCustomRpcs` we are creating a new object so the loop at lines 239-247 only updates the object in `sourcifyExtension` but not `rpcs`
```ts
[...]
 } else if (sourcifyRpc.type === "FetchRequest") {
[...]
      rpcs.push({
        rpc: fetchRequestRpc,
        urlWithoutApiKey: sourcifyRpc.url,
        maskedUrl: sourcifyRpc.url,
        traceSupport: sourcifyRpc.traceSupport,
      });
}
```

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
```diff
@@ -149,92 +240,76 @@ export class SourcifyChain {
...
-            creatorTxHash,
-            maskedProviderUrl: provider.maskedUrl,
-            chainId: this.chainId,
-            error: err.message,
-          });
-          continue;
-        } else {
-          throw err;
-        }
+        return { rpcFailed: true };
```
marcocastignoli: I think here it makes sense to log the error or at least understand why the rpc is failing.

services/server/src/sourcify-chains.ts
```diff
@@ -238,14 +247,14 @@ for (const chain of allChains) {
       }
     });
```
marcocastignoli: It's not strictly related to this PR, but I think it makes sense to log if the process.env variable doesn't exist so it's clear the the header value was set to "" because a missing var

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
manuelwedler: There was already a log in `executeWithCircuitBreaker` but it is missing the error object. I decided to return the error instead of `rpcFailed` in order to add it in the existing log.

services/server/src/sourcify-chains.ts
manuelwedler: I think that's true... very good catch!

manuelwedler:
@marcocastignoli I addressed the feedback and noticed that I forgot one method (getCreationBytecodeForFactory). Added a commit to also support it.

Edit: also fixed the failing test now.

marcocastignoli (approved):
lgtm! The refactoring was needed, thanks :)

Issue #2422, opened 2025-10-07
Explore easy options to reduce GCP costs


manuelwedler:
Todos:
-  [x] scale down vera sql instances?
- [x] close python parquet export
- [x] close filestore NAS
- [x] check why test vera SQL instance has an unexpectedly high storage consumption
- [x] label resources for production and staging on VerA  and Sourcify
- [x] set price alerts

kuzdogan:
I'm prioritizing this to no waste more money

manuelwedler:
Also check why test vera SQL instance has an unexpectedly high storage consumption

kuzdogan:
I think I took this over inadvertently, assigning myself.

kuzdogan:
Turns out the large storage on test-verifier-alliance was some dangling replication slots (when I tried to do parquet exports via Fivetran etc. data processing services). The storage should reduce to a reasonable amount

PR #2428, opened 2025-10-08
Turn off writing to repositoryV1 and V2

kuzdogan:
Following #2422 we are turning off the NAS and the filesystem, as they are not required. We will continue writing files to IPFS via the S3StorageService.


manuelwedler:
What do we do about the people that still depend on the file export? 
https://github.com/argotorg/sourcify/issues/2235

kuzdogan:
We won't delete the files for a while. We already communicated them that at a point there won't be new contracts and they were fine with it.

manuelwedler:
okay sounds good

Issue #2423, opened 2025-10-07
Implement EtherscanVerificationStorageService to verify on external verifiers


marcocastignoli:
See #2414

manuelwedler:
Maybe it's helpful to have a look at the implementation in the Remix plugin:
https://github.com/remix-project-org/remix-project/blob/master/apps/contract-verification/src/app/Verifiers/EtherscanVerifier.ts

especially for error handling

PR #2430, opened 2025-10-09
Implement EtherscanVerifyAPIService

marcocastignoli:
See #2423

TODO Before merging:
- [x] add new `ETHERSCAN_API_KEY_MAINNET` to GCP
- [x] execute migration on staging database


services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+export interface EtherscanVerifyAPIServiceOptions {
+  /** Mapping of chainId to the explorer API base URL */
+  chainApiUrls?: ChainApiUrls;
+  /** Optional mapping of chainId to API keys */
+  apiKeys?: Record<string, string>;
+  /** Optional fallback API key when chain-specific key is missing */
+  defaultApiKey?: string;
+}
+
+export class EtherscanVerifyAPIService implements WStorageService {
```
manuelwedler: I think we should be consistent about the naming style.

```suggestion
export class EtherscanVerifyApiService implements WStorageService {
```

I know there are different approaches to acronyms for camel case. I personally prefer to treat acronyms like any other word. In any case, we should try to be consistent about this, in the same file at the very least (you are writing `getEtherscanApiKeyEnvironmentVariables` and `EtherscanRpcResponse`).

services/server/src/server/services/storageServices/identifiers.ts
```diff
@@ -7,6 +7,9 @@ export enum WStorageIdentifiers {
   AllianceDatabase = "VerifierAllianceDatabase",
   RepositoryV2 = "RepositoryV2",
   S3Repository = "S3Repository",
+  EtherscanVerify = "EtherscanVerify",
```
manuelwedler, line 10: I think we can already enable them in the configuration files for staging and production

services/server/src/server/services/utils/Database.ts
```diff
@@ -996,6 +997,47 @@ ${
     );
   }
 
+  async upsertExternalVerification(
+    verificationJobId: Tables.VerificationJob["id"],
```
manuelwedler, line 1002: can you also update the `Tables.VerificationJob` type in `database-utils` to include the new column?

services/database/migrations/20251009141621_add_external_verification.sql
```diff
@@ -0,0 +1,7 @@
+-- migrate:up
+
+ALTER TABLE verification_jobs ADD external_verification json NULL;
```
manuelwedler: I tend to say that we can use `jsonb` for the column, because we might do some queries on the keys and query by verifier. Or what do you think?

services/server/src/server/services/utils/Database.ts
manuelwedler: Might be worth to also define a type for the `data` parameter there.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+  CompilationLanguage,
+  VerificationExport,
+  VyperSettings,
+} from "@ethereum-sourcify/lib-sourcify";
+import logger from "../../../common/logger";
+import { WStorageService } from "../StorageService";
+import { WStorageIdentifiers } from "./identifiers";
+import { Database } from "../utils/Database";
+import { SourcifyDatabaseService } from "./SourcifyDatabaseService";
+import { sourcifyChainsMap } from "../../../sourcify-chains";
```
manuelwedler: I think it would be better to not import the sourcifyChainsMap here, because you should usually get it via the `chainRepository` of the express app.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+  if (Object.keys(apiUrls).length === 0) {
+    throw new Error(
+      "Routescan chain list did not contain any usable endpoints",
+    );
+  }
+
+  return apiUrls;
+};
+
+export const getEtherscanApiKeyEnvironmentVariables = (): Record<
```
manuelwedler: Rather pass the `sourcifyChainsMap` as parameter. Or even better, move this function to `cli.ts`. It's not really relevant for the rest in this file.

services/server/src/server/cli.ts
```diff
@@ -160,6 +161,28 @@ const server = new Server(
...
+            return acc;
+          },
+          {} as Record<string, string>,
+        ),
+      },
+      BlockscoutVerify: {
+        defaultApiKey: process.env.BLOCKSCOUT_API_KEY as string,
+      },
+      RoutescanVerify: {
+        defaultApiKey: process.env.ROUTESCAN_API_KEY as string,
```
manuelwedler, line 184: The new api keys should be in `.env.dev`

services/server/src/server/cli.ts
```diff
@@ -160,6 +161,28 @@ const server = new Server(
...
+      },
```
manuelwedler, line 185: Could you explain again why we need the api keys for Blockscout and Routescan? Do they have rate limits? I think in the Remix plugin they were not necessary.

services/server/src/server/cli.ts
```diff
@@ -160,6 +161,28 @@ const server = new Server(
         ? parseInt(process.env.ALLIANCE_DB_MAX_CONNECTIONS)
         : undefined,
     },
+    etherscanVerifyAPIServiceOptions: {
+      EtherscanVerify: {
+        defaultApiKey: process.env.ETHERSCAN_API_KEY as string,
```
manuelwedler, line 176: I think we should not use `ETHERSCAN_API_KEY` as default to not overload mainnet calls.
Or maybe it would be better to introduce a new api key for mainnet `ETHERSCAN_API_KEY_MAINNET`.

services/server/src/server/cli.ts
```diff
@@ -160,6 +161,28 @@ const server = new Server(
...
+        apiKeys: Object.entries(
+          getEtherscanApiKeyEnvironmentVariables(),
+        ).reduce(
+          (acc, [chainId, envName]) => {
+            const value = process.env[envName];
+            if (value) acc[chainId] = value;
+            return acc;
+          },
+          {} as Record<string, string>,
+        ),
```
manuelwedler: I think the additional logic here could already be done in the helper function.

services/server/src/server/services/StorageService.ts
```diff
@@ -255,6 +265,48 @@ export class StorageService {
...
+          !options.etherscanVerifyAPIServiceOptions[identifier]
+        ) {
+          logger.error(
+            `${identifier} enabled, but options are not fully set`,
+            options.etherscanVerifyAPIServiceOptions,
+          );
+          throw new Error(
+            `${identifier} enabled, but options are not fully set`,
+          );
+        }
```
manuelwedler: I don't think we should not throw in this case. All options are optional on the `EtherscanVerifyAPIServiceOptions` interface.

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
```diff
@@ -0,0 +1,586 @@
...
+    this.IDENTIFIER = identifier;
+    this.database = sourcifyDatabaseService.database;
+    this.options = {
+      chainApiUrls: options.chainApiUrls || {},
+      apiKeys: options.apiKeys || {},
+      defaultApiKey: options.defaultApiKey || "",
+    };
+  }
+
+  async init(): Promise<boolean> {
```
manuelwedler, line 294: I saw in `StorageService.init` that the init functions are awaited sequentially. Can we make the init calls concurrent with `Promise.all`?

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+  const apiUrls: Record<string, string> = {};
+
+  for (const entry of payload.result) {
+    if (!entry || entry.status !== 1) continue;
+    if (!entry.chainid || !entry.apiurl) continue;
+
+    try {
+      const parsedUrl = new URL(entry.apiurl);
+      parsedUrl.hash = "";
+      apiUrls[entry.chainid] = parsedUrl.toString();
```
manuelwedler: Why is the parsing here needed? Why do we need to set the hash to `""`?

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+    throw new Error(
+      `Failed to load Etherscan chain list (${response.status}): ${body}`,
+    );
+  }
+
+  interface ChainListEntry {
+    chainid?: string;
+    apiurl?: string;
+    status?: number;
+  }
```
manuelwedler: I think it's better to not mix type definitions with logic and rather define interfaces outside function bodies.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+    throw new Error(
+      `Failed to load Blockscout chain list (${response.status}): ${body}`,
+    );
+  }
+  const payload = (await response.json()) as Record<
+    string,
+    {
+      explorers?: Array<{ url?: string; hostedBy?: string }>;
+    }
+  >;
```
manuelwedler: the response type could be defined as an interface to be consistent with the Etherscan function

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+  const apiUrls: Record<string, string> = {};
+  for (const [chainId, chainData] of Object.entries(payload)) {
+    const explorers = chainData?.explorers;
+    if (!Array.isArray(explorers)) continue;
+    const blockscoutExplorer = explorers.find(
+      (explorer) => explorer?.hostedBy === "blockscout" && explorer.url,
```
manuelwedler: why do you limit it to explorers hosted by Blockscout themselves? I think we can also support self-hosted instances.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+    );
+  }
+  return apiUrls;
+};
+
+const fetchRoutescanChainApiUrls = async (): Promise<ChainApiUrls> => {
+  type RoutescanEntry = {
+    chainId?: string | number;
+    evmChainId?: string | number;
+  };
```
manuelwedler: again, I think it's fine to define the type outside this function.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+          entry,
+        });
+        continue;
+      }
+
+      try {
+        const url = new URL("https://api.routescan.io/");
+        url.pathname = `/v2/network/${encodeURIComponent(
+          workspace,
+        )}/evm/${encodeURIComponent(String(chainId))}/etherscan/api`;
```
manuelwedler: This should be defined as a constant on top of the file. We can use a replacement scheme like we do in the contract creation utils.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+        url.search = "";
```
manuelwedler: One thing I noticed is that we don't have the `/api` route on the `ROUTESCAN_API_URL` in contract creation util. Can you double check which is the right one, and if contract creation fetching wasn't working maybe?

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
```diff
@@ -0,0 +1,586 @@
...
+    verification: VerificationExport,
+    jobData?: {
+      verificationId: string;
+      finishTime: Date;
+    },
+  ): Promise<void> {
+    if (
+      verification.compilation.language === "Vyper" &&
+      (this.IDENTIFIER === WStorageIdentifiers.RoutescanVerify ||
+        this.IDENTIFIER === WStorageIdentifiers.BlockscoutVerify)
```
manuelwedler: According to https://docs.blockscout.com/devs/apis/rpc/contract#verify-a-vyper-contract-with-its-source-code-and-contract-creation-information, Vyper verification is possible on Blockscout with another api action. Not sure if we should have it. For the issue with the iframe, it would be good to have at least.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+      chainId: verification.chainId,
+      address: verification.address,
+    };
+
+    const apiBaseUrl = this.getApiBaseUrl(verification.chainId);
+    if (!apiBaseUrl) {
+      logger.warn(
+        "No Etherscan API endpoint configured for chain",
+        submissionContext,
+      );
```
manuelwedler: I think we don't need to log a warning in this case. It is totally expected that it won't run for a lot of chains. Rather change it to debug level.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+      return null;
+    }
+    return constructorArgs.replace(/^0x/i, "");
+  }
+
+  private buildApiUrl(
+    apiBaseUrl: string,
+    action: string,
+    chainId: number,
+    extraParams: Record<string, string> = {},
```
manuelwedler: not necessary to have a parameter that is nowhere used

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+      throw new Error(
+        `Explorer verification request failed (${response.status}): ${responseText}`,
+      );
+    }
+
+    return (await response.json()) as EtherscanRpcResponse;
+  }
+
+  private buildVerificationPayload(verification: VerificationExport): string {
+    const params = new URLSearchParams();
```
manuelwedler: This is very confusing to me. To my knowledge the values are passed as `FormData`. But here we create search parameters and then pass them as the http body. Not sure how this works.

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
```diff
@@ -0,0 +1,586 @@
...
+      },
+      body,
+    });
+
+    if (!response.ok) {
+      const responseText = await response.text();
+      throw new Error(
+        `Explorer verification request failed (${response.status}): ${responseText}`,
+      );
+    }
```
manuelwedler, line 456: We should also store an error message in this case.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+    try {
+      await this.database.upsertExternalVerification(
+        jobData.verificationId,
+        this.IDENTIFIER,
+        {
+          verificationId: response.result,
+        },
+      );
+    } catch (error) {
+      logger.warn("Failed to record external verification receipt", {
```
manuelwedler:
```suggestion
      logger.error("Failed to record external verification receipt", {
```
This is an unexpected case, meaning that something is wrong with the db. Therefore, should be error level.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+        verificationJobId: jobData.verificationId,
+        error: response.result,
+      });
+      await this.database.upsertExternalVerification(
+        jobData.verificationId,
+        this.IDENTIFIER,
+        {
+          error: response.result,
+        },
+      );
```
manuelwedler: This call should also be wrapped in a try catch for consistency. Could also be moved to a private function to reduce code duplication.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+
+    if (target?.path && target?.name) {
+      return `${target.path}:${target.name}`;
+    }
+
+    if (target?.name) {
+      return target.name;
+    }
+
+    return "Contract";
```
manuelwedler:
```suggestion
    const target = verification.compilation.compilationTarget;
    return `${target.path}:${target.name}`;
```

None of the values are optional.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
```diff
@@ -0,0 +1,586 @@
...
+  }
+
+  private getCompilerVersion(verification: VerificationExport): string {
+    const version = verification.compilation.compilerVersion;
+
+    if (!version) {
+      throw new Error("Missing compiler version in verification export");
+    }
```
manuelwedler: This check is not needed.

services/server/test/integration/storageServices/EtherscanVerifyAPIService.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+  type EtherscanVerifyAPIServiceOptions,
+} from "../../../src/server/services/storageServices/EtherscanVerifyAPIService";
+import { WStorageIdentifiers } from "../../../src/server/services/storageServices/identifiers";
+import { MockVerificationExport } from "../../helpers/mocks";
+import type { SourcifyDatabaseService } from "../../../src/server/services/storageServices/SourcifyDatabaseService";
+import type { Database } from "../../../src/server/services/utils/Database";
+
+use(chaiAsPromised);
+
+describe("EtherscanVerifyAPIService", function () {
```
manuelwedler: this test suite doesn't require a running database and should be moved to unit tests folder.

services/server/test/integration/storageServices/EtherscanVerifyAPIService.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+          (init.headers as Record<string, string>)["Content-Type"],
+        ).to.equal("application/x-www-form-urlencoded");
+        expect(init.body).to.be.a("string");
+
+        const body = init.body as string;
+        expect(body).to.include("codeformat=solidity-standard-json-input");
+        expect(body).to.include(
+          `contractaddress=${encodeURIComponent(verification.address)}`,
+        );
+        expect(body).to.include("constructorArguements=");
```
manuelwedler: I think we can assert all body parameters to the exact data that we expect.

services/server/test/integration/storageServices/EtherscanVerifyAPIService.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      result: "receipt-456",
+    };
+    fetchStub.resolves(mockFetchResponse(fetchPayload));
+
+    await expect(service.storeVerification(vyperVerification, jobData)).to
+      .eventually.be.fulfilled;
+
+    const [, requestInit] = fetchStub.firstCall.args;
+    const body = (requestInit as RequestInit).body as string;
+    expect(body).to.include("codeformat=vyper-standard-json-input");
```
manuelwedler: also here, would be nice to check the exact data of all body fields.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
manuelwedler: This counts for all variable names, just be consistent.

services/database/migrations/20251009141621_add_external_verification.sql
marcocastignoli: I agree

services/server/src/server/cli.ts
marcocastignoli: There are still limits on the APIs. I already asked Blockscout and Routescan and I'm waiting for these keys.

services/server/src/server/cli.ts
marcocastignoli: Why one for mainnet specifically?

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
marcocastignoli: No we should not have it, this storage service is just for the EtherscanApiVerifaction imo

services/server/src/server/cli.ts
manuelwedler: Because `ETHERSCAN_API_KEY` is used for mainnet and I don't want to overload it with calls accidentally because it is used as a fallback here.

services/server/src/server/services/storageServices/EtherscanVerifyAPIService.ts
marcocastignoli: From the [official documentation](https://routescan.io/documentation/api/etherscan-like/contracts) I see 

```
https://api.routescan.io/v2/network/mainnet/evm/1/etherscan/api
```

services/server/src/server/cli.ts
manuelwedler: Please don't forget to create an Etherscan account for the new env variable and add the key to the production deployment

marcocastignoli: It's in the todos in the PR description :)

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
manuelwedler: This one seems not solved to me

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
```diff
@@ -0,0 +1,567 @@
...
+      throw error;
+    }
+
+    logger.info("Submitted verification to explorer", {
+      ...submissionContext,
+      receiptId: response.result,
+    });
+
+    if (!jobData?.verificationId) {
+      return;
```
manuelwedler, line 391: I now think we should at least log in this case. This also means that something about the API response is wrong.
Maybe we should also store an error message in the db, such that the user knows that was submitted but the response is malformed.

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
marcocastignoli: now it's fixed

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
manuelwedler: ah wait this is not the response

manuelwedler: ignore my comment haha

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
```diff
@@ -0,0 +1,570 @@
...
+      logger.debug(
+        "No Etherscan API endpoint configured for chain",
+        submissionContext,
+      );
+      return;
+    }
+
+    if (!jobData?.verificationId) {
+      return;
+    }
```
manuelwedler: This was moved in the latest commit. This will mean that any call to API v1 will not be submitted. I think we can still submit and just don't store anything

services/server/test/unit/storage/EtherscanVerifyApiService.spec.ts
```diff
@@ -38,7 +38,7 @@ const buildExpectedStandardJsonInput = (
   });
 };
 
-describe("EtherscanVerifyApiService", function () {
+describe.only("EtherscanVerifyApiService", function () {
```
manuelwedler: Please remove `.only`

PR #2443, opened 2025-10-14
Update 4byte after bulk import and implementing `/import` endpoint

kuzdogan:
- Implement `/import` endpoint according to openchain
  - Use ether.js `Fragment.from()` to parse and validate functions. This is normally using [Human-readable ABI](https://docs.ethers.org/v5/api/utils/abi/formats/#abi-formats--human-readable-abi) and hence needs a `function ` keyword prefix. It works reliably as can be seen from the tests. This way we avoid porting the complete openchain [abi.go](https://github.com/openchainxyz/openchain-monorepo/blob/a348474d05524982e2d370755d3be0b24a9c71d2/internal/solidity/abi.go#L396) parsing module and outsource to ethers.
  - Added a bulk test for long and complex signatures taken from the existing DB.
- Changed `signature_stats` materialized view query to have `unknown` and `total` fields.
- Remove `error` from response types. Add a `hasVerifiedContract` field to indicate the signature is associated with a verified contract and not just `/import`ed.
- Fixes the VS Code debug `tasks.json` and `launch.json`
- Remove `process.exit(0)` from the `shutdown()` function and only exit on signal. Otherwise Mocha kills the process without printing which tests fail why.

TODO:
- [x] Update the materialized view definition on staging
- [x] Update the materialized view definition on production (create card/issue)
- [x] Use RW db users instead of the read-only on staging and prod. Give limited access

### DIFF
--- a/services/4byte/src/api/validation.ts
+++ b/services/4byte/src/api/validation.ts
@@ -30,22 +30,16 @@ export function validateHashQueries(
     {
       function?: string;
       event?: string;
-      error?: string;
     }
   >,
   res: Response,
   next: NextFunction,
 ): void {
-  const {
-    function: functionQuery,
-    event: eventQuery,
-    error: errorQuery,
-  } = req.query;
+  const { function: functionQuery, event: eventQuery } = req.query;
 
   const hashes = [
     ...(functionQuery?.split(",") || []),
     ...(eventQuery?.split(",") || []),
-    ...(errorQuery?.split(",") || []),
   ];
 
   for (const hash of hashes) {


services/4byte/src/api/handlers.ts
```diff
@@ -106,38 +130,29 @@ export function createSignatureHandlers(
       try {
         const functionHashes = sanitizeHashes(req.query.function);
         const eventHashes = sanitizeHashes(req.query.event);
-        const errorHashes = sanitizeHashes(req.query.error);
         const shouldFilter = sanitizeFilter(req.query.filter);
 
-        const result: SignatureResult = { function: {}, event: {}, error: {} };
+        const result: SignatureResult = { function: {}, event: {} };
```
marcocastignoli, line 135: Here we removed error, but it's still in `services/4byte/src/api/validation.ts`

kuzdogan: Good catch

kuzdogan:
@marcocastignoli removed the error field

PR #2447, opened 2025-10-16
Fix openchain compatiblity

kuzdogan:
I realized the openchain API has an unusual behavior where for not found function hashes and event hashes it returns `null` vs. `[]`

https://api.openchain.xyz/signature-database/v1/lookup?function=0xa9059cbb,0xffffffff,0xaa&event=0xaa

```json
{
  "ok": true,
  "result": {
    "event": {
      "0xaa": []
    },
    "function": {
      "0xa9059cbb": [
        {
          "name": "transfer(address,uint256)",
          "filtered": false
        }
      ],
      "0xaa": null,
      "0xffffffff": null
    }
  }
}
```

The changes make it fully compatible


marcocastignoli (approved):
lgtm!

PR #2451, opened 2025-10-16
Remove the signature database tag from server openapi.yaml

kuzdogan:


manuelwedler (approved):
The added endpoints are also documented here: https://docs.sourcify.dev/docs/api/#other (on the bottom of the page)
We can remove this section.

kuzdogan:
Yes I'll do it after a release

Issue #2374, opened 2025-09-16
Match bytecodes when `bytecodeHash: none` for same contracts, even if lengths are different


manuelwedler:
A user in Matrix reported the following issue to us: They wanted to verify a contract that was deployed with setting the metadata hash disabled:

```json
    "metadata": {
        "bytecodeHash": "none"
    },
```

You can see that this setting was used to verify the contract on Blockscout: https://scan.9mm.pro/address/0xDA9aBA4eACF54E0273f56dfFee6B8F1e20B23Bba?tab=contract

And you can see that the onchain runtime cbor auxdata is missing the metadata hash:

```json
{
  "solc": "0x000706"
}
```

However, the user seems to have tried to verify the contract without this compiler setting, meaning that the recompilation in Sourcify will have a metadata hash in the cbor auxdata. I also tried this and got a failed compilation with error code `bytecode_length_mismatch`: 

https://verify.sourcify.dev/jobs/331fb426-d485-4cf2-8bec-b746a08f33d3

When you include this setting for verification, you get a partial match as expected:

https://verify.sourcify.dev/jobs/788b9ffd-6aaf-4ebc-8829-c7e91876027a

I think this shows a case where we shouldn't expect the bytecodes to have the same length. Sourcify should completely ignore the CBOR auxdata in this case and also get a partial match if the bytecode hash is appended only in the recompilation.

kuzdogan:
I can't find it but I recall at the time we considered this to be rather rare case and it's ok to break early if the bytecode lengths are different. 

We had multiple similar cases like this and it seems it's not as rare as we've thought. I also think we should consider verifying contracts with different length. Maybe we can throw early if there is a substantial length difference between two contracts. AFAIK how Blockscout handles this is, they mark the "executional" bytecode and metadata and they ignore the metadata fully. We should follow a similar approach. If possible we should also think about other places where we assume bytecode lengths are equal.

manuelwedler:
Yes, probably we should also only compare the "executional" bytecode lengths for throwing the length mismatch error. It should at least fix this issue.

kuzdogan:
We can come up with couple test cases for this and solve it for each:
- `bytecodeHash: none` vs. standard
- `appendCbor: false` vs. standard
- `appendCbor: false` vs. `bytecodeHash: none`
- when there are nested cbors (factory)

kuzdogan:
Related issue: https://github.com/argotorg/sourcify/issues/2315 

[...]

PR #2456, opened 2025-10-23
Support contracts without auxdata

marcocastignoli:
See #2374

### DIFF
--- a/packages/lib-sourcify/test/utils.ts
+++ b/packages/lib-sourcify/test/utils.ts
@@ -2,12 +2,13 @@
 import path from 'path';
 import { expect } from 'chai';
 import type { Signer } from 'ethers';
-import { ContractFactory, JsonRpcSigner } from 'ethers';
+import { ContractFactory, type JsonRpcSigner } from 'ethers';
 import {
   useSolidityCompiler,
   useVyperCompiler,
 } from '@ethereum-sourcify/compilers';
 import type {
+  Metadata,
   SolidityJsonInput,
   SolidityOutput,
   SolidityOutputContract,
@@ -22,8 +23,8 @@ import type {
 } from '../src/Compilation/CompilationTypes';
 import fs from 'fs';
 import {
-  PathContent,
-  SolidityCompilation,
+  type PathContent,
+  type SolidityCompilation,
   SolidityMetadataContract,
   VyperCompilation,
 } from '../src';


packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -131,9 +132,33 @@ export class Verification {
...
+          );
+          if (
+            !decodedOnchainAuxdata.ipfs &&
+            !decodedOnchainAuxdata.bzzr0 &&
+            !decodedOnchainAuxdata.bzzr1
+          ) {
+            noOnchainMetadata = true;
+          }
+        } catch (err: any) {
+          noOnchainMetadata = true;
```
manuelwedler: for my understanding,  `decode` throws if there is no auxdata, or how does it work?

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -131,9 +132,33 @@ export class Verification {
...
+        }
+      }
+
+      // we throw the bytecode length mismatch error only if there is onchain metadata
+      if (!noOnchainMetadata) {
```
manuelwedler: For Vyper, there will never be auxdata in the onchain runtime code. This means the logic is changed for Vyper. Is this intentional?

packages/lib-sourcify/src/Verification/Verification.ts
manuelwedler, line 127: Can we handle the Vyper case in a separate if-else clause? I think the repeated Solidity checks are confusing.

How about just adding this snippet outside the rest of this block?

```ts
    if (this.compilation.language === 'Vyper' &&
        compiledRuntimeBytecode.length > this.onchainRuntimeBytecode.length) {
        throw new VerificationError({
          code: 'bytecode_length_mismatch',
        });
    }
```

Then only check `(this.compilation.language === 'Solidity')` once in the separate Solidity block.

packages/lib-sourcify/src/Verification/Verification.ts
manuelwedler: I see the new comments make it clear :)

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -323,10 +323,20 @@ export function extractAuxdataTransformation(
...
-        onchainAuxdata +
-        populatedRecompiledBytecode.slice(offsetEnd);
+      if (
+        onchainAuxdata.length === 0
+        // TODO with this we could potentially support multiple auxdata sections, but needs more testing
+        // && (true || index === Object.values(cborAuxdataPositions).length - 1)
+      ) {
+        populatedRecompiledBytecode =
+          populatedRecompiledBytecode.slice(0, offsetStart - 2) +
+          populatedRecompiledBytecode.slice(offsetEnd);
```
manuelwedler, line 336: I don't fully understand this addition. What case does this solve?

packages/lib-sourcify/test/Verification/Verification.spec.ts
```diff
@@ -84,6 +120,46 @@ async function getCompilationFromMetadata(contractFolderPath: string) {
...
+) {
+  const compilation = await getCompilationFromMetadata(
+    contractFolderPath,
+    metadataMutator,
+  );
+  await compilation.compile();
+  return compilation;
+}
+
+async function deployCompiledContract(
```
manuelwedler: can we move helper functions to `test/util.ts`?

packages/lib-sourcify/src/Verification/Transformations.ts
marcocastignoli: I added a comment to explain

manuelwedler: thanks!

manuelwedler: didn't know about this additonal `ff` at the end

services/database/database-specs
manuelwedler, line 1: I don't know how this happened but the submodule should not have been touched here.

marcocastignoli: Probably I did something wrong while merging

marcocastignoli: I fixed it

PR #2458, opened 2025-10-24
feat: Add SSL support for SOURCIFY_POSTGRES

yorik:
* Add `SOURCIFY_POSTGRES_SSL` to enable/disable SSL. 
* Add `SOURCIFY_POSTGRES_SSL_REJECT_UNAUTHORIZED` to control server certificate verification.
* Removed conflicting with prettier eslint configuration.


yorik:
Not sure why, but if I run `prettier "./**/*.ts" --fix` then eslint unhappy, if I run `eslint . --ext .ts --fix` then prettier unhappy.

PR #2462, opened 2025-10-27
feat: add Reactive Network Mainnet

daniil-parsiq:
## Checklist

- [✓] The branch is named as `add-chain-<chainId>`.
- [✓] I haven't modified the `chains.json` file directly.
- [✓] In `sourcify-chains.json` file
  - [✓] I've set `supported: true`.
  - [✓] I haven't added an `rpc` field but the one in [chains.json](../../src/chains.json) is used (if not, please explain why).
- [✓] I've added a test in [chain-tests.js](../../test/chains/chains-test.js) file.
- [✓] `test-new-chain` test in Circle CI is passing.

### DIFF
--- a/services/server/src/sourcify-chains-default.json
+++ b/services/server/src/sourcify-chains-default.json
@@ -2397,5 +2397,15 @@
     "fetchContractCreationTxUsing": {
       "etherscanApi": true
     }
+  },
+  "1597": {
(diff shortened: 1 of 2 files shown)


kuzdogan:
The PR looks good but our CI is doing weird things and not detecting the chainid somehow. We need to double check what's going on there

PR #2464, opened 2025-10-27
Add compiler version to unique constraint of `compiled_contracts`

### LINKED ISSUE #2264: compiler version is not part of the unique constraint of `compiled_contracts`
The unique constraint of the `compiled_contracts` table looks like this at the moment:
```sql
CONSTRAINT compiled_contracts_pseudo_pkey UNIQUE (compiler, language, creation_code_hash, runtime_code_hash),
```

I think it was created under the misconception that `compiler` would also include the compiler version, but actually `compiler` is either "solc" or "vyper" and the compiler version is stored in the `version` column.

The issue with this is that if two compiler versions produce the same bytecodes (you would need to turn off metadata), it would not be possible to store a compilation for both versions.

We should modify the constraint to also include the `version` column.

manuelwedler:
Closes #2264

Updates the git submodule to pull the new migration from VerA database-specs

Also updates sourcify_database.sql for the changes of #2430, because it wasn't done there.

Removes some lines from the base sourcify migration, because they weren't reflecting the live databases.

### DIFF
--- a/services/server/src/server/services/storageServices/AbstractDatabaseService.ts
+++ b/services/server/src/server/services/storageServices/AbstractDatabaseService.ts
@@ -119,7 +119,7 @@ export default abstract class AbstractDatabaseService {
       return verifiedContractInsertResult.rows[0].id;
     } catch (e) {
       throw new Error(
-        `cannot insert verified_contract address=${databaseColumns.contractDeployment.address.toString()} chainId=${databaseColumns.contractDeployment.chain_id}\n${e}`,
+        `cannot insert verified_contract address=0x${databaseColumns.contractDeployment.address.toString("hex")} chainId=${databaseColumns.contractDeployment.chain_id}\n${e}`,
       );
     }
   }
@@ -214,7 +214,7 @@ export default abstract class AbstractDatabaseService {
       return verifiedContractInsertResult.rows[0].id;
     } catch (e) {
       throw new Error(
-        `cannot update verified_contract address=${databaseColumns.contractDeployment.address.toString()} chainId=${databaseColumns.contractDeployment.chain_id}\n${e}`,
+        `cannot update verified_contract address=0x${databaseColumns.contractDeployment.address.toString("hex")} chainId=${databaseColumns.contractDeployment.chain_id}\n${e}`,
       );
     }
   }


services/database/sourcify-database.sql
```diff
@@ -928,8 +914,8 @@ CREATE TABLE public.compiled_contracts (
     fully_qualified_name character varying NOT NULL,
     compiler_settings jsonb NOT NULL,
     compilation_artifacts jsonb NOT NULL,
-    creation_code_hash bytea,
-    creation_code_artifacts jsonb,
```
kuzdogan, line 932: Did you double check in DBs if there were any NULL values for these?

services/database/sourcify-database.sql
```diff
@@ -9,20 +9,6 @@ SET xmloption = content;
...
-CREATE EXTENSION IF NOT EXISTS pg_cron WITH SCHEMA pg_catalog;
-
-
---
--- Name: EXTENSION pg_cron; Type: COMMENT; Schema: -; Owner: -
---
-
-COMMENT ON EXTENSION pg_cron IS 'Job scheduler for PostgreSQL';
-
-
```
kuzdogan, line 25: Why is it removing pg_cron?

services/database/sourcify-database.sql
manuelwedler: the columns  are already set up as `NOT NULL` in the live dbs. There is really none.

services/database/sourcify-database.sql
manuelwedler: oooh didn't see that. Probably due to the fallback mechanism?

manuelwedler:
The tests were failing because the ON CONFLICT clause in the insertCompiledContract function wasn't containing the updated unique constraint. This means the logic to insert works only with the updated schema.

This also means, when we deploy this to production, we need to upgrade our db at the exactly same time as the deployment. Verifications using the old logic will fail after upgrading the db. So there will be a short time window where some verifications might fail. Any ideas how to solve this?

manuelwedler:
Due to this I rolled the migration on VerA back... storing verifications from our side to vera failed probably in the last hour...

services/database/sourcify-database.sql
kuzdogan: Ah yeah

services/database/sourcify-database.sql
kuzdogan: Hmm yeah probably it doesn't see pg_crypto so it does not execute these lines. 

I think they should stay? Not a huge deal though

kuzdogan:
The only thing I think of is to time the deployment with the migration. We can do this and test on staging when we're merging this PR. This PR should also contain the changes needed in the `insertCompiledContract` code

services/database/sourcify-database.sql
manuelwedler: yes they should

manuelwedler:
I found a solution for migrating: It is possible to just specify the constraint name in the ON CONFLICT clause. This makes the logic compatible with both versions.

services/server/src/server/services/storageServices/AbstractDatabaseService.ts
```diff
@@ -119,7 +119,7 @@ export default abstract class AbstractDatabaseService {
       return verifiedContractInsertResult.rows[0].id;
     } catch (e) {
       throw new Error(
-        `cannot insert verified_contract address=${databaseColumns.contractDeployment.address} chainId=${databaseColumns.contractDeployment.chain_id}\n${e}`,
+        `cannot insert verified_contract address=${databaseColumns.contractDeployment.address.toString()} chainId=${databaseColumns.contractDeployment.chain_id}\n${e}`,
```
kuzdogan: Curious, what was the error there. I thought the string conversion would be implicit?

manuelwedler: Actually, you are right here. just calling toString doesn't change anything. I needed to add `toString('hex')`. Fixed it in the latest commit.

PR #2471, opened 2025-10-28
Add canonical signature

kuzdogan:
Came across a signature with a collusion: https://4byte.sourcify.dev/?q=0x09b81346

### DIFF
--- a/services/4byte/src/utils/canonical-signatures.json
+++ b/services/4byte/src/utils/canonical-signatures.json
@@ -133,5 +133,9 @@
   "0x8abf6077": {
     "signature": "proxyOwner()",
     "source": "generic:proxy"
+  },
+  "0x09b81346": {
+    "signature": "exactOutput((bytes,address,uint256,uint256))",
+    "source": "manual"
   }
 }


manuelwedler:
I don't know if we should add any random signature collision to the canonical json file. It would make more sense to me if we know this signature comes from a widely used contract.

~~Also, this signature looks weird to me. It uses a tuple as parameter. I wonder how the Solidity Source to this looks like. I thought tuples would only be a return type.~~
Nevermind. Struct types are transformed to tuple in the ABI.

kuzdogan:
I checked the number of `compiled_contracts` that use the canonical vs. the spam and it's 177 vs. 1. To me this is an intentional collusion. Don't you say so?

manuelwedler:
Okay, I do agree with that. My concern was rather about the process and if we really want to open a PR for any such case we come randomly across. I expect that there are tons of such cases. Maybe it makes sense to just search in the db for all of these cases and create a bundled PR?

kuzdogan:
Eventually yes but not a prio yet to me

PR #2473, opened 2025-10-29
Only count timeouts and network errors as rpc failures

manuelwedler:
We noticed that some RPCs get blocked on the monitor because they are returning an error when querying an unpublished block. The usual behavior for querying unpublished block is returning `null`. Since these rpcs behave unexpectedly the errors are counted as rcp failures in the rpc health state. This PR changes the logic to only count timeouts and network errors as rpc failures. Other errors are returned to the caller.

### DIFF
--- a/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
+++ b/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
@@ -222,8 +222,8 @@ export class SourcifyChain {
           maskedUrl: rpc.maskedUrl,
           chainId: this.chainId,
         });
-        // Don't mark as unhealthy. This is an error that should be handled by the caller.
-        throw error;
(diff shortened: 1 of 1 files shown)


packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
```diff
@@ -212,16 +202,28 @@ export class SourcifyChain {
...
         logError('RPC operation threw error', {
           operation: operationName,
-          error: err,
+          error,
           maskedUrl: rpc.maskedUrl,
           chainId: this.chainId,
         });
-        throw err;
+        // Don't mark as unhealthy. This is an error that should be handled by the caller.
+        throw error;
```
marcocastignoli: We shouldn't throw here otherwise we'll break the loop and not try the next RPC

marcocastignoli (approved):
lgtm!

PR #2477, opened 2025-10-30
Disable Holesky Testnet because of deprecation

manuelwedler:


services/server/test/chains/chain-tests.spec.ts
```diff
@@ -159,12 +159,13 @@ describe("Test Supported Chains", function () {
...
-    "Holesky",
-    "shared/",
-  );
+  // Ethereum Holesky Testnet
+  // verifyContract(
+  //   "0x7ecedB5ca848e695ee8aB33cce9Ad1E1fe7865F8",
+  //   "17000",
+  //   "Holesky",
+  //   "shared/",
+  // );
```
kuzdogan, line 168: Just FYI we don't need to do this, the tests already check if the chain is supported

kuzdogan (approved):

services/server/test/chains/chain-tests.spec.ts
manuelwedler: ah right

PR #2481, opened 2025-10-30
Similarity search verification

marcocastignoli:
This PR implements a new API endpoint POST `/v2/verify/similarity/{chainId}/{address}` with no body specified. 

This new endpoint starts a verification job that searches the Sourcify database for contracts whose runtime bytecode is similar to the contract deployed at the given address (checking the first 75 bytes). The job will attempt to verify against each candidate until a match is found or the candidate list is exhausted.


services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -90,6 +95,12 @@ export async function verifyFromEtherscan(
   return runWorkerFunctionWithContext(_verifyFromEtherscan, input);
 }
 
+export async function verifySimilarity(
```
manuelwedler, line 97: Can we have unit tests for this function in `verificationWorker.spec.ts`?

services/server/test/integration/apiv2/verification/verify.similarity.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      .post(
+        `/v2/verify/similarity/${chainFixture.chainId}/${chainFixture.defaultContractAddress}`,
+      )
+      .send();
+
+    chai.expect(verifyRes.status).to.equal(202);
+    chai.expect(verifyRes.body).to.have.property("verificationId");
+
+    // We need to wait because `getSimilarityCandidatesByRuntimeCode` is delaying the job push to the queue
+    await waitSecs(1);
```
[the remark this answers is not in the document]
marcocastignoli: Read the comment above to understand why I used it

services/server/src/server/services/utils/Database.ts
```diff
@@ -250,6 +251,27 @@ ${
...
+      `
+        SELECT
+          compiled_contracts.id as compilation_id,
+          contract_deployments.chain_id,
+          concat('0x', encode(contract_deployments.address, 'hex')) as address
+        FROM ${this.schema}.code code
+        JOIN ${this.schema}.compiled_contracts ON compiled_contracts.runtime_code_hash = code.code_hash
+        JOIN ${this.schema}.verified_contracts ON verified_contracts.compilation_id = compiled_contracts.id
+        JOIN ${this.schema}.contract_deployments ON verified_contracts.deployment_id = contract_deployments.id
+        WHERE substring(code.code FROM 1 FOR 75) = substring($1::bytea FROM 1 FOR 75)
```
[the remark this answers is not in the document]
marcocastignoli: Makes sense maybe?

services/server/src/server/services/utils/database-util.ts
```diff
@@ -243,6 +243,12 @@ export type GetSourcifyMatchesByChainResult = Pick<
   "id" | "creation_match" | "runtime_match"
 > & { address: string; verified_at: string };
 
+export interface CodePrefixMatchResult {
+  compilation_id: Tables.VerifiedContract["compilation_id"];
+  chain_id: number;
```
manuelwedler: `chain_id` is returned as a string. Better to use `Pick<Tables.ContractDeployment, "chain_id">`

services/database/migrations/20251101120000_add_code_first_75_index.sql
```diff
@@ -0,0 +1,9 @@
+-- migrate:up
```
manuelwedler, line 1: You have to update sourcify-database.sql too when you add migrations

services/server/src/server/services/utils/Database.ts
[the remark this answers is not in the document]
manuelwedler: not too important IMO (copilot suggestion)

services/server/src/server/services/utils/Database.ts
```diff
@@ -250,6 +251,27 @@ ${
...
+    limit: number = 20,
+  ): Promise<QueryResult<CodePrefixMatchResult>> {
+    return await this.pool.query(
+      `
+        SELECT
+          compiled_contracts.id as compilation_id,
+          contract_deployments.chain_id,
+          concat('0x', encode(contract_deployments.address, 'hex')) as address
+        FROM ${this.schema}.code code
+        JOIN ${this.schema}.compiled_contracts ON compiled_contracts.runtime_code_hash = code.code_hash
```
manuelwedler, line 265: Not sure if it would make a difference practically, but shouldn't we compare against the deployed runtime code since we also use the deployed bytecode from the contract in question?

services/server/src/server/services/VerificationService.ts
```diff
@@ -345,13 +352,96 @@ export class VerificationService {
...
+              customCode: "no_similar_match_found",
+              errorId: uuidv4(),
+            },
+          ]);
+          return;
+        }
+
+        const creatorTxHash =
+          (await getCreatorTx(this.sourcifyChainMap[chainId], address)) ||
+          undefined;
```
manuelwedler: For consistency, `getCreatorTx` should be called inside the worker

services/database/migrations/20251101120000_add_code_first_75_index.sql
```diff
@@ -0,0 +1,9 @@
...
+
+CREATE INDEX IF NOT EXISTS idx_code_code_first_75
```
manuelwedler, line 3: In case you already manually added the index to our dbs, please make sure to also add a row in the `schema_migrations` table

services/server/src/server/apiv2/verification/verification.handlers.ts
```diff
@@ -146,3 +146,31 @@ export async function verifyFromEtherscanEndpoint(
 
   res.status(StatusCodes.ACCEPTED).json({ verificationId });
 }
+
+interface VerifySimilarityRequest extends Request {
+  params: {
+    chainId: string;
+    address: string;
+  };
```
manuelwedler, line 154: I wonder if we should also support providing the `creationTransactionHash`. I think it makes sense if this API is called by the monitor or explorers.

services/server/src/apiv2.yaml
```diff
@@ -1,6 +1,6 @@
 openapi: 3.1.0
 paths:
-  '/v2/verify/{chainId}/{address}':
+  "/v2/verify/{chainId}/{address}":
```
manuelwedler, line 3: Just a general feedback. These formatting changes make the review hard. If you think they are required, put them in a separate commit.

services/server/src/server/apiv2/verification/verification.routes.ts
```diff
@@ -49,4 +50,14 @@ router
     verifyFromEtherscanEndpoint,
   );
 
+router
+  .route("/verify/similarity/:chainId/:address")
+  .post(
+    validateChainId,
+    validateAddress,
+    checkIfAlreadyVerified,
+    checkIfJobIsAlreadyRunning,
```
manuelwedler, line 59: can we have tests that verify that these validators run? We also have these tests on each of the other verification endpoints.

See `should return a 429 if the contract is being verified at the moment alread` and all following tests inside `verify.metadata.spec.ts`.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -1053,4 +1054,59 @@ export class SourcifyDatabaseService
...
+
+  async getSimilarityCandidatesByRuntimeCode(
+    runtimeBytecode: string,
+    limit: number,
+  ): Promise<GetSourcifyMatchByChainAddressWithPropertiesResult[]> {
+    await this.init();
+
+    const runtimeBuffer = bytesFromString(runtimeBytecode);
+    if (!runtimeBuffer || runtimeBuffer.length === 0) {
+      return [];
```
manuelwedler: I think we should throw here. If this case is encountered, we have an unexpected server error.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -1053,4 +1054,59 @@ export class SourcifyDatabaseService
...
+
+    const results: GetSourcifyMatchByChainAddressWithPropertiesResult[] = [];
+
+    for (const row of prefixMatches.rows) {
+      const addressBuffer = bytesFromString(row.address);
+      if (!addressBuffer) {
+        continue;
+      }
+      const matchResult =
+        await this.database.getSourcifyMatchByChainAddressWithProperties(
```
manuelwedler, line 1151: Is it possible to run this concurrently with `Promise.all`?

services/server/src/server/services/utils/Database.ts
```diff
@@ -250,6 +251,27 @@ ${
...
+        JOIN ${this.schema}.verified_contracts ON verified_contracts.compilation_id = compiled_contracts.id
```
manuelwedler, line 266: since you are not joining on sourcify_matches, we might get the same chain_id and address pair multiple times. We later use them to get the data from sourcify_matches, so some candidates will be duplicated.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -1053,4 +1054,59 @@ export class SourcifyDatabaseService
...
+            "fully_qualified_name",
+            "version",
+            "creation_cbor_auxdata",
+            "runtime_cbor_auxdata",
+            "metadata",
+          ],
+        );
+
+      if (matchResult.rows.length === 0) {
+        continue;
```
manuelwedler: We can continue in this case, but I think we should log a warning.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -1053,4 +1054,59 @@ export class SourcifyDatabaseService
...
+      if (!addressBuffer) {
+        continue;
+      }
+      const matchResult =
+        await this.database.getSourcifyMatchByChainAddressWithProperties(
+          row.chain_id,
+          addressBuffer,
+          [
+            "chain_id",
+            "address",
```
manuelwedler: You already have these values. Let's rather reuse them and not fetch them another time.

services/server/src/server/services/workers/workerTypes.ts
```diff
@@ -36,6 +39,14 @@ export interface VerifyFromEtherscanInput extends VerificationWorkerInput {
   etherscanResult: EtherscanResult;
 }
 
+export interface VerifySimilarityInput extends VerificationWorkerInput {
+  chainId: string;
+  address: string;
+  runtimeBytecode: string;
+  creatorTxHash?: string;
+  candidates: GetSourcifyMatchByChainAddressWithPropertiesResult[];
```
manuelwedler: Can we use a more specific interface here? The properties we need are all optional on this interface. I think it would improve the code if the properties we expect from a candidate are required before passing to the worker.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -1053,4 +1054,59 @@ export class SourcifyDatabaseService
       );
     });
   }
+
+  async getSimilarityCandidatesByRuntimeCode(
+    runtimeBytecode: string,
+    limit: number,
+  ): Promise<GetSourcifyMatchByChainAddressWithPropertiesResult[]> {
```
manuelwedler: Regarding the comment before, I would already return the more specific interface here.

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -29,13 +31,16 @@ import type {
 import logger, { setLogLevel } from "../../../common/logger";
 import { getCompilationFromEtherscanResult } from "../utils/etherscan-util";
 import { asyncLocalStorage } from "../../../common/async-context";
+import SourcifyChainMock from "../utils/SourcifyChainMock";
+import { getAddress } from "ethers";
+import { VerifySimilarityInput } from "./workerTypes";
```
manuelwedler: for consistency, move `VerifySimilarityInput` to the `import type` clause a few lines above.

Just for you to be aware of this: I think it is good to keep the actual imports on a minimum in the worker because workers are frequently started and more imports increase the start up time.

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -29,13 +31,16 @@ import type {
 import logger, { setLogLevel } from "../../../common/logger";
 import { getCompilationFromEtherscanResult } from "../utils/etherscan-util";
 import { asyncLocalStorage } from "../../../common/async-context";
+import SourcifyChainMock from "../utils/SourcifyChainMock";
```
manuelwedler, line 34: SourcifyChainMock will import Database which will import a bunch of dependencies which is not necessary in the worker IMO. This can simply be fixed by making the Database import a type import inside SourcifyChainMock.

manuelwedler: I think we should actually enforce this by enabling the `consistent-type-imports` rule of eslint. Maybe in a separate PR to not pollute the changes here.

services/server/src/server/services/VerificationService.ts
```diff
@@ -345,13 +352,96 @@ export class VerificationService {
...
+      "storeVerificationJob",
+      [new Date(), chainId, address, verificationEndpoint],
+    );
+
+    const sourcifyDatabaseService = this.storageService.rwServices[
+      "SourcifyDatabase"
+    ] as SourcifyDatabaseService;
+
+    sourcifyDatabaseService
+      .getSimilarityCandidatesByRuntimeCode(
```
manuelwedler: I think it would be more consistent if this function is defined on the `RWStorageService` interface as an optional method. 

The issue currently is that `SourcifyDatabase` can be undefined if it is disabled. Calling via `performServiceOperation` will throw the appropriate error message for this case at least.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -1053,4 +1054,59 @@ export class SourcifyDatabaseService
...
+            "address",
+            "std_json_input",
+            "std_json_output",
+            "fully_qualified_name",
+            "version",
+            "creation_cbor_auxdata",
+            "runtime_cbor_auxdata",
+            "metadata",
+          ],
+        );
```
manuelwedler: I wonder if we should split the ~~`prepareCompilerSettingsFromVerification`~~ `extractCompilationFromDatabase` utility function and reuse it here. If we make it two functions, we can use one part here, and the other part inside the worker. In the worker, the second part of the function is basically duplicated in `createPreRunCompilationFromCandidate`.

btw, I wonder why `extractCompilationFromDatabase` is not a method of `SourcifyDatabase` since it requires a database instance. So you could also improve this aspect if you agree.

services/server/src/server/services/VerificationService.ts
```diff
@@ -345,13 +352,96 @@ export class VerificationService {
...
+        runtimeBytecode,
+        DEFAULT_SIMILARITY_CANDIDATE_LIMIT,
+      )
+      .then(async (candidates) => {
```
manuelwedler: Don't use `async` inside a `then` or `catch` call. It's mixing the two styles of using promises. I think it can be dangerous since it is hard to understand what is happening. In your case, if `getCreatorTx` throws, it won't be handled here because of this.

services/server/src/server/services/VerificationService.ts
```diff
@@ -345,13 +352,96 @@ export class VerificationService {
...
+          address,
+          runtimeBytecode,
+          creatorTxHash,
+          candidates,
+          traceId: asyncLocalStorage.getStore()?.traceId,
+        };
+
+        this.verifyViaWorker(verificationId, "verifySimilarity", input);
+      })
+      .catch(async (error) => {
```
manuelwedler: same here. Just return the promise instead of awaiting it such that it is properly added to the promise chain.

services/server/src/server/services/VerificationService.ts
```diff
@@ -345,13 +352,96 @@ export class VerificationService {
...
+        });
+        await this.storageService.performServiceOperation("setJobError", [
+          verificationId,
+          new Date(),
+          {
+            customCode: "internal_error",
+            errorId: uuidv4(),
+          },
+        ]);
+      });
```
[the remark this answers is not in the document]
manuelwedler: This comment is of course not relevant. We intended this behavior.

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -256,6 +267,219 @@ async function _verifyFromEtherscan({
...
+  const mockChain = new SourcifyChainMock(
+    {
+      onchain_runtime_code: runtimeBytecode,
+      onchain_creation_code: creationData.creationBytecode,
+      deployer: creationData.deployer,
+      block_number: creationData.blockNumber,
+      transaction_index: creationData.txIndex,
+      transaction_hash: creatorTxHash,
+      address: checksumAddress,
+    },
```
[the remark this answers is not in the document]
manuelwedler: I agree with Copilot here. Ideally, we could have a better type for the parameter of the `SourcifyChainMock` constructor.

services/server/test/integration/apiv2/verification/verify.similarity.spec.ts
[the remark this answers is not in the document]
manuelwedler: I agree with Copilot that waiting with a fixed delay is not great. I think the promise from `getSimilarityCandidatesByRuntimeCode` should be added to the `runningTasks` inside the `VerificationService` and via that property it can be awaited here.

services/server/src/server/services/VerificationService.ts
```diff
@@ -345,13 +352,96 @@ export class VerificationService {
...
+    verificationEndpoint: string,
+    chainId: string,
+    address: string,
+  ): Promise<VerificationJobId> {
+    let runtimeBytecode: string;
+    try {
+      runtimeBytecode =
+        await this.sourcifyChainMap[chainId].getBytecode(address);
+    } catch (error) {
+      throw new VerificationError({
```
manuelwedler: can we have a test for this case?

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -256,6 +267,219 @@ async function _verifyFromEtherscan({
...
+      chainId,
+      address,
+    });
+    return {
+      errorExport: {
+        customCode: "internal_error",
+        errorId: uuidv4(),
+      },
+    };
+  }
```
manuelwedler: In the other functions we just expect the `sourcifyChain` to be present. If you want to check for it, maybe do it in all functions of the worker. Also, I think we should just throw and not return an error export. The verification service will handle this case and already set it to `internal_error`.

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -256,6 +267,219 @@ async function _verifyFromEtherscan({
...
+
+  if (runtimeBytecode === "0x") {
+    return {
+      errorExport: {
+        customCode: "contract_not_deployed",
+        errorId: uuidv4(),
+      },
+    };
+  }
```
manuelwedler: These are also for an unexpected state. We should just throw in these cases. The `VerificationService` should have thrown the custom error codes earlier, such that it is not necessary to do it here.

services/server/src/server/services/VerificationService.ts
```diff
@@ -345,13 +352,96 @@ export class VerificationService {
...
   }
 
+  public async verifyFromSimilarityViaWorker(
+    verificationEndpoint: string,
+    chainId: string,
+    address: string,
+  ): Promise<VerificationJobId> {
+    let runtimeBytecode: string;
+    try {
+      runtimeBytecode =
```
manuelwedler, line 367: We should add a check for `runtimeBytecode === "0x"`

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -256,6 +267,219 @@ async function _verifyFromEtherscan({
...
+
+  // Fetch creation data to be used in the SourcifyChainMock
+  let creationData: {
+    creationBytecode?: string;
+    deployer?: string;
+    blockNumber?: number;
+    txIndex?: number;
+  } = {};
+  if (creatorTxHash) {
```
manuelwedler: I would also like to see a test if `creatorTxHash` fetching fails.

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -256,6 +267,219 @@ async function _verifyFromEtherscan({
...
+        });
+        continue;
+      }
+
+      return {
+        errorExport: {
+          customCode: "internal_error",
+          errorId: uuidv4(),
+        },
+      };
```
manuelwedler: Again, just throw here.

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -256,6 +267,219 @@ async function _verifyFromEtherscan({
...
+        verificationExport: verification.export(),
+      };
+    }
+  }
+
+  return {
+    errorExport: {
+      customCode: "no_similar_match_found",
+      errorId: uuidv4(),
+    },
```
manuelwedler, line 395: Should we include the `onchainRuntimeCode`, `onchainCreationCode` and `creationTransactionHash` here?

manuelwedler (changes requested):
Good job! looks already pretty clean. I see mainly some room for code reuse and refactoring, and more tests.

services/server/src/server/services/utils/Database.ts
marcocastignoli: mmmh I really don't know:
1) if it makes a difference
2) logically I like the idea that the query is used to extract compilations_ids so it makes sense to search for compilation's bytecodes

services/server/src/server/services/workers/verificationWorker.ts
manuelwedler: I opened a PR for the eslint rule: https://github.com/argotorg/sourcify/pull/2496

services/server/src/server/services/VerificationService.ts
marcocastignoli: From my point of view this is not a problem. why does it break error propagation? From codex:

> At services/server/src/server/services/VerificationService.ts (line 377), the handler passed to .then is declared async, so it always returns a promise. Any synchronous throw inside it, or any rejection from an await (like the setJobError call at lines 388‑395), will reject that promise and flow into the .catch block at line 409. Likewise, if the initial performServiceOperation("getSimilarityCandidatesByRuntimeCode", …) call rejects, the same .catch handles it. Mixing async/await with .then/.catch is mostly a style question; it doesn’t break error propagation in this case.

services/server/src/server/services/VerificationService.ts
marcocastignoli: As replied above

services/server/src/server/services/workers/verificationWorker.ts
marcocastignoli: What should we throw here?

services/server/src/server/services/workers/verificationWorker.ts
marcocastignoli: Can it fail? I see also in the rest of the v2 handlers (e.g json) that it is never catched.

services/server/src/server/services/workers/verificationWorker.ts
marcocastignoli: Why?

services/server/src/server/apiv2/verification/verification.routes.ts
marcocastignoli: Does it make sense to test for each endpoint the same middleware?

services/server/src/server/services/workers/verificationWorker.ts
marcocastignoli: not necessary

services/server/src/server/services/workers/verificationWorker.ts
marcocastignoli: I'm worried that there will be many conflicts, can we first merge this and only then re-do what you did in #2496?

services/server/src/apiv2.yaml
marcocastignoli: makes sense, but let's make sure that everything is always formatted.

services/database/migrations/20251101120000_add_code_first_75_index.sql
marcocastignoli: I want to run it manually when we'll merge. I will remove before merging

services/server/src/server/services/utils/Database.ts
manuelwedler: Okay,  fine with me :+1:

services/server/src/apiv2.yaml
manuelwedler: For that, we should enforce our prettier checks on yaml files. Currently, they only run on .ts

manuelwedler: created a pr for it: #2499

services/server/src/server/apiv2/verification/verification.routes.ts
manuelwedler: From my point of view, yes. Otherwise, you don't don't if it was removed by accident or it doesn't work as intended anymore. We can refactor the checks into common functions if you think it's being duplicated too much.

manuelwedler: We actually already have common functions like `testAlreadyBeingVerified`

services/server/src/server/services/workers/verificationWorker.ts
manuelwedler: sure makes sense

services/server/src/server/services/VerificationService.ts
manuelwedler: True, sorry, I was wrong about the error handling. The logic should be fine. Stylistically it looks still very odd to me.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
manuelwedler: rereading my comment and I realised I copied the wrong function name

services/server/src/server/services/workers/verificationWorker.ts
manuelwedler: For debugging purposes. Same reasoning why we store it on the other verification endpoints.

manuelwedler: It would be stored to the database I meant.

services/server/src/server/services/workers/workerTypes.ts
```diff
@@ -36,6 +37,14 @@ export interface VerifyFromEtherscanInput extends VerificationWorkerInput {
   etherscanResult: EtherscanResult;
 }
 
+export interface VerifySimilarityInput extends VerificationWorkerInput {
+  chainId: string;
+  address: string;
+  runtimeBytecode: string;
+  creatorTxHash?: string;
```
manuelwedler:
```suggestion
  creationTransactionHash?: string;
```

For consistency, the other interfaces of the worker use this naming

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -25,17 +26,21 @@ import type {
   VerifyFromMetadataInput,
   VerifyOutput,
   VerificationWorkerInput,
+  VerifySimilarityInput,
 } from "./workerTypes";
 import logger, { setLogLevel } from "../../../common/logger";
 import { getCompilationFromEtherscanResult } from "../utils/etherscan-util";
 import { asyncLocalStorage } from "../../../common/async-context";
+import SourcifyChainMock from "../utils/SourcifyChainMock";
+import { getAddress } from "ethers";
```
manuelwedler: unused import

services/server/src/server/services/utils/database-util.ts
```diff
@@ -858,121 +864,86 @@ export async function getDatabaseColumnsFromVerification(
...
-      );
-      throw new Error("Verified contract not found");
-    }
+export function prepareCompilerSettingsFromVerification(
+  verification: VerificationExport,
+): CompilerSettings {
+  return removeOutputSelectionFromSettings(
+    verification.compilation.jsonInput.settings,
+  );
+}
```
manuelwedler, line 873: Sorry about me copy-pasting the wrong function name earlier. I think `prepareCompilerSettingsFromVerification` function should never have been touched here.

Also there is a `prepareCompilerSettingsFromJsonInput` now that doesn't make sense

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
marcocastignoli: yeah yeah I understood

services/server/src/server/services/VerificationService.ts
marcocastignoli: idk, to me it seems fine, otherwise also managing the early errors would become quite difficult

services/server/src/server/services/workers/verificationWorker.ts
```diff
@@ -256,6 +267,148 @@ async function _verifyFromEtherscan({
...
+    return createPreRunCompilationFromStoredCandidate(
+      { solc, vyper },
+      candidate,
+    );
+  } catch (error: any) {
+    throw new Error(
+      `Failed to create PreRunCompilation from similarity candidate: ${error.message}`,
+    );
+  }
+}
```
manuelwedler: This function doesn't add any value. it just calls another function

manuelwedler: you can just call createPreRunCompilationFromStoredCandidate directly

services/server/src/server/services/utils/database-util.ts
```diff
@@ -858,121 +864,86 @@ export async function getDatabaseColumnsFromVerification(
...
+    logger.error("createPreRunCompilationFromStoredCandidate: missing fields", {
+      hasJsonInput: !!jsonInput,
+      hasJsonOutput: !!jsonOutput,
+      hasVersion: !!version,
+      hasFullyQualifiedName: !!fullyQualifiedName,
+      hasCreationCborAuxdata: !!creationCborAuxdata,
+      hasRuntimeCborAuxdata: !!runtimeCborAuxdata,
+    });
+    throw new Error("Compilation properties not found");
+  }
```
manuelwedler: you don't need all of these checks anymore. The point of introducing `SimilarityCandidate` was reducing the necessity of such checks.

services/server/src/apiv2.yaml
```diff
@@ -444,6 +490,16 @@ components:
...
+    SimilarityVerificationRequest:
+      description: ""
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              creationTransactionHash:
+                $ref: "#/components/schemas/Keccak256"
+                description: The hash of the transaction that created this contract. Optional.
```
manuelwedler: The description will be pruned. Please use `allOf`. See https://github.com/argotorg/sourcify/pull/2482

services/server/src/server/services/workers/verificationWorker.ts
marcocastignoli: we decided not to store them because there are no use-cases to use the stored data after calling the similarity verification

services/server/src/server/apiv2/verification/verification.routes.ts
marcocastignoli: Ok I added these tests

services/server/src/server/services/VerificationService.ts
```diff
@@ -345,13 +351,103 @@ export class VerificationService {
...
+            customCode: "internal_error",
+            errorId: uuidv4(),
+          },
+        ]);
+      });
+
+    this.runningTasks.add(similarityLookupTask);
+    similarityLookupTask.finally(() => {
+      this.runningTasks.delete(similarityLookupTask);
+    });
```
manuelwedler: I think it's a problem for the shutdown logic that we add another task inside this promise through verifyViaWorker. I think the promises should only be added once for each chained asynchronous background operation. 

To address this, I would propose to move the logic for adding and deleting tasks in another wrapper function outside of verifyViaWorker.

Let's have such a function:

```ts
runInBackground(promise) {
  const task = promise
      .finally(() => {
        this.runningTasks.delete(task);
      });
  this.runningTasks.add(task);
}
```

Remove the logic from verifyViaWorker and just return the promise:
```ts
  private verifyViaWorker(
    verificationId: VerificationJobId,
    functionName: string,
    input:
      | VerifyFromJsonInput
      | VerifyFromMetadataInput
      | VerifyFromEtherscanInput
      | VerifySimilarityInput,
  ): Promise<void> {
    return this.workerPool
      .run(input, { name: functionName })
    ...
  }
```

Then just call your promise wrapped with 
```ts
this.runInBackground(this.storageService
      .performServiceOperation("getSimilarityCandidatesByRuntimeCode", [
        runtimeBytecode,
        DEFAULT_SIMILARITY_CANDIDATE_LIMIT,
      ]).then(...)
)
```

The other verification methods need to be adapted too then, like

```ts
this.runInBackground(this.verifyViaWorker(verificationId, "verifyFromMetadata", input))
```

services/server/test/helpers/helpers.ts
```diff
@@ -474,10 +474,11 @@ export function hookIntoVerificationWorkerRun(
...
       }
-      // Wait for workers to complete
-      await Promise.all(
-        serverFixture.server.services.verification["runningTasks"],
-      );
+      const runningTasks =
+        serverFixture.server.services.verification["runningTasks"];
+      while (runningTasks.size > 0) {
+        await Promise.all(runningTasks);
+      }
```
manuelwedler: This shouldn't be needed if we adapt the task handling as I proposed.

services/server/src/server/services/workers/verificationWorker.ts
manuelwedler: Nevermind, I think it's fine now

manuelwedler (approved):

PR #2497, opened 2025-11-05
Match deletion script

manuelwedler:
For the case, we have to delete data due to accidental verifications. I made it support BigQuery and Postgres via programmatically adapting the queries for BigQuery syntax.

### DIFF
--- a/scripts/delete-sourcify-match.ts
+++ b/scripts/delete-sourcify-match.ts
@@ -106,20 +106,44 @@ class PostgresClient implements DatabaseClient {
 // BigQuery client wrapper
 class BigQueryClient implements DatabaseClient {
   private bigquery: any;
-  private session: any;
+  private sessionId: string | null = null;
   private dataset: string;
+  private location: string;
 
-  constructor(bigquery: any, dataset: string) {
+  constructor(bigquery: any, dataset: string, location: string) {
     this.bigquery = bigquery;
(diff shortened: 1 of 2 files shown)


manuelwedler:
I made a few fixes. Do you want to have another look before I merge this @marcocastignoli ?

scripts/delete-sourcify-match.ts
```diff
@@ -677,8 +699,19 @@ async function deleteVerifiedContracts(
       `✓ Deleted ${stats.codeEntriesDeleted} code entrie(s) (not referenced elsewhere)`,
     );
 
-    await client.commitTransaction();
-    console.log("\n✅ Transaction committed successfully\n");
+    // Check if this is a dry run
+    const isDryRun = process.env.DRY_RUN === "true";
+
+    if (isDryRun) {
+      await client.rollbackTransaction();
```
marcocastignoli, line 759: what's the need for dryrun if you rollback in it? just to see if it runs? I see this useful if we can somehow output what it would be deleted

manuelwedler: It outputs ids and the number of rows for each table it would have deleted. It is to see if it works and doesn't delete anything unexpected.

marcocastignoli: Ah sorry I didn't notice!

Issue #2424, opened 2025-10-07
Include external verifiers verificationId in the v2 contract lookup endpoint


marcocastignoli:
Also display the information in the verification ui

marcocastignoli:
Closed by #2504

PR #2504, opened 2025-11-11
Add external verification links to verification job responses

marcocastignoli:
See #2424

**Notice**
I had to find a way to pass the etherscan/routescan/blockscout api url to the `SourcifyDatabaseService.getVerificationJob()` function, I tried to keep all the functions related to external verification functionality inside the `EtherscanVerifyApiService` storage service. Let me know if the code is not clear, the only alternative that I have in mind is to implement a mechanism so that every storage service can access functions from the other services.

### DIFF
--- a/services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
+++ b/services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
@@ -315,9 +315,9 @@ export const buildJobExternalVerificationsObject = (
   }
   return Object.keys(externalVerification).reduce((verifiersData, verifier) => {
     const verifierIdentifier = verifier as EtherscanVerifyApiIdentifiers;
-    const verifierService = storageService.wServices[
-      verifierIdentifier
-    ] as EtherscanVerifyApiService | undefined;
+    const verifierService = storageService.wServices[verifierIdentifier] as
+      | EtherscanVerifyApiService
+      | undefined;
     const verifierData = externalVerification[verifierIdentifier];
     if (!verifierData) {
       return verifiersData;
@@ -345,9 +345,7 @@ export const buildJobExternalVerificationsObject = (
     let explorerUrl;
     if (verifierData.verificationId) {
       try {
-        const apiBaseUrl = verifierService?.getExplorerUrl(
-          parseInt(chainId),
-        );
+        const apiBaseUrl = verifierService?.getExplorerUrl(parseInt(chainId));
         if (apiBaseUrl) {
           explorerUrl = apiBaseUrl.replace(
             "${ADDRESS}",


services/server/src/apiv2.yaml
```diff
@@ -536,6 +536,19 @@ components:
                     type: string
                     description: Time it took to compile the contract on the server in milliseconds.
                     example: "1333"
+                  externalVerificationLinks:
```
manuelwedler: I think it's nice to return the links to the api directly, but we should also return the raw guids. Rather return an object for each verification service. The current structure is not extensible.

services/server/src/server/types.ts
```diff
@@ -124,6 +124,17 @@ export type ProxyResolution = Partial<ProxyDetectionResult> & {
...
 
+export interface ExternalVerificationData {
+  url?: string;
+  error?: string;
+}
+
+export interface ExternalVerificationLinks {
+  etherscan?: ExternalVerificationData;
+  blockscout?: ExternalVerificationData;
+  routescan?: ExternalVerificationData;
```
manuelwedler: these types don't match with the structure defined in the apiv2.yaml

manuelwedler: I think the types here are better because they use an object for each verifier and this makes the data extensible.

services/server/src/server/types.ts
```diff
@@ -132,6 +143,7 @@ export interface VerificationJob {
   compilationTime?: string;
   error?: MatchingErrorResponse;
   contract: VerifiedContractMinimal;
+  externalVerificationLinks?: ExternalVerificationLinks;
```
manuelwedler: How about simplifying this field's name to `externalVerifications` instead of `externalVerificationLinks`?

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
```diff
@@ -262,6 +263,96 @@ interface EtherscanRpcResponse {
...
 
+// Helper function that returns the getApiUrl() function from the initialized StorageService
+export function createGetEtherscanVerifyApiServiceApiUrl(
+  storageService: StorageService,
+) {
+  return (
+    service: EtherscanVerifyApiIdentifiers,
+    action: string,
+    chainId: number,
+    includeApiKey = true,
```
manuelwedler: we shouldn't include api keys in the response. I don't see why we should have this parameter at all.

services/server/src/server/apiv2/jobs/jobs.handler.ts
```diff
@@ -21,7 +22,10 @@ export async function getJobEndpoint(req: GetJobRequest, res: GetJobResponse) {
 
   const job = await services.storage.performServiceOperation(
     "getVerificationJob",
-    [req.params.verificationId],
+    [
+      req.params.verificationId,
+      createGetEtherscanVerifyApiServiceApiUrl(services.storage),
+    ],
   );
```
manuelwedler, line 26: I don't see why we have to generate the links inside the SourcifyDatabaseService. The code would be much cleaner if we simply get the external verification data from the database here, and then adapt the returned job object within this handler function by adding the links.

manuelwedler (changes requested):
Mainly see my last comment. I think the `SourcifyDatabaseService` shouldn't fetch data from sources other than the db (in this case it fetches from a passed function). We can simply modify the object returned from the db inside the api handler before sending it to the user.

services/server/src/server/types.ts
marcocastignoli: sorry, I forgot to update the yaml file :D

services/server/src/server/apiv2/jobs/jobs.handler.ts
marcocastignoli: I like the idea that the type of the returned job from the sourcifyDatabaseService match with the types of the returned verificationjob. But I agree that it makes more sense to apply a transformation after the returned databse job and before req.send()

services/server/src/server/types.ts
```diff
@@ -124,14 +128,36 @@ export type ProxyResolution = Partial<ProxyDetectionResult> & {
...
+export type ApiExternalVerification = ExternalVerification & {
+  statusUrl?: string;
+  explorerUrl?: string;
+};
+
+export interface ApiExternalVerifications {
+  etherscan?: ApiExternalVerification;
+  blockscout?: ApiExternalVerification;
+  routescan?: ApiExternalVerification;
+}
```
manuelwedler, line 140:
```suggestion
export type ApiExternalVerifications = Record<EtherscanVerifyApiIdentifiers, ApiExternalVerification>;
```

Would this work?

services/server/src/apiv2.yaml
```diff
@@ -1255,6 +1257,33 @@ components:
...
+      properties:
+        verificationId:
+          type: string
+          description: Identifier returned by the external verifier when the job was created.
+          example: "x3ryqcqr1zdknhfhkimqmizlcqpxncqc6nrvp3pgrcpfsqedqi"
+        error:
+          type: string
+          description: Error message returned by the external verifier. Present when the verification failed.
+          example: "NOTOK: Max rate limit reached"
+        url:
```
manuelwedler: We need `statusUrl` and `explorerUrl`.

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
```diff
@@ -262,9 +303,130 @@ interface EtherscanRpcResponse {
...
+
+// Helper function that returns the getExplorerUrl() function from the initialized StorageService
+function createGetEtherscanVerifyApiServiceExplorerUrl(
+  storageService: StorageService,
+) {
+  return (service: EtherscanVerifyApiIdentifiers, chainId: number) =>
+    (
+      storageService.wServices[service] as EtherscanVerifyApiService | undefined
+    )?.getExplorerUrl.bind(storageService.wServices[service])(chainId);
+}
```
manuelwedler: These two helper functions seem to complicate more than they improve. Can't we just access the related service inside `buildJobExternalVerificationsObject` and call the function on it directly? As far as I can see, we can get rid of these two helpers.

services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
```diff
@@ -262,9 +303,130 @@ interface EtherscanRpcResponse {
...
+
+export type GetExternalVerificationApiUrl = (
+  service: EtherscanVerifyApiIdentifiers,
+  action: string,
+  chainId: number,
+) => string | undefined;
```
manuelwedler: This type is not needed anymore.

services/server/src/server/types.ts
marcocastignoli: EtherscanVerifyApiIdentifiers are structured like "EtherscanVerify". I think I prefer having just the name of the verifier.

services/server/src/apiv2.yaml
marcocastignoli: Good catch! I always forget this :S

LonglifeXD (commented):
XDC

PR #2521, opened 2025-11-26
Add support for Yul compilation and verification

marcocastignoli:
See https://github.com/argotorg/sourcify/issues/2517
- Implement YulCompilation class for handling Yul contracts.
- Extend compilation utilities to support Yul alongside Solidity and Vyper.
- Update VerificationService to handle Yul contracts.
- Add tests for Yul compilation and verification (including a new alliance test)

Notice:
- Yul doesn't support metadata so we are generating one like we do for Vyper
- Yul doesn't support ABI so I had to update signatures and tests functions to support this case

### DIFF
--- a/packages/lib-sourcify/src/utils/utils.ts
+++ b/packages/lib-sourcify/src/utils/utils.ts
@@ -58,7 +58,7 @@ export function convertLibrariesToMetadataFormat(
  * metadata format: { "contracts/1_Storage.sol:Journal": "0x..." }
  * jsonInput format: { "contracts/1_Storage.sol": { Journal: "0x..." } }
  */
-export function convertLibrariesFromMetadataFormat(
+export function convertLibrariesToStdJsonFormat(
   metadataLibraries?: MetadataCompilerSettings['libraries'],
 ): Libraries | undefined {
   if (!metadataLibraries) {
--- a/packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
+++ b/packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts
@@ -32,7 +32,7 @@ import {
   getVariationsByContentHash,
 } from './variationsUtils';
 import { logDebug } from '../logger';
-import { convertLibrariesFromMetadataFormat } from '../utils/utils';
+import { convertLibrariesToStdJsonFormat } from '../utils/utils';
 
 export class SolidityMetadataContract {
   metadata: Metadata;
@@ -316,7 +316,7 @@ export class SolidityMetadataContract {
 
     this.solcJsonInput.language = this.metadata.language;
 
-    const libraries = convertLibrariesFromMetadataFormat(metadataLibraries);
+    const libraries = convertLibrariesToStdJsonFormat(metadataLibraries);
     if (libraries) {
       this.solcJsonInput.settings.libraries = libraries;
     }


packages/lib-sourcify/src/Compilation/YulCompilation.ts
```diff
@@ -0,0 +1,71 @@
...
+      libraries: convertLibrariesToMetadataFormat(
+        this.jsonInput.settings.libraries,
+      ),
+    };
+    this._metadata = {
+      compiler: { version: this.compilerVersion },
+      language: 'Yul',
+      output: outputMetadata,
+      settings: {
+        ...metadataSettings,
```
kuzdogan, line 66: Small thing, this should exclude `outputSelection`, it's not in metadata.settings: https://docs.soliditylang.org/en/latest/metadata.html

Same change needed in Vyper metadata generation

services/server/src/server/services/utils/compilation.ts
kuzdogan, line 1: I feel like this should be in server instead? It isn't used anywhere in lib-sourcify, and we also don't refer to this in Readme as reference example

packages/lib-sourcify/src/utils/utils.ts
```diff
@@ -21,3 +26,29 @@ export function splitFullyQualifiedName(fullyQualifiedName: string): {
...
+ * jsonInput format: { "contracts/1_Storage.sol": { Journal: "0x..." } }
+ * metadata format: { "contracts/1_Storage.sol:Journal": "0x..." }
+ */
+export function convertLibrariesToMetadataFormat(
+  libraries?: Libraries,
+): MetadataCompilerSettings['libraries'] {
+  if (!libraries) {
+    return undefined;
+  }
+
```
kuzdogan, line 41: Isn't this a duplicate of this code? 

https://github.com/argotorg/sourcify/blob/a1f21e77b49231dd93736b0bb2dcc215de339685/packages/lib-sourcify/src/Validation/SolidityMetadataContract.ts#L320-L327

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -914,7 +914,7 @@ export class SourcifyDatabaseService
 
       const abi = verification.compilation.contractCompilerOutput.abi;
       if (!abi) {
-        throw new Error("No ABI found in compilation output");
+        return;
       }
```
kuzdogan, line 921: Let's check if the language is Yul here but still throw otherwise

services/server/test/helpers/helpers.ts
```diff
@@ -57,7 +57,7 @@ export async function deployFromAbiAndBytecodeForCreatorTxHash(
   bytecode: BytesLike | { object: string },
   args?: any[],
 ): Promise<DeploymentInfo> {
-  const contractFactory = new ContractFactory(abi, bytecode, signer);
+  const contractFactory = new ContractFactory(abi || [], bytecode, signer);
```
kuzdogan, line 60: We should set `abi` to be undefined above as a type, if we expect it to be undefined

packages/lib-sourcify/src/utils/utils.ts
marcocastignoli: It's the opposite code, now I created another function so it's more clear. Let me know if it is more understandable now

packages/lib-sourcify/src/utils/utils.ts
```diff
@@ -52,3 +52,36 @@ export function convertLibrariesToMetadataFormat(
 
   return Object.keys(metadataLibraries).length ? metadataLibraries : undefined;
 }
+
+/**
+ * Converts libraries from the metadata format to the solc JSON input format.
+ * metadata format: { "contracts/1_Storage.sol:Journal": "0x..." }
+ * jsonInput format: { "contracts/1_Storage.sol": { Journal: "0x..." } }
+ */
+export function convertLibrariesFromMetadataFormat(
```
kuzdogan: Can you rename this to `convertLibrariesToStdJsonFormat`? It's confusing

PR #2531, opened 2025-12-05
Clean up PulseChain RPC URLs in chains.json

dreaded369:
Removed no longer available WebSocket RPC URLs.


dreaded369:
oops... this is imported from chains.list.

PR #2535, opened 2025-12-08
Add contractApiUrl to externalVerifcations returned properties

marcocastignoli:
In order to display the contract verification status in the UI, we need to call an Etherscan endpoint to retrieve information about the contract, checking its existence.

With this PR we add `contractApiUrl` as an additional property of the externalVerifications object, together with statusUrl (to check the status of the verification) and `explorerUrl` (link to the explorer contract's page): this new property contains a link to the `getabi` endpoint, that the user can call to check the existence of the contract on the explorer.

### DIFF
--- a/services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
+++ b/services/server/src/server/services/storageServices/EtherscanVerifyApiService.ts
@@ -324,18 +324,27 @@ export const buildJobExternalVerificationsObject = (
     }
 
     let statusUrl;
+    let contractApiUrl;
     if (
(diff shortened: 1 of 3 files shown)


kuzdogan:
Why are we calling `getabi` and not `getsourcecode`? It just felt a little unintuitive

marcocastignoli:
I tried to call `getsourcecode` and, from what I've seen, the status "0" and "1" doesn't always follow the same result for all block explorers :S while `getabi` is more reliable

kuzdogan (approved):

PR #2539, opened 2025-12-10
Implement similarity verification in Monitor

marcocastignoli:
See https://github.com/argotorg/sourcify/issues/2506

This PR:
- Adds support for similarity verification in monitor as fallback in any case in which the contract cannot be assembled via IPFS (ipfs is disabled, errors while assembling contract or not fully assembled)
- Adds a test for the new similarity verification functionality
- I noticed wrong types in the Monitor constructor so I fixed them

### DIFF
--- a/services/monitor/src/ChainMonitor.ts
+++ b/services/monitor/src/ChainMonitor.ts
@@ -290,7 +290,7 @@ export default class ChainMonitor extends EventEmitter {
           address,
           origin: metadataHash.origin,
         });
-        await this.similarityVerificationClient.trigger(
+        this.similarityVerificationClient.trigger(
           this.sourcifyChain.chainId,
           address,
           creatorTxHash,
@@ -309,7 +309,7 @@ export default class ChainMonitor extends EventEmitter {
         await pendingContract.assemble();
       } catch (err: any) {
         this.chainLogger.info("Couldn't assemble contract", { address, err });
-        await this.similarityVerificationClient.trigger(
+        this.similarityVerificationClient.trigger(
           this.sourcifyChain.chainId,
           address,
           creatorTxHash,
@@ -321,7 +321,7 @@ export default class ChainMonitor extends EventEmitter {
           address: pendingContract.address,
           pendingSources: pendingContract.pendingSources,
         });
-        await this.similarityVerificationClient.trigger(
+        this.similarityVerificationClient.trigger(
           this.sourcifyChain.chainId,
           address,
           creatorTxHash,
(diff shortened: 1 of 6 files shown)


services/monitor/src/ChainMonitor.ts
```diff
@@ -365,6 +381,36 @@ export default class ChainMonitor extends EventEmitter {
...
+    address: string,
+  ) {
+    if (!this.similarityVerificationClient) {
+      this.chainLogger.debug("Similarity search disabled, skipping fallback", {
+        address,
+      });
+      return;
+    }
+
+    this.chainLogger.info("Triggering similarity search fallback", {
```
kuzdogan: This whole section seems unnecessary and can be handled inside the SimilarityVerificationClient. Then here we can simple call `this.client.trigger()`

services/monitor/src/SimilarityVerificationClient.ts
```diff
@@ -0,0 +1,76 @@
+import logger from "./logger";
+import type { Logger } from "winston";
+
+const trimTrailingSlash = (url: string) => url.replace(/\/+$/, "");
+
+export default class SimilarityVerificationClient {
+  private baseUrls: string[];
+  private clientLogger: Logger;
+  private enabled: boolean;
```
kuzdogan: What's the use case of having `enabled`? We always start it with `true` and don't change at all no?

kuzdogan (commented):
Small comments. Looks good in general. Just, we should check if the monitor is really working on prod now. I checked the logs and there seemed to be a lot of errors and no INFO level logs

manuelwedler:
One question regarding this. Do I remember correctly that we wanted to have a delay between contract deployment and similarity search submission, such that it is possible for users to verify themselves before monitor does? Is this implemented in any way here?

marcocastignoli:
> One question regarding this. Do I remember correctly that we wanted to have a delay between contract deployment and similarity search submission, such that it is possible for users to verify themselves before monitor does? Is this implemented in any way here?

Right!!! I totally forgot about it while implementing this PR. Do you think is it as straightforward as as adding a delay before triggering each similarity request? That would be easy to implement. IMO it's not worth to implement any other advanced mechanism involving queues.

manuelwedler:
yes, simply add a delay, but don't await the call to not block anything. I don't think this needs any fancy mechanism. Also make it configurable through an env variable.

marcocastignoli:
@manuelwedler I pushed this change. Good catch! thanks :)

kuzdogan (approved):

PR #2547, opened 2025-12-15
Add Sei Network specific Etherscan API key variable

manuelwedler:
we missed to add a specific api key in https://github.com/argotorg/sourcify/pull/2537


marcocastignoli (approved):
lgtm

Issue #2441, opened 2025-10-14
Parquet export v2


manuelwedler:
### Current issues

The parquet export has multiple issues at the moment, which make it difficult to use. At the moment, for most use cases, it is required to download all export files to be sure to have all referenced rows for one verified contract. People also complaint that it is not possible to know if one parquet file changed since the last export (#2212, #1683). Another issue are monetary costs of the export, which mainly come from the upload to CloudFlare (#2422). 

### Solution

To address all of these issues, I propose to do the following:

- [x] Store the parquet files in a Google Bucket (-> no network costs)
- [x] When creating the export, order by `created_at` column
  - [x] Makes the export incremental (all tables excluding `sourcify_matches` are append-only)
  - [x] We will need an index on `created_at` on all tables but I think that's okay to have for this use case. For the tables with incremental ids, it should be sufficient to use the id instead.
  - [x] If we also output the `created_at` of the first row per file in the manifest, we can make the export job running only from the last file, and generating only new data since the last export. The job would run much faster.
  - [x] For `sourcify_matches`, we could order by `updated_at`. We would duplicate entries in the export files when a match was upgraded, but I think duplication is not an issue. People should just assume that the last row is the up-to-date version.
- [x] Add a checksum per file in the manifest
- [x] Add the upload date per file in the manifest
- [x] Add the size of each file in the manifest (https://github.com/verifier-alliance/parquet-export/issues/4)
- [x] GCP buckets return a XML file with a checksum, upload date and file size for each file when you query the root `/` of the bucket. This is all the information we need in the manifest. We should consider replacing our custom generated manifest with the default metadata XML from GCP.
- [x] Revisit the size per parquet file. It's more efficient to have smaller file sizes. Research what a good size per parquet file is.
- [x] Add compression to parquet files (#1680)
- [x] Documentation about the new structure. Especially, make sure to explain about the possible duplication in the `sourcify_matches` export (in case of match upgrades), and tell people to use upserts.
- [x] Side note: check if alerts for failing jobs are set up properly on GCP. I think VerA staging is missing an alert.

**Important**: We can implement the new export in a separate bucket. We can make only requests to `/v2` being routed to the new bucket via the load balancer. We will keep the last export of the v1 bucket and stay compatible with anyone using it.

### Future work

In #2235, a use case is described, where data for a specific chain is needed. We can facilitate that later via #2500 .

kuzdogan:
[...]

PR #2551, opened 2025-12-19
Add migrations for adding indexes on `created_at` columns

manuelwedler:
Related to #2441

For the parquet export v2, we need indexes on these columns to order them efficiently.

### DIFF
--- a/services/database/migrations/20251219160923_add_created_at_indexes_sourcify.sql
+++ b/services/database/migrations/20251219160923_add_created_at_indexes_sourcify.sql
@@ -1,5 +1,15 @@
 -- migrate:up
 
+/*
+  Add indexes on created_at/updated_at columns for Sourcify-specific tables.
+  These indexes are required for the Parquet Export v2:
+  https://github.com/argotorg/sourcify/issues/2441
+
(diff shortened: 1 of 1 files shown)


services/database/migrations/20251219160923_add_created_at_indexes_sourcify.sql
kuzdogan, line 1: Can you add some context in this file in comments why we make these changes? Including a link to the issue

services/database/sourcify-database.sql
```diff
@@ -2074,4 +2159,6 @@ INSERT INTO public.schema_migrations (version) VALUES
     ('20250922141802'),
     ('20251009141621'),
     ('20251023134207'),
-    ('20251101120000');
+    ('20251101120000'),
+    ('20251106144315'),
+    ('20251219160923');
```
kuzdogan, line 2164: Why is there two added migrations? We forgot one?

manuelwedler: One comes from the database-specs repo: https://github.com/verifier-alliance/database-specs/blob/master/migrations/20251106144315_add_created_at_indexes.sql

Issue #2529, opened 2025-12-04
VerifierAllianceDatabase fails to update contracts


marcocastignoli:
Looking at GCP logs for production I noticed we have many `cannot update verified_contract` for the Verifier Alliance storage service. These cases are only for contracts that need update. New contracts insertion works normally.

We need also to investigate if it's correct to update Vera contracts in the first place (I remember we decided to never update contracts on Vera).

### Error details
```
Error: cannot update verified_contract address=0x7364fa77a2ef8eefcc485fee10e6b3e60736d79c chainId=81457
TypeError: Cannot read properties of undefined (reading 'id')
    at AllianceDatabaseService.updateExistingVerifiedContract (/home/app/services/server/dist/server/services/storageServices/AbstractDatabaseService.js:138:19)
```

The problem is that one of the rows[0].id here is failing:
```js
            // add the onchain contract in contract_deployments
            const contractDeploymentInsertResult = await this.database.insertContractDeployment(client, Object.assign(Object.assign({}, databaseColumns.contractDeployment), { contract_id: contractInsertResult.rows[0].id }));
            const contractDeploymentId = contractDeploymentInsertResult.rows[0].id;
            // Add recompiled bytecodes
            let recompiledCreationCodeInsertResult;
            if (databaseColumns.recompiledCreationCode) {
                recompiledCreationCodeInsertResult = await this.database.insertCode(client, databaseColumns.recompiledCreationCode);
            }
            const recompiledRuntimeCodeInsertResult = await this.database.insertCode(client, databaseColumns.recompiledRuntimeCode);
            // insert new recompiled contract
            const compiledContractsInsertResult = await this.database.insertCompiledContract(client, Object.assign(Object.assign({}, databaseColumns.compiledContract), { creation_code_hash: recompiledCreationCodeInsertResult === null || recompiledCreationCodeInsertResult === void 0 ? void 0 : recompiledCreationCodeInsertResult.rows[0].bytecode_hash, runtime_code_hash: recompiledRuntimeCodeInsertResult.rows[0].bytecode_hash }));
            const compiledContractId = compiledContractsInsertResult.rows[0].id;
            await this.database.insertCompiledContractsSources(client, {
                sourcesInformation: databaseColumns.sourcesInformation,
                compilation_id: compiledContractId,
            });
            // update verified contract with the newly added recompiled contract
[...]

PR #2552, opened 2025-12-22
Fix verified contracts unique constraint error

marcocastignoli:
See [#2529](https://github.com/argotorg/sourcify/issues/2529)

The verifier alliance storage service fails for contract update verification. In reality we never update information but we always insert a new verified contract.

The problem happens when trying to insert a verified contract with already existing compilation_id and deployment_id. The insert query didn't have `DO NOTHING` in case of a constraint error. We never see this case with the `SourcifyDatabase` because we never insert an already existing contract, we throw "ConflictError" instead in `StorageService.storeVerification()`.

This PR adds `DO NOTHING` to the verified contract insert query and then handles the case in which no new verified_contract is created by throwing a "ConflictError"

### DIFF
--- a/services/server/test/integration/storageServices/SourcifyDatabaseService.spec.ts
+++ b/services/server/test/integration/storageServices/SourcifyDatabaseService.spec.ts
@@ -70,10 +70,6 @@ describe("SourcifyDatabaseService", function () {
       expect.fail("Expected storeVerification to throw");
     } catch (error) {
       thrownError = error;
-      console.log(
-        "SourcifyDatabaseService.storeVerification error:",
-        thrownError,
-      );
     }
 
     expect(thrownError).to.be.instanceOf(ConflictError);


manuelwedler (commented):
thanks, looks good.

Only if it doesn't take too long, it would be nice to add a test for this (a new test file `AllianceDatabaseService.spec.ts` might be needed).

services/server/test/integration/storageServices/SourcifyDatabaseService.spec.ts
```diff
@@ -62,8 +63,24 @@ describe("SourcifyDatabaseService", function () {
...
+    let thrownError: unknown;
+    try {
+      await databaseService.storeVerification(MockVerificationExport);
+      expect.fail("Expected storeVerification to throw");
+    } catch (error) {
+      thrownError = error;
+      console.log(
+        "SourcifyDatabaseService.storeVerification error:",
+        thrownError,
+      );
```
manuelwedler:
```suggestion
```
I think we don't need to add logs in the test.

marcocastignoli: ah I forgot to delete that, I removed it only in AllianceDatabase

PR #2573, opened 2026-01-07
Fallback to chains.json rpcs

### LINKED ISSUE #2565: Use chains.json rpcs as fallback
Currently, if `sourcify-chains-default.json` has rpcs for a chain, we don't use the ones in `chains.json`. `chains.json` rpcs are only used if no rpcs are defined in `sourcify-chains-default.json`. We can use the `chains.json` rpcs as a fallback and merge them into the others (but keep at the end of the array).

Also, I would step back from the 1 immediate retry in the circuit breaker pattern if we have more rpcs available. The `chains.json` ones are more likely to fail and we shouldn't retry them too often.

manuelwedler:
Fixes #2565


marcocastignoli (approved):
lgtm

PR #2577, opened 2026-01-07
Prevent .only tests via lint rule

manuelwedler:
I think it happens too often that these tests are pushed or merged and it is easy to prevent this via a lint rule


marcocastignoli:
that's a good idea! sorry I always forget those around

marcocastignoli (approved):

PR #2583, opened 2026-01-11
Add Tempo Moderato Chain (42431)

kuzdogan:
Replaces the chain in extra-chains file since the prev. Tempo testnet was added to chains.json. I've updated the chains.json


manuelwedler (approved):
lgtm

PR #2585, opened 2026-01-12
Support factory contracts in Monitor

### LINKED ISSUE #2507: Add support for factory contracts in monitor
Right now monitor only checks for EOA contract creation, we need to add support to factory created contracts by using traces like we do in Sourcify's server

manuelwedler:
Closes #2507 

- Implements new methods for `SourcifyChain` to fetch created contract addresses from traces via the `debug_traceBlockByNumber` and `trace_block` rpc methods.
- Calls the new `getCreatedAddressesFromBlockTraces` method inside Monitor to run the new contract procedure on any contract found in the traces.
- Adds Quicknode rpcs to Monitor for trace support and performs some general maintenance of the RPCs.

Unfortunately, it was not possible to add tests inside Monitor because neither `debug_traceBlockByNumber` nor `trace_block` is supported by Hardhat Network. I tested it locally, and it successfully submitted factory deployed contracts for verification.

### DIFF
--- a/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
+++ b/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
@@ -656,7 +656,7 @@ export class SourcifyChain {
       const txHash = tracesForTx.txHash;
       const createCalls: CallFrame[] = [];
       this.findCreateInDebugTraceTransactionCalls(
-        (tracesForTx.result as CallFrame).calls,
+        (tracesForTx.result as CallFrame)?.calls || [],
         createCalls,
       );
       if (createCalls.length === 0) {
@@ -669,10 +669,10 @@ export class SourcifyChain {
     }
 
     logDebug('Found created addresses from create traces', {
-      numberOfCreatedAddresses: createdAddresses.length,
       blockNumber,
       maskedProviderUrl: rpc.maskedUrl,
       chainId: this.chainId,
+      createdAddresses,
     });
     return createdAddresses;
   };
--- a/services/monitor/src/ChainMonitor.ts
+++ b/services/monitor/src/ChainMonitor.ts
@@ -191,31 +191,35 @@ export default class ChainMonitor extends EventEmitter {
       }
     }
 
-    // Check factory contracts with traces
-    let factoryCreatedAddresses: Record<string, string[]>;
-    try {
-      factoryCreatedAddresses =
-        await this.sourcifyChain.getCreatedAddressesFromBlockTraces(
-          block.number,
-        );
-    } catch (error: any) {
-      this.chainLogger.error("Error fetching created addresses from traces", {
-        blockNumber: block.number,
-        error,
-      });
-      return;
-    }
-    for (const [txHash, addresses] of Object.entries(factoryCreatedAddresses)) {
-      for (const address of addresses) {
-        this.chainLogger.info(
-          "Found new contract created by factory in block",
-          {
-            blockNumber: block.number,
-            address,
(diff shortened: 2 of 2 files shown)


packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
```diff
@@ -524,11 +611,77 @@ export class SourcifyChain {
...
+          blockNumber,
+          maskedProviderUrl: rpc.maskedUrl,
+          chainId: this.chainId,
+        });
+        continue;
+      }
+      const txHash = tracesForTx.txHash;
+      const createCalls: CallFrame[] = [];
+      this.findCreateInDebugTraceTransactionCalls(
+        (tracesForTx.result as CallFrame).calls,
```
marcocastignoli: Are we sure calls always exists in result?

services/monitor/src/ChainMonitor.ts
```diff
@@ -191,6 +190,34 @@ export default class ChainMonitor extends EventEmitter {
         this.processNewContract(tx.hash, address);
       }
     }
+
+    // Check factory contracts with traces
+    let factoryCreatedAddresses: Record<string, string[]>;
+    try {
+      factoryCreatedAddresses =
```
marcocastignoli: can we check if the rpc support traces before calling this? otherwise we'll get errors even for rpc that explicitly don't support traces

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
```diff
@@ -524,11 +611,77 @@ export class SourcifyChain {
...
+        continue;
+      }
+      if (!createdAddresses[txHash]) {
+        createdAddresses[txHash] = [];
+      }
+      createdAddresses[txHash].push(...createCalls.map((call) => call.to));
+    }
+
+    logDebug('Found created addresses from create traces', {
+      numberOfCreatedAddresses: createdAddresses.length,
```
marcocastignoli: Length here is Object.keys(createdAddresses).length

manuelwedler: good catch. I changed the log to match the one in the other method, which I forgot earlier.

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
manuelwedler: It should but let's better be safe and add a check :)

services/monitor/src/ChainMonitor.ts
manuelwedler: This will only throw if there is no trace supported RPC at all for a chain. But yes, let's add a check and skip the logic otherwise.

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
```diff
@@ -487,14 +574,14 @@ export class SourcifyChain {
     );
 
     if (traces?.calls instanceof Array && traces.calls.length > 0) {
-      logInfo('Fetched tx traces', {
+      logInfo('Fetched tx traces for creation tx hash', {
```
kuzdogan, line 577: Small thing I came across now. Maybe this shouldn't be an info? Basically logs all trace fetches from what I see

manuelwedler: Yes I also saw some other logs in SourcifyChain that seem too verbose to me. Will open a PR for them.

PR #2598, opened 2026-01-14
Replace toBeHex with toQuantity for block number conversion in SourcifyChain

manuelwedler:
RPCs are returning: "invalid argument 0: hex number with leading zero digits".
Currently monitor produces a lot of errors in the logs on staging due to this.

toQuantity is the correct method for providing numbers to rpcs because it does not produce leading 0s.


marcocastignoli (approved):
lgtm, but why didn't tests catch this case?

manuelwedler:
It only happens when the current block number has a leading zero in hex

PR #2600, opened 2026-01-14
Stabilize factory deployed contract fetching inside Monitor

manuelwedler:
Related to #2507 

I wanted to make the fetching of factory created contracts more reliable. This mimics the retry mechanism we already have for the bytecode fetching. Additionally, I added a delay to Gnosis Chain because its rpc seems to provide traces relatively late after the block is available. For Polygon Amoy Testnet, trace support is removed since the rpc methods do not reply properly.

### DIFF
--- a/services/monitor/src/ChainMonitor.ts
+++ b/services/monitor/src/ChainMonitor.ts
@@ -210,7 +210,7 @@ export default class ChainMonitor extends EventEmitter {
       this.chainLogger.debug("Fetching created addresses", {
         retryNumber: this.traceNumberOfTries - i + 1,
         blockNumber: block.number,
-        bytecodeInterval: this.bytecodeInterval,
+        traceInterval: this.traceInterval,
         maxRetries: this.traceNumberOfTries,
       });
 


services/monitor/src/ChainMonitor.ts
```diff
@@ -193,33 +199,57 @@ export default class ChainMonitor extends EventEmitter {
...
+  };
+
+  private async getCreatedAddressesWithRetries(
+    block: Block,
+  ): Promise<Record<string, string[]>> {
+    for (let i = this.traceNumberOfTries; i > 0; i--) {
+      this.chainLogger.debug("Fetching created addresses", {
+        retryNumber: this.traceNumberOfTries - i + 1,
+        blockNumber: block.number,
+        bytecodeInterval: this.bytecodeInterval,
```
marcocastignoli: you wanted to log bytecodeInterval or traceInterval?

manuelwedler: oops, that's a copy-paste error

PR #2613, opened 2026-01-20
Store failed verification inputs in a S3 bucket for debugging purposes

### LINKED ISSUE #2359: Temporarily store standard json input for failed verifications
For debugging cases like https://github.com/argotorg/sourcify/issues/2358, it would be very helpful to have access to the standard json input which resulted in a failed verification.

~~We can add it to the `verification_jobs_ephemeral` table, and prune the data whenever it grows too big.~~
Better option might be storing it in a bucket.

manuelwedler:
Fixes #2359

I created a GCP bucket for staging and made it delete each file after 14 days.


manuelwedler:
The test fails. I will look into it tomorrow

PR #2617, opened 2026-01-22
Enable schema validation CI check by comparing cleaned schema files

### LINKED ISSUE #2314: Decide what to do about skipped schema validation CI checks
In #2313 we started skipping the db schema validation CI checks, because `pg_dump` had some changes in the recent postgres release and doesn't produce a consistent result across machines anymore. 

`pg_dump` adds these lines since 16.10:

```
\restrict wzpNUSKP7T1c6qgyqXCflEagKcM9dkwutqIx5cIHPVh7KHWPDu4pd6WdTgxRfqu

-- Dumped from database version 16.10 (Ubuntu 16.10-1.pgdg24.04+1)
-- Dumped by pg_dump version 16.10 (Ubuntu 16.10-1.pgdg24.04+1)

[...]

\unrestrict wzpNUSKP7T1c6qgyqXCflEagKcM9dkwutqIx5cIHPVh7KHWPDu4pd6WdTgxRfqu
```

For the restrict command there is actually an option to use a specific key (see https://www.postgresql.org/docs/16/app-pgdump.html -> `--restrict-key`), but dbmate would need to add support, or we could work around it by running the dump ourselves. But we would still have the issue of the version comments which don't have an option to opt out.

Ideally, `pg_dump` and dbmate allow consistent dumps and we can reenable the test eventually, otherwise we might just remove it.

manuelwedler:
Fixes #2314

### DIFF
--- a/.circleci/new_branch.yml
+++ b/.circleci/new_branch.yml
@@ -145,35 +145,38 @@ jobs:
           command: |
             cd services/database
             npm install
-      # Skipping the following tests because postgres 16.10 introduced changes to pg_dump
-      # which prevents us to have consistent dumps.
-      # - run:
-      #     name: Validate schema is in sync with migrations
-      #     command: |
-      #       cd services/database
-      #       # Initialize git submodules
-      #       git submodule update --init
-      #       # Apply migrations using npm script
-      #       npm run migrate:up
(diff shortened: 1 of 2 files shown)


marcocastignoli (approved):
lgtm! nice that we found a solution

PR #2619, opened 2026-01-26
Add delete type to auxdata transformation

marcocastignoli:
See https://github.com/argotorg/sourcify/issues/2602

In this PR:
- Add a new `delete` type for auxdata transformation
- refactor extractAuxdataTransformation code
- update tests

### DIFF
--- a/packages/lib-sourcify/src/Verification/Transformations.ts
+++ b/packages/lib-sourcify/src/Verification/Transformations.ts
@@ -1,4 +1,4 @@
-import { AuxdataStyle } from '@ethereum-sourcify/bytecode-utils';
+import { AuxdataStyle, isCborEncoded } from '@ethereum-sourcify/bytecode-utils';
 import type {
   ImmutableReferences,
   LinkReferences,
@@ -360,8 +360,28 @@ export function extractAuxdataTransformation(
 
       const offsetStart = recompiledAuxdataOffset;
       const offsetEnd = offsetStart + recompiledAuxdata.length;
-      // Instead of zeroing out this segment, get the value from the onchain bytecode.
+
+      // Get the value from the onchain bytecode.
       const onchainAuxdata = onchainBytecode.slice(offsetStart, offsetEnd);
+      if (
+        // We need to validate the onchain auxdata is actually a valid CBOR object
+        // If the recompiled auxdata length is different from the onchain auxdata length,
+        // then `onchainAuxdata` will contain bytes that are not part of the auxdata.
+        onchainAuxdata.length > 0 &&
+        !(
+          // We first try to decode the auxdata removing the auxdata length bytes,
+          // if it fails we try to decode it as is, since some Vyper auxdata doesn't
+          // include the auxdata length in the bytecode.
+          (
+            isCborEncoded(onchainAuxdata.slice(0, -4)) ||
+            isCborEncoded(onchainAuxdata)
+          )
+        )
+      ) {
+        throw new Error(
+          `Failed to decode onchain auxdata at offset ${offsetStart} with length ${onchainAuxdata.length}.`,
+        );
+      }
 
       if (transformationValues.cborAuxdata === undefined) {
         transformationValues.cborAuxdata = {};
--- a/packages/bytecode-utils/src/lib/bytecode.ts
+++ b/packages/bytecode-utils/src/lib/bytecode.ts
@@ -302,7 +302,7 @@ const extractExecutionBytecode = (
  * @param auxdata - The auxdata string to decode
  * @returns True if auxdata is CBOR-encoded, otherwise false
  */
-const isCborEncoded = (auxdata: string): boolean => {
+export const isCborEncoded = (auxdata: string): boolean => {
   try {
     CBOR.decode(arrayify(`0x${auxdata}`));
     return true;


packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -319,11 +324,42 @@ export function extractAuxdataTransformation(
...
+      let transformationType: 'replace' | 'delete' = 'replace';
+      // By default the transformation index is index + 1
+      let transformationIndex: string | undefined = `${index + 1}`;
+      // By default we don't store the auxdata length, because it's the same as the original one.
+      let transformationLength = undefined;
+
+      // But if the lengths are different we need to store the auxdata length in the transformation
+      if (recompiledAuxdata.length !== onchainAuxdata.length) {
+        transformationLength = recompiledAuxdata.length / 2;
+      }
```
kuzdogan: Isn't this block never entered? Because you get the `onchainAuxdata` by slicing the section (ie. same length) of the `recompiledAuxdata` so you will always have same length? 

Edit: Ah no, this works in the tests because the cborAuxdata you are testing is only at the end. When you have an `offsetEnd` that's out of bounds, it's equal to `offsetStart` effectively. You are trying to slice outside. We are not testing nested CBORs from what I see which we need to.

packages/lib-sourcify/src/Verification/Transformations.ts
kuzdogan: Small thing I notice now. Maybe we should double check that the removed bytes are indeed `ff` here

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -365,8 +373,14 @@ export function extractAuxdataTransformation(
         // TODO with this we could potentially support multiple auxdata sections, but needs more testing
         // && (true || index === Object.values(cborAuxdataPositions).length - 1)
       ) {
-        // If cborAuxdata is disabled Solidity adds a ff byte at the end of the bytecode
-        // so we need to remove it from the populated recompiled bytecode to match
+        const isFE =
```
kuzdogan, line 393: Why aren't you doing this check where you have the comment about it on line 353?

packages/lib-sourcify/src/Verification/Transformations.ts
kuzdogan: This comment still holds. This block is never entered unless the auxdata is at the end. If we want to keep it we should make the comment explicit about this case.

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -40,13 +41,17 @@ export const ConstructorTransformation = (offset: number): Transformation => ({
...
+  length?: number,
 ): Transformation => ({
-  type: 'replace',
+  type: transformationType,
   reason: 'cborAuxdata',
   offset,
-  id,
+  // Add id or length only if defined
+  ...(id != undefined ? { id } : {}),
+  ...(length != undefined ? { length } : {}),
```
manuelwedler:
```suggestion
  ...(id !== undefined ? { id } : {}),
  ...(length !== undefined ? { length } : {}),
```

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -40,13 +41,17 @@ export const ConstructorTransformation = (offset: number): Transformation => ({
 });
 
 export const AuxdataTransformation = (
```
manuelwedler, line 43: should we add basic checks in this function such that it only produces valid AuxdataTransformations?

e.g.

- id should be present when type === 'replace'
- id is not present when type === 'delete'
- length > 0

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -309,28 +314,73 @@ export function extractLibrariesTransformation(
...
+      );
+      if (!transformationValues.cborAuxdata) {
+        transformationValues.cborAuxdata = {};
+      }
       if (
         onchainAuxdata.length === 0
         // TODO with this we could potentially support multiple auxdata sections, but needs more testing
         // && (true || index === Object.values(cborAuxdataPositions).length - 1)
       ) {
-        // If cborAuxdata is disabled Solidity adds a ff byte at the end of the bytecode
```
manuelwedler, line 332: Could you explain the removal of this comment? I see we still remove some byte. Why do we have the `populatedRecompiledBytecode.slice(0, offsetStart - 2)`?

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -309,28 +314,73 @@ export function extractLibrariesTransformation(
...
+      let transformationLength = undefined;
+      // By default the transformation offset is the recompiled auxdata offset
+      let transformationOffset = recompiledAuxdataOffset;
+
+      // But if the lengths are different we need to store the auxdata length in the transformation
+      if (recompiledAuxdata.length !== onchainAuxdata.length) {
+        transformationLength = recompiledAuxdata.length / 2;
+      }
+      // And if onchain auxdata is empty, we set the transformation index to undefined and the type to 'delete'
+      if (onchainAuxdata.length === 0) {
```
manuelwedler, line 390: Why do we have two `(onchainAuxdata.length === 0)` checks in this function? I feel like the code needs some refactoring.

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -309,28 +314,73 @@ export function extractLibrariesTransformation(
...
+        AuxdataTransformation(
+          transformationType,
+          transformationOffset / 2, // Convert length in bytes
+          transformationIndex,
+          transformationLength,
+        ),
+      );
+      if (!transformationValues.cborAuxdata) {
+        transformationValues.cborAuxdata = {};
+      }
```
manuelwedler: I think this should be inside the `if (transformationIndex) {` block

packages/lib-sourcify/src/Verification/Transformations.ts
manuelwedler: From my understanding, we cannot support multiple auxdatas if the auxdata lengths of recompiled and onchain are not the same. It would still be good to document / add a comment regarding Kaan's finding.

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -309,28 +314,73 @@ export function extractLibrariesTransformation(
...
+        transformationLength = recompiledAuxdata.length / 2;
+      }
+      // And if onchain auxdata is empty, we set the transformation index to undefined and the type to 'delete'
+      if (onchainAuxdata.length === 0) {
+        transformationType = 'delete';
+        transformationIndex = undefined;
+        // By default Solidity adds 'fe' byte before the cborAuxdata when appending it
+        const isFE =
+          populatedRecompiledBytecode.slice(offsetStart - 2, offsetStart) ===
+          'fe';
```
manuelwedler, line 395: I think we cannot expect to always have the 'fe' byte. Please see my example in this issue: https://github.com/argotorg/sourcify/issues/2048 (0x2110c882c446357Ac38e36Cae03DF1578EB390b2 on polygon)

Did you understand in which case it is added?

packages/lib-sourcify/src/Verification/Transformations.ts
manuelwedler: I also think this check should be more clearly separated from the delete type, i.e. happen in a block that is never reached if `onchainAuxdata.length === 0`.

Generally, I think the logic should be split into one block for the `delete` type case and one block for the `replace` type case. I hope that makes sense to you.

packages/lib-sourcify/test/Verification/Verification.spec.ts
```diff
@@ -812,6 +842,23 @@ describe('Verification Class Tests', () => {
           },
         });
         assertCborTransformations(verification.transformations.runtime?.list);
+        expect(verification.transformations.runtime?.list[0]).to.deep.equal({
```
manuelwedler, line 849: Now I wonder again: did any of the tested cases here make it is way into our database with a different behavior than in this PR?

manuelwedler: Generally it would be good to add tests that test the behavior of the server for these cases. So, we should add them when https://github.com/argotorg/sourcify/pull/2630 is merged

packages/lib-sourcify/src/Verification/Transformations.ts
marcocastignoli: The comment was wrong. First of all the byte wasn't an ff, it was an fe byte. And it's before every appended auxdata.

packages/lib-sourcify/test/Verification/Verification.spec.ts
marcocastignoli: Yes :) I'll rebase this PR on staging once #2630 is merged, then I'll create the tests for these cases

kuzdogan:
> we won't be able to verify contracts with nested cbor auxdatas if the cbor has different lengths in onchain and recompiled

This is only true for the creation bytecode though, right?

manuelwedler:
> > we won't be able to verify contracts with nested cbor auxdatas if the cbor has different lengths in onchain and recompiled
> 
> This is only true for the creation bytecode though, right?

No, this should be true for runtime bytecode too. Or am I missing something?

packages/lib-sourcify/src/Verification/Transformations.ts
marcocastignoli: We are already supporting cases in which fe is missing because we don't check `fe` for replace cbor transformations. But we've never seen a case of delete transformation with missing `fe`. So for me it's fine to leave this throw.

packages/lib-sourcify/test/Verification/Verification.spec.ts
marcocastignoli: Done

kuzdogan:
@manuelwedler I think we can verify runtime. The case you are referring to I imagine is:
- recompiled: with CBOR
- onchain: without CBOR

What you can do is you first check the CBOR at the end of the bytecodes. If you slice the same offsets on the recompiled and onchain and if you get `""` onchain, that means the onchain is shorter and you can assume it has no CBOR. Then you strip out the CBOR sections in recompiled and check if   1. are the lenghts same?   2. Do bytecodes match?

The reason we can't verify creation is the offset byte change we talked about. We actually can even verify with that, it should have a very predictable pattern, but that's a very niche case as we said.

packages/lib-sourcify/src/Verification/Transformations.ts
```diff
@@ -309,49 +340,94 @@ export function extractLibrariesTransformation(
...
-        // so we need to remove it from the populated recompiled bytecode to match
+
+      // By default the transformation type is 'replace'
+      let transformationType: 'replace' | 'delete' = 'replace';
+      // By default the transformation index is index + 1
+      let transformationIndex: string | undefined = `${index + 1}`;
+      // By default we don't store the auxdata length, because it's the same as the original one.
+      let transformationLength = undefined;
+      // By default the transformation offset is the recompiled auxdata offset
+      let transformationOffset = recompiledAuxdataOffset;
```
manuelwedler: I think we don't need to define default values here. I find it confusing that we define them here, and later overwrite them with the same values in the replace type block.

marcocastignoli:
The problem that this PR is fixing and all the know limitations are documented here: https://github.com/verifier-alliance/database-specs/issues/39

Issue #2049, opened 2025-04-02
Generalized integration test suite for special verification cases


manuelwedler:
We have a lot of big changes at the moment on Sourcify (lib-sourcify refactoring, API v2). To get more comfortable with these, I think it would be nice to improve our testing again. Especially API v2 isn't tested as good as v1, since we don't run the `database.spec.ts` tests on it. Also there are tests in `apiv1/verification-handlers/verify.stateless.spec.ts` that I didn't want to add to the API v2 tests, because they don't test the behavior of the API but rather if our core logic handles special Solidity cases correctly.

I would like to take this chance to have a more generalized test suite for all kinds of verification cases. The `database.spec.ts` tests should serve as a basis for it, since they already have a good json structure for test cases. This test suite should be the place where we test if transformations are set correctly for all kinds of cases and if edge cases of specific compiler versions are verified correctly.

The test suite should run a ServerFixture and a ChainFixture, but not do any mocking to have an E2E kind of approach.

### Goals

- One central place where we can add special verification cases (e.g. testing behavior that is specific to a certain solc version). 
- Reusable test cases for testing all of v1 and v2 API and the database. The results in the database and in the v2 lookup endpoint should be checked.
- Complete testing of all stored fields. All database fields and all API v2 fields (`?fields=all`) should be tested.
- Reusable for Solidity and Vyper.
- Standardize the test format. We should have a typescript interface and testcase json files that fulfill this interface.
- Group the fields in the test file format into input fields for the API and expected artifacts. The latter should be the values which we check against. In the current `database.spec.ts`, I find it confusing that we partly use the same fields as input and as values to check against.
- Nice to have: Standardize the way to add more tests (e.g. a script to generate a testcase json file from compiler input).
- Nice to have: Run the ServerFixture with a Sourcify and a **VerA database** at the same time and check the correct behavior of both.

### Test cases that should be covered

- all existing from `database.spec.ts`
- test cases in `verify.stateless.spec.ts` that don't test the API behavior but instead a special Solidity case should be moved there
- match upgrades from partial to perfect
- specific Solidity versions, especially pre 0.4.11 versions (https://github.com/argotorg/sourcify/issues/2296)

manuelwedler:
Related: #2035

manuelwedler:
Implemented in #2630 and #2635 

For the remaining points I opened these issues: #2640 and #2641

PR #2630, opened 2026-02-04
Refactor database.spec.ts into a generalized verification-cases.spec.ts test suite

manuelwedler:
See #2049 

- Central place to test any weird verification cases we find
- Now uses APIv2 for verification
- Simplified and properly structured the testing code
- Better typing through the VerificationTestCase interface
- Adds documentation on the fields of a test case
- Meant to be extensible for adding testing of more storage services
- Refactored all existing verifier alliance test cases into the new format
- Tests by asserting against the database directly and against API v2 lookup
- Adds assertions for signature fields

### DIFF
--- a/services/server/test/integration/verification-cases/verification-cases.spec.ts
+++ b/services/server/test/integration/verification-cases/verification-cases.spec.ts
@@ -4,7 +4,6 @@ import {
   deployFromBytecodeForCreatorTxHash,
   hookIntoVerificationWorkerRun,
 } from "../../helpers/helpers";
-import { keccak256 } from "ethers";
 import { LocalChainFixture } from "../../helpers/LocalChainFixture";
 import { ServerFixture } from "../../helpers/ServerFixture";
 import type {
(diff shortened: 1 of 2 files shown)


marcocastignoli:
Initial comment: the idea about the verifier alliance json files is that we can share them with the other verifiers, so it would be better to keep them in a json format. In the future we could potentially treat them as an external submodule (like we do with the vera database)

services/server/test/integration/verification-cases/verification-cases.spec.ts
```diff
@@ -0,0 +1,715 @@
...
+    expectedSignatures.function.sort(sortBySignature);
+    expectedSignatures.event.sort(sortBySignature);
+    expectedSignatures.error.sort(sortBySignature);
+    res.body.signatures.function.sort(sortBySignature);
+    res.body.signatures.event.sort(sortBySignature);
+    res.body.signatures.error.sort(sortBySignature);
+    chai.expect(res.body.signatures).to.deep.equal(expectedSignatures);
+  };
+
+  it("Libraries have been linked manually instead of using compiler settings. Placeholders are replaced with zero addresses", async () => {
```
marcocastignoli, line 140: What if we keep in this file only the tests and move everything above this line in a separate helper file?

manuelwedler:
> Initial comment: the idea about the verifier alliance json files is that we can share them with the other verifiers, so it would be better to keep them in a json format. In the future we could potentially treat them as an external submodule (like we do with the vera database)

Agreed. I mainly moved to .ts files to get better typing, but the sharing aspect makes it necessary to got back to a json structure. Also, I found that if we want to generate test files via a script, .ts files are not the greatest.

manuelwedler:
Okay I moved to JSON testdata files. Be aware that this prevents from running type checks against the testdata.

kuzdogan:
Beautiful <3

PR #2635, opened 2026-02-09
Add test case generator script for verification cases test suite

manuelwedler:
See #2049


manuelwedler:
> Can't we also verify the contract and extract the missing information? We should of course double check that those info are correct, but I think it makes sense to fill them in advance
> 
> * verification.creationMatch ("exact_match", or "match")
> * verification.runtimeMatch
> * verification.creationTransformations
> * verification.creationValues
> * verification.runtimeTransformations
> * verification.runtimeValues
> * output.creationCodeArtifacts.cborAuxdata
> * output.runtimeCodeArtifacts.cborAuxdata

This means we would generate testdata via the code which we want to test, which makes the testdata useless IMO. I still added the option to generate this data, but asking the user to manually verify the testdata.

Issue #2284, opened 2025-08-07
Rethinking the chains.json and RPCs


kuzdogan:
While doing some housekeeping on RPCs #2283 I felt we should move away from public RPCs and instead go for a stable paid one for main chains. We can check how much it would cost and I assume it won't be an outrageous amount. If we don't request much, then fine we just pay couple bucks, but if we need it for some bulk verification etc. (like when we've been using the `/replace-contract` for #1990) then we are in general ok to pay a little and get this task done without rate limits. Still we should do some estimations and keep a track of how much we are using, and see the worst-case costs.

This also led me to think if we should move away from having a huge sourcify-chains.json config. This has couple facets:

1. Create a Provider module for RPCs we use and use them for all chains supported.

E.g. if we are using Quicknode, let's not handpick and handwrite Quicknode under each chain in the JSON file. We can have a QuicknodeProvider module that will programmatically check which chains are supported (hopefully they have such a spec) and add itself as an RPC for that SourcifyChain. With that we don't have to manually keep track of newly launched chains or Quicknode adding support to new chains.

2. Programmatically support `fetchCreationTxHashUsing`
Since all block explorers we use for creation tx hash fetching have machine readable support JSONs, we can also use this to remove this spec from every single chain.

3. Potentially, support all chains?
And finally we can actually support all chains automatically. Just load up everything from chains.json (or even dynamically from chainlist.org/chains.json) -> check if they have Quicknode etc. support -> Check if they have block explorer support and add them as supported. 

I am actually more skeptical about the 3rd option but putting it out there as a thought exercise

kuzdogan:
- For the chain tests, we can just use the [CreateX](https://createx.rocks/deployments) and [Multicall3](https://www.multicall3.com/deployments) deployments as a pre-requisite for a chain to be automatically supported.
- How do we update the automatic list?
  - Consensus is it should at least restart the instance to update the chain support list.
  - Static file on build time or dynamically fetching on instance start.
  - Consider, to be able to support a chain we need to make sure the chain support tests on the CI pass.
  - The CI can run nightly and create an updated list and trigger restarting all of the instances
  - The tests can be just flaky etc. and we wouldn't want those tests to turn off support for chains randomly. Turning on should be automatic but turning off should be manual. 
- Even if we do an automatic/dynamic update it should still allow existing users and self-hostings shoudl be able to pass a simple chains.json still
- We can check Quicknode and Alchemy supported chains dynamically as well to use those RPCs
[...]

PR #2648, opened 2026-02-13
Refactor chain tests: use CreateX, Multicall3, and Storage only, use v2 API, concurrency

kuzdogan:
## Summary

Part of #2284 

- Replace ~200 hardcoded `verifyContract()` calls with a dynamic loop over all supported chains from `sourcify-chains-default.json`, using only 3 shared contracts (CreateX, Multicall3, Storage.sol)
- Switch from the deprecated legacy `POST /` API to the v2 async verification flow (`POST /v2/verify` + polling `GET /v2/verify/{id}`)
- Replace per-chain source directories (~39k lines of duplicated Solidity) with 3 compact standard JSON input files
- Run up to 20 verifications concurrently using a concurrency pool, instead of sequentially — tests now run in under 3 minutes total
- Surface the server's error messages in test failures instead of unhelpful "expected null to not be null"

## How it works

Each supported chain is verified using one of three contracts in priority order:
1. **CreateX** (`0xba5Ed...`) — deployment list fetched dynamically from GitHub
2. **Multicall3** (`0xcA11...`) — deployment list fetched dynamically from GitHub
3. **Storage.sol** — per-chain addresses in `storage-contract-chain-addresses.json`

All verifications run in the `before()` hook with a concurrency pool (max 20 in-flight). Each verification submits via `POST /v2/verify/{chainId}/{address}` and polls `GET /v2/verify/{verificationId}` until complete. Results are stored in a Map, and each `it()` block just asserts the pre-computed result.

## Chains disabled

4 chains with unique source contracts (no CreateX/Multicall3/Storage.sol deployed) are set to `supported: false`:
- UPTN (6119)
- Beam Testnet (13337)
- Beam (4337)
- Shrapnel (2044)

## Test plan

- [ ] `cd services/server && npx tsc --noEmit` passes
- [ ] `npm run test-local:chains` runs successfully
- [ ] CI chain tests pass

# Todos
- [ ] Update adding chain support docs

The tests have been failing for a number of chains for me locally but I'm assuming it's because of their RPC flakiness. They are mostly non-popular chains.

### DIFF
--- a/services/server/test/chains/chain-tests.spec.ts
+++ b/services/server/test/chains/chain-tests.spec.ts
@@ -124,9 +124,9 @@ describe("Test Supported Chains", function () {
         }
       })(); // run the task immediately
 
-      const tracked = task.then(() => {
+      const tracked = task.finally(() => {
         pending.delete(tracked);
-      }); // add .then() to delete the task from the pending set when it completes
+      }); // add .finally() to delete the task from the pending set when it completes
       pending.add(tracked);
 
       if (pending.size >= CONCURRENCY) {


services/server/test/chains/chain-tests.spec.ts
```diff
@@ -2138,84 +224,47 @@ describe("Test Supported Chains", function () {
...
-
-      chai
+    // Poll until the verification job completes
+    const verificationId = verifyRes.body.verificationId;
+    const maxPolls = Math.floor(parseInt(TEST_TIME) / POLL_INTERVAL);
+    let jobRes;
+    let polls = 0;
+    do {
+      await new Promise((resolve) => setTimeout(resolve, POLL_INTERVAL));
+      jobRes = await chai
```
manuelwedler, line 262: Why don't you use `hookIntoVerificationWorkerRun`?

manuelwedler: due to the concurrency pool?

services/server/test/chains/chain-tests.spec.ts
```diff
@@ -41,22 +82,59 @@ describe("Test Supported Chains", function () {
...
-        supportedChains = res.body.filter(
-          (chain: ChainApiResponse) => chain.supported,
-        );
-      });
+      })(); // run the task immediately
+
+      const tracked = task.then(() => {
+        pending.delete(tracked);
+      }); // add .then() to delete the task from the pending set when it completes
+      pending.add(tracked);
```
manuelwedler:
```suggestion
      const tracked = task.finally(() => {
        pending.delete(tracked);
      }); // add .then() to delete the task from the pending set when it completes
      pending.add(tracked);
```

`finally` seems more appropriate when cleaning up

Issue #2393, opened 2025-09-25
Remove the session endpoints


kuzdogan:
Since we replaced the verify UI with https://verify.sourcify.dev we can fully remove the session endpoints and related code parts including the migrations, db tables

PR #2652, opened 2026-02-17
Support all Solidity versions and drop the metadata dependency

marcocastignoli:
This PR closes https://github.com/argotorg/sourcify/issues/2296 and https://github.com/argotorg/sourcify/issues/2525.

In this PR:
- Fixed the behavior of solcjs in worker in debugger execution
- Always run solcjs in the debugger (not only for < 0.4.0)
- Support all solidity versions
  - Support Solidity < 0.1.3 with no compiled runtime byteode in output
  - Support Solidity < 0.4.9 with no source path key in output
  - Support Solidity < 0.3.6 with no `id` in compiler output `sources`
- Support verification without metadata (remove generated metadata from Vyper and Yul). Remove support of Vyper from `/session/etherscan` API
- Add database migration to support old solidity compiler output and null metadata
- Move RepositoryV2 to `writeOrWarn` in test server config
  - Remove check of repository stored data in tests (in test we now only assert for the database data, we don't check the fs repository data anymore)
- Skip storing matches without metadata in RepositoryV1 / RepositoryV2 / S3Repository
- Fix binary search timeout: successful binary search executions were not canceling the timeout "Binary search for contract creation tx timed out ...", so every successful binary search execution was logging that warning after 120s.
- Add verification cases for Solidity versions:
  - 0.1.1
  - 0.1.3
  - 0.1.6
  - 0.3.6
  - 0.4.0
  - 0.4.7
  - 0.4.9
- Removed metadata from Vyper and Yul verification cases

### DIFF
--- a/services/server/test/integration/apiv1/repository-handlers/files.spec.ts
+++ b/services/server/test/integration/apiv1/repository-handlers/files.spec.ts
@@ -12,6 +12,8 @@ import { ServerFixture } from "../../../helpers/ServerFixture";
 import path from "path";
 import fs from "fs";
 import { RWStorageIdentifiers } from "../../../../src/server/services/storageServices/identifiers";
+import { id as keccak256 } from "ethers";
+import contractVyperArtifact from "../../../sources/vyper/testcontract/artifact.json";
 
 chai.use(chaiHttp);
 
@@ -667,5 +669,86 @@ describe("Verify repository endpoints", function () {
           );
       });
     }
+  });
+
+  it("should generate metadata.json on the fly for Vyper contracts stored without metadata", async function () {
+    // Vyper contracts no longer have metadata stored in the DB (generateMetadata() was removed).
+    // SourcifyDatabaseService.getFiles() must regenerate it via generateMetadataOnTheFly().
+    const contractFileName = "test.vy";
+    const contractName = contractFileName.split(".")[0];
+    const compilerVersion = "0.3.10+commit.91361694";
+    const evmVersion = "istanbul";
+    const contractSourcePath = path.join(
+      __dirname,
+      "..",
+      "..",
+      "..",
+      "sources",
+      "vyper",
+      "testcontract",
+      contractFileName,
+    );
+    const contractFileContent = fs.readFileSync(contractSourcePath, "utf8");
+
+    const contractAddress = await deployFromAbiAndBytecode(
+      chainFixture.localSigner,
+      contractVyperArtifact.abi,
+      contractVyperArtifact.bytecode,
+    );
+
+    await chai
+      .request(serverFixtureWithDatabase.server.app)
+      .post("/verify/vyper")
+      .send({
+        address: contractAddress,
+        chain: chainFixture.chainId,
+        files: { [contractFileName]: contractFileContent },
+        contractPath: contractFileName,
+        contractName,
+        compilerVersion,
+        compilerSettings: {
+          evmVersion,
+          outputSelection: { "*": ["evm.bytecode"] },
+        },
+      });
+
+    const res = await chai
+      .request(serverFixtureWithDatabase.server.app)
+      .get(`/files/any/${chainFixture.chainId}/${contractAddress}`);
+
+    chai.expect(res.status).to.equal(200);
+    const metadataFile = res.body.files.find(
+      (f: { name: string }) => f.name === "metadata.json",
+    );
+    chai.expect(metadataFile, "metadata.json should be present").to.exist;
+
+    const metadata = JSON.parse(metadataFile.content);
+    chai.expect(metadata).to.deep.equal({
+      compiler: { version: compilerVersion },
+      language: "Vyper",
+      output: {
+        abi: [
+          {
+            inputs: [],
+            name: "helloWorld",
+            outputs: [{ name: "", type: "string" }],
+            stateMutability: "view",
+            type: "function",
+          },
+        ],
+        devdoc: {},
+        userdoc: {},
+      },
+      settings: {
+        evmVersion,
+        compilationTarget: { [contractFileName]: contractName },
+      },
+      sources: {
+        [contractFileName]: {
+          keccak256: keccak256(contractFileContent),
+        },
+      },
+      version: 1,
+    });
   });
 });


services/server/src/server/services/utils/contract-creation-util.ts
```diff
@@ -574,8 +574,9 @@ export async function findContractCreationTxByBinarySearchWithTimeout(
   contractAddress: string,
   binarySearchTimeoutMs = BINARY_SEARCH_TIMEOUT_MS,
 ): Promise<string | null> {
-  const timeoutPromise = new Promise<null>((resolve) =>
-    setTimeout(() => {
+  let timeoutId: ReturnType<typeof setTimeout> | undefined;
+  const timeoutPromise = new Promise<null>((resolve) => {
+    timeoutId = setTimeout(() => {
```
marcocastignoli, line 579: Here we forgot to clear this timeout in case of a success

packages/lib-sourcify/src/Compilation/SolidityCompilation.ts
```diff
@@ -270,7 +263,11 @@ export class SolidityCompilation extends AbstractCompilation {
   public async compile(forceEmscripten = false) {
     const contract =
       await this.compileAndReturnCompilationTarget(forceEmscripten);
-    this._metadata = JSON.parse(contract.metadata.trim());
+    try {
+      this._metadata = JSON.parse(contract.metadata.trim());
+    } catch (error: any) {
```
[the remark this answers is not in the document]
marcocastignoli: Not needed.

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -609,13 +615,26 @@ export class Verification {
...
+        for (const source of Object.keys(
+          this.compilation.compilerOutput.sources,
+        )) {
+          compilerOutputSources[source] = {
+            id:
+              typeof this.compilation.compilerOutput.sources[source].id ===
+              'number'
+                ? this.compilation.compilerOutput.sources[source].id
+                : // In older solidity versions, source ids were strings, so we parse them to numbers
+                  parseInt(this.compilation.compilerOutput.sources[source].id),
```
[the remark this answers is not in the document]
marcocastignoli: Implemented

services/server/src/server/apiv1/verification/etherscan/session/etherscan.session.handlers.ts
```diff
@@ -46,6 +47,12 @@ export async function sessionVerifyFromEtherscan(req: Request, res: Response) {
     apiKey,
   );
 
+  if (EtherscanUtils.isVyperResult(etherscanResult)) {
+    throw new BadRequestError(
+      "Etherscan Vyper verification results are not supported in the session-based verification flow.",
```
manuelwedler: If you agree, I suggest to remove the session API soon. No one should be using it anymore. Should be done in another PR.

services/database/migrations/20260216165100_nullable_metadata.sql
```diff
@@ -0,0 +1,39 @@
...
+        is_jsonb_object(obj) AND
+        validate_json_object_keys(
+            obj,
+            array ['abi', 'sources'],
+            array ['userdoc', 'devdoc', 'storageLayout']
+        ) AND
+        validate_compilation_artifacts_abi(obj -> 'abi') AND
+        validate_compilation_artifacts_sources(obj -> 'sources');
+END;
+$$;
```
manuelwedler: Do we want to allow older Solidity versions in VerA too? I'd say yes. For that, we should move this to a VerA migration.

packages/lib-sourcify/src/Compilation/SolidityCompilation.ts
```diff
@@ -270,7 +263,11 @@ export class SolidityCompilation extends AbstractCompilation {
   public async compile(forceEmscripten = false) {
     const contract =
       await this.compileAndReturnCompilationTarget(forceEmscripten);
-    this._metadata = JSON.parse(contract.metadata.trim());
+    try {
+      this._metadata = JSON.parse(contract.metadata.trim());
```
manuelwedler, line 273: Should we better change the type of `SolidityOutputContract.metadata` and check for undefined here?

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -102,7 +103,9 @@ export class Verification {
     const compiledRuntimeBytecode = this.compilation.runtimeBytecode;
     const compiledCreationBytecode = this.compilation.creationBytecode;
 
-    if (compiledRuntimeBytecode === '0x' || compiledCreationBytecode === '0x') {
+    // If compiled creation bytecode is empty it means that the contract is an abstract contract
+    // Compiled runtime bytecode is empty in < 0.1.3 Solidity versions, but we can still verify the contract with creation bytecode
+    if (compiledCreationBytecode === '0x') {
```
manuelwedler: Should we be more specific here, and still check for no compiledRuntimeBytecode when version >= 0.1.3 ?

services/server/test/helpers/assertions.ts
```diff
@@ -103,12 +97,10 @@ export const assertVerificationSession = async (
     chai.expect(contract.files.invalid).to.be.empty;
 
     await assertContractSaved(
-      serverFixture?.sourcifyDatabase ?? null,
+      serverFixture?.sourcifyDatabase,
```
manuelwedler:
```suggestion
      serverFixture.sourcifyDatabase,
```

services/server/test/helpers/assertions.ts
```diff
@@ -171,82 +163,15 @@ export async function assertTransformations(
...
-        fs.existsSync(partialMatchMetadataPath),
-        "Partial match should not exist",
-      ).to.be.false;
-    }
-
-    const expectedMetadataContent = fs.readFileSync(metadataPath).toString();
-    const expectedMetadataHash = id(expectedMetadataContent);
-
-    // Check if saved to S3
-    if (testS3Path && testS3Bucket) {
```
manuelwedler, line 169: Why do we remove the assertions for storing at S3?

packages/lib-sourcify/src/Verification/Verification.ts
manuelwedler: Actually, we said in https://github.com/argotorg/sourcify/issues/2296 that we don't want to support solc < 0.1.3 because it lacks deployed bytecode output. I still think it makes sense to draw the line here.

services/server/package.json
```diff
@@ -23,7 +23,7 @@
     "test-local": "export DOCKER_HOST_POSTGRES_TEST_PORT=${DOCKER_HOST_POSTGRES_TEST_PORT:-5431} && npm run postgres-test:start && sleep 2 && SOURCIFY_POSTGRES_HOST=localhost npm run test; status=$?; npm run postgres-test:stop; exit $status",
     "test-local:chains": "export DOCKER_HOST_POSTGRES_TEST_PORT=${DOCKER_HOST_POSTGRES_TEST_PORT:-5431} && npm run postgres-test:start && sleep 2 && SOURCIFY_POSTGRES_HOST=localhost npm run test:chains; status=$?; npm run postgres-test:stop; exit $status",
     "test-local:etherscan-instances": "export DOCKER_HOST_POSTGRES_TEST_PORT=${DOCKER_HOST_POSTGRES_TEST_PORT:-5431} && npm run postgres-test:start && sleep 2 && SOURCIFY_POSTGRES_HOST=localhost npm run test:etherscan-instances; status=$?; npm run postgres-test:stop; exit $status",
-    "generate-test-case": "ts-node test/integration/verification-cases/generate-test-case.ts",
+    "generate-test-case": "NODE_OPTIONS='--disable-warning=V8' TS_NODE_IGNORE='(^|/)(node_modules|dist|build)/|soljson-v.*[.]js$' ts-node test/integration/verification-cases/generate-test-case.ts",
```
manuelwedler: Are we sure we need this publicly?

services/server/test/integration/verification-cases/assertions.ts
```diff
@@ -272,7 +273,7 @@ export async function assertDatabase(
     .expect(row.runtime_transformations)
     .to.deep.equal(testCase.verification.runtimeTransformations);
   chai
-    .expect(row.runtime_metadata_match)
+    .expect(row.runtime_metadata_match ? true : false)
```
manuelwedler: In case we decide to not support 0.1.1, we don't need this I suppose

services/server/test/integration/verification-cases/assertions.ts
```diff
@@ -347,9 +348,15 @@ export async function assertApiV2Lookup(
   chai
     .expect(res.body.runtimeBytecode.onchainBytecode)
     .to.equal(testCase.onchain.deployedBytecode);
-  chai
-    .expect(res.body.runtimeBytecode.recompiledBytecode)
-    .to.equal(testCase.output.deployedBytecode);
+  // In Solidity < 0.1.3 the compiled_runtime_bytecode is not available in the compilation artifacts
```
manuelwedler: same here

services/server/test/integration/verification-cases/generate-test-case.ts
```diff
@@ -198,7 +202,15 @@ function mapToTestCaseOutput(
...
 ) {
   // Extract metadata
-  const metadata: Metadata = JSON.parse(contractOutput.metadata);
+  let metadata: Metadata | undefined = undefined;
+
+  if (contractOutput.metadata) {
+    try {
+      metadata = JSON.parse(contractOutput.metadata);
+    } catch (e) {
+      console.warn("Failed to parse metadata JSON:", e);
```
manuelwedler: we should fail in this case. this wouldn't lead to valid testdata.

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -609,13 +615,29 @@ export class Verification {
...
+          compilerOutputSources[source] = {
+            id:
+              typeof this.compilation.compilerOutput.sources[source].id ===
+              'number'
+                ? this.compilation.compilerOutput.sources[source].id
+                : // In older solidity versions, source ids were strings, so we parse them to numbers
+                  parseInt(
+                    this.compilation.compilerOutput.sources[source].id,
+                    10,
+                  ),
```
manuelwedler: Do we need to document somewhere, that we store a modified version of the compiler output in the db for this Solidity version?

services/server/test/helpers/assertions.ts
```diff
@@ -171,82 +163,15 @@ export async function assertTransformations(
...
   expectedAddress: string | undefined,
   expectedChain: string | undefined,
   expectedStatus: VerificationStatus,
-  testS3Path?: string | null,
-  testS3Bucket?: string | null,
 ) {
   if (expectedStatus === "perfect" || expectedStatus === "partial") {
-    // Check if saved to fs repository
-    const match = expectedStatus === "perfect" ? "full_match" : "partial_match";
-    const getMetadataPath = (match: string) =>
```
manuelwedler, line 142: I don't get the changes here. I earlier raised that we should keep tests against the repository until we fully removed the RepositoryServices code. AFAIK this `assertContractSaved` function is only run in old apiv1 tests where we don't test the pre 0.4.11 Solidity versions. So I don't understand why it is necessary to remove these checks here.

services/server/src/server/apiv1/verification/etherscan/session/etherscan.session.handlers.ts
kuzdogan: Agreed, I thought we removed it

Priorotized #2393

packages/lib-sourcify/src/Compilation/SolidityCompilation.ts
kuzdogan: Yes agreed. Assuming the metadata returns undefined for old versions. Other errors should propogate

packages/lib-sourcify/src/Verification/Verification.ts
kuzdogan: Agreed. We really don't need to support these versions. It will be very confusing if we make runtime optional too.

services/server/test/helpers/assertions.ts
kuzdogan: Are you sure? Here it's checking verification of a job

https://github.com/argotorg/sourcify/blob/72d085be80e5b2987c73d7b69dbc778fd4d0ff3f/services/server/test/helpers/assertions.ts#L291-L299

kuzdogan: I think it was necessary because there's no metadata to save in the filesystem. I guess what we can do is if the contract metadata is undefined we don't check the metadata file and that should be enough.

services/server/test/helpers/assertions.ts
kuzdogan: Here also we should just skip the metadata.json check if it's undefined and keep the rest

services/server/test/integration/verification-cases/generate-test-case.ts
kuzdogan: Yes if there's a metadata it must be valid and the Error should propagate.

packages/lib-sourcify/src/Verification/Verification.ts
kuzdogan: Wait does this mean we don't store the sources at all for below this version? If so it didn't seem acceptable to me

services/server/test/helpers/assertions.ts
manuelwedler: right, I missed that. yeah skipping the metadata for this case should be enough

packages/lib-sourcify/src/Verification/Verification.ts
manuelwedler: No, it's just the source ids. The ids used to be strings but changed to be numbers in some solc version. Here we enforce them to be numbers to be compliant with some database constraint I assume. This just means we don't store the exact same format for the source ids as they are produced by the older solidity version.

kuzdogan: No I mean above at line 623, it's set to `undefined`

manuelwedler: `compilerOutputSources = undefined;` is for the source ids. Seems still fine to me.

packages/compilers/src/lib/solidityCompiler.ts
```diff
@@ -58,6 +58,16 @@ export async function useSolidityCompiler(
...
+      // Early catch for unsupported versions and platforms
+      if (
+        (solcPlatform === 'linux-amd64' && semver.lt(version, '0.4.10')) ||
+        (solcPlatform === 'macosx-amd64' && semver.lt(version, '0.3.6')) ||
+        (solcPlatform === 'windows-amd64' && semver.lt(version, '0.4.1'))
+      ) {
+        throw new Error(
+          `No solc binary available for version ${version} and platform ${solcPlatform}`,
+        );
+      }
```
kuzdogan, line 70: Shouldn't we switch to solcjs instead of throwing here?

packages/compilers/src/lib/solidityCompiler.ts
```diff
@@ -86,32 +96,21 @@ export async function useSolidityCompiler(
...
-          });
-        });
-      } else {
-        compiled = solJson.compile(inputStringified);
-      }
-    }
+    compiled = await new Promise((resolve, reject) => {
+      const worker = importWorker(path.resolve(__dirname, './compilerWorker'), {
+        workerData: { solJsonRepoPath, version, inputStringified },
+      });
```
kuzdogan, line 106: I think we should spawn a clean context only for <0.4.0?, as the comment suggests

packages/compilers/src/lib/solidityCompiler.ts
```diff
@@ -86,32 +96,21 @@ export async function useSolidityCompiler(
       throw error;
     }
   } else {
-    logDebug('Compiling with solc-js', { version });
-    const solJson = await getSolcJs(solJsonRepoPath, version);
```
kuzdogan, line 90: Ok to remove this? How does the worker know if the binary is there?

services/database/sourcify-database.sql
```diff
@@ -406,8 +426,8 @@ BEGIN
         is_jsonb_object(obj) AND
         validate_json_object_keys(
             obj,
-            array ['abi', 'userdoc', 'devdoc', 'sources', 'storageLayout'],
-            array []::text[]
+            array ['abi', 'sources'],
+            array ['userdoc', 'devdoc', 'storageLayout', 'transientStorageLayout']
```
kuzdogan, line 430: Why are these changes for transientStorageLayout here?

services/database/sourcify-database.sql
```diff
@@ -220,6 +220,26 @@ END;
...
+BEGIN
+    RETURN obj IS NULL OR (
+        is_jsonb_object(obj) AND
+        validate_json_object_keys(
+            obj,
+            array []::text[],
+            array ['storage_layout_overrides']
+        )
+    );
+END;
```
kuzdogan, line 239: Additional Input changes are here?

services/server/src/config/local-test.js
```diff
@@ -15,11 +15,9 @@ module.exports = {
     read: RWStorageIdentifiers.SourcifyDatabase,
     writeOrWarn: [
       RWStorageIdentifiers.RepositoryV1,
-      WStorageIdentifiers.S3Repository,
-    ],
-    writeOrErr: [
       WStorageIdentifiers.RepositoryV2,
```
kuzdogan, line 18: Are we moving repo to Warn in this PR?

services/server/test/helpers/assertions.ts
```diff
@@ -125,14 +127,18 @@ export async function assertTransformations(
...
   expectedChain: string | undefined,
   expectedStatus: VerificationStatus,
   testS3Path?: string | null,
   testS3Bucket?: string | null,
+  metadataExpected: boolean = true,
 ) {
-  if (expectedStatus === "perfect" || expectedStatus === "partial") {
+  if (
+    (expectedStatus === "perfect" || expectedStatus === "partial") &&
+    metadataExpected
```
kuzdogan, line 145: We are skipping the whole check if the metadata is not expected? In the prev. discussion we said we'll only skip the metadata related assertions?

services/server/test/integration/verification-cases/assertions.ts
```diff
@@ -18,6 +18,7 @@ import {
 } from "../../../src/server/services/utils/util";
 import type { VerificationTestCase } from "./verification-cases.spec";
 import type { DeploymentInfo } from "../../helpers/helpers";
+import semver from "semver";
```
kuzdogan: Seems unused

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
```diff
@@ -103,49 +102,6 @@ export class VyperCompilation extends AbstractCompilation {
   // Vyper version is not semver compliant, so we need to handle it differently
   public compilerVersionCompatibleWithSemver: string;
 
-  /**
-   * Vyper compiler does not produce a metadata but we generate it ourselves for backward
-   * compatibility reasons e.g. in the legacy Sourcify API that always assumes a metadata.json
-   */
-  generateMetadata() {
```
kuzdogan, line 110: We are removing metadata generation fully as we are not going to generate + write. But I just notice this is a breaking change to the API. This means there's no metadata.json files at all and everything on the legacy API depends on metadata.json.

Now, it's the legacy API and only the future contracts will be affected. It's actually fine and I don't have a strong opinion on this. 

But to avoid the breaking change, should we move this logic to server and generate the metadata.json on the fly if it doesn't exist? It would be at the SourcifyDatabaseService:326 as an else clause. The logic is quite simple. 

Our filesystem endpoints are already using the DB as the source of truth, so this is already in line with this.

services/server/test/helpers/assertions.ts
```diff
@@ -213,32 +218,25 @@ export async function assertContractSaved(
...
+      [Buffer.from(expectedAddress?.substring(2) ?? "", "hex"), expectedChain],
+    );
 
-      const contract = res.rows[0];
-      chai.expect(contract).to.not.be.null;
-      chai
-        .expect("0x" + contract.address.toString("hex"))
-        .to.equal(expectedAddress?.toLowerCase());
-      chai.expect(contract.chain_id).to.equal(expectedChain);
-      chai
```
kuzdogan, line 232: This code block assumes metadata is here. Why is this line checking hashes removed?

Edit: Look lines below actually 228-230

packages/compilers/src/lib/solidityCompiler.ts
manuelwedler: We already do. The error is caught, and since `solcPath` will be undefined the solc-js block is entered later in the function.

services/database/sourcify-database.sql
manuelwedler: I updated the database-specs submodule to get the `20260224135405_support_old_solc_output.sql` migration in here. The `transientStorageLayout` migration was in the same PR, so it's not possible to exclude it.

services/database/sourcify-database.sql
manuelwedler: same due to updating the database-specs

services/server/src/config/local-test.js
manuelwedler: This is just for the tests. Actually it reflects what we have on production too.

services/server/test/helpers/assertions.ts
manuelwedler: Ah my bad. I wanted the if block to end before the database assertions. Thanks for catching this.

services/server/test/helpers/assertions.ts
manuelwedler: I'm bringing it back

packages/compilers/src/lib/solidityCompiler.ts
manuelwedler: true, this might be an issue

packages/compilers/src/lib/solidityCompiler.ts
manuelwedler: I think Marco had some issue with other versions too, and I said it would be fine to always use a worker. I don't remember the exact reason but I think it's related to these bullets from the description:

> Fixed the behavior of solcjs in worker in debugger execution
> Always run solcjs in the debugger (not only for < 0.4.0)

I think I will still try to bring the previous logic back

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
manuelwedler: Very good catch. Totally agreed. We should generate the metadata on the fly for API v1.

packages/compilers/src/lib/solidityCompiler.ts
manuelwedler: Tests fail with it. I think we should just spawn the worker in all cases.

packages/compilers/src/lib/solidityCompiler.ts
manuelwedler: I tried to add a check before running the worker, but the tests were failing after this commit. I  couldn't understand why. But I doubled checked the logic of the worker, and actually the worker will return in an error if an invalid solidity version is passed. The worker itself downloads solc js.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -282,6 +286,90 @@ export class SourcifyDatabaseService
     return { pagination, results };
   };
 
+  /**
+   * Generates a metadata.json on the fly for contracts that don't have one stored
+   * (e.g. Vyper and Yul contracts verified after metadata generation was removed).
+   * Replicates the format previously produced by VyperCompilation.generateMetadata()
+   * and YulCompilation.generateMetadata().
+   */
+  private async generateMetadataOnTheFly(
```
kuzdogan, line 295: I don't want to drag this PR any further but this function really needs testing. Either a unit test or a integration test that will also test lines 416-424

packages/compilers/src/lib/solidityCompiler.ts
kuzdogan: I think the changes you made were missing the <0.4.0 check and just spawn a new worker all the time in https://github.com/argotorg/sourcify/pull/2652/changes/e43417dc93ccd3870894ead5e6436ec6f4a32915 You don't have `solJson.compile()` there.

image

Check out the old code. It does:

1. Try fetching solcJs
2. Check if it's there
3. Check if version is less than 0.4.0
4. If yes spawn a clean worker
5. If not just call solJson.compile()

If you think this works it's not a block from my side but I don't understand why we are touching this part of the code in the first place. The original code should've just been working?

manuelwedler: I already tried rolling back to the previous logic. See commit https://github.com/argotorg/sourcify/commit/baa93913e565534764cbb75c5fa81d02c2128136 and https://github.com/argotorg/sourcify/pull/2652#discussion_r2906482445.

> If you think this works it's not a block from my side but I don't understand why we are touching this part of the code in the first place. The original code should've just been working?

I also don't understand why it is necessary, but I remember that Marco had a reason for it and spent sometime debugging with coming to this solution in the end. I feel like it's not worth spending the time again to understand it.

PR #2664, opened 2026-02-20
feat(server): dynamic chain config via provider/explorer discovery

kuzdogan:
WIP: Replace static sourcify-chains-default.json with a build-time generation system using QuickNode, dRPC, Etherscan, Blockscout, and Routescan.


kuzdogan:
Closing in favor of a different approach: chain config will be fetched from an external sourcifyeth/sourcify-chains repo at startup.

PR #2671, opened 2026-02-26
Remove /session API endpoints

### LINKED ISSUE #2393: Remove the session endpoints

manuelwedler:
Closes #2393

they were deprecated with the introduction of API v2 and are not used anywhere.

Note: I moved two test cases for factory contracts over to the stateless API v1 tests because these cases were not covered elsewhere and it was quick to do move them there. Eventually we should move them to the verification cases suite (https://github.com/argotorg/sourcify/issues/2640).

### DIFF
--- a/services/database/migrations/20260302082853_drop_session_table.sql
+++ b/services/database/migrations/20260302082853_drop_session_table.sql
@@ -1,0 +1,17 @@
+-- migrate:up
+
+/*
+  The session API is removed, and the table no longer needed.
+*/
+
+DROP TABLE IF EXISTS "session";
+
+-- migrate:down
+
+CREATE TABLE "session" (
+    "sid" varchar NOT NULL COLLATE "default",
+    "sess" json NOT NULL,
+    "expire" timestamp(6) NOT NULL
+) WITH (OIDS=FALSE);
+ALTER TABLE "session" ADD CONSTRAINT "session_pkey" PRIMARY KEY ("sid") NOT DEFERRABLE INITIALLY IMMEDIATE;
+CREATE INDEX "IDX_session_expire" ON "session" ("expire");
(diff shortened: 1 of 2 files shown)


services/server/package.json
```diff
@@ -60,15 +60,13 @@
...
     "directory-tree": "3.6.0",
     "dotenv": "16.6.1",
     "ethers": "6.16.0",
     "express": "5.2.1",
     "express-fileupload": "1.5.2",
     "express-openapi-validator": "5.6.2",
-    "express-session": "1.19.0",
     "http-status-codes": "2.3.0",
     "json-refs": "3.0.15",
     "memorystore": "1.6.7",
```
kuzdogan: `memorystore` was only used by `initMemoryStore()` in `cli.ts`, which was deleted. The package itself was missed.

```suggestion
    "json-refs": "3.0.15",
```

services/server/src/server/common.ts
```diff
@@ -37,15 +37,3 @@ export type ContractWrapperData = {
 export type ContractWrapper = ContractMeta & {
   contract: ContractWrapperData;
 };
```
kuzdogan: With `ContractWrapperMap` and the `express-session` augmentation now removed, none of the remaining types in this file (`PathContentMap`, `ContractMeta`, `ContractWrapperData`, `ContractWrapper`) are imported anywhere in the codebase.

The entire file is now dead code and can be deleted.

```suggestion
```

manuelwedler:
> Let's remember to delete the session tables. But also do we need to delete them in DB migrations too?

Ah forgot about it. Yes, we should also remove the table via a migration. Added this in the last commit.

PR #2675, opened 2026-03-03
Update release script to set --latest for sourcify-server

manuelwedler:
We don't have the monorepo release anymore, and now I think server should be set as the latest release as our main service.


kuzdogan (approved):
Agreed, thanks

PR #2678, opened 2026-03-04
Add deprecation warning field and Warning HTTP header to legacy v1 API responses

### LINKED ISSUE #2624: Make sure LLMs don't use legacy APIs
Right now LLMs are calling v1 APIs, we should find a way to let LLMs know v1 is deprecated

kuzdogan:
Closes #2624

## Summary

- Adds a `"warning"` field to JSON response bodies of all legacy v1 API endpoints, so AI agents and developers inspecting payloads are immediately prompted to switch to the v2 API
- Adds the RFC 7234 `Warning: 299` HTTP header alongside the existing `Deprecation: true` header across all v1 routes
- Updates OpenAPI schemas to document the new `warning` field in response schemas and examples
- Updates tests to assert the `warning` field is present and correct

## Details

**New `warning` field in JSON responses:**
- `/check-all-by-addresses` — added to each `chainIds[*]` entry and to top-level "not found" address objects (`status: "false"`)
- `/check-by-addresses` — added to each top-level address object (found and not found)
- `/verify`, `/verify/etherscan`, `/verify/solc-json`, `/verify/vyper` — added to each `result[*]` item

Warning messages are endpoint-specific and name the v2 replacement directly:
- Check endpoints: `"... Use GET /v2/contract/{chain}/{address} for each address+chain combination instead."`
- Verify endpoints: `"... Use POST /v2/verify instead."`

**OpenAPI schema fix:** The `check-all-by-addresses` schema now correctly documents the top-level `status` field as only being present (with value `"false"`) when the contract is not found on any of the requested chains. Verified against the live API.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ workspace/
 
 # Sourcify release script temporary file
 .release_package_data.tmp
+.claude/
--- a/services/server/src/server/apiv1/controllers.common.ts
+++ b/services/server/src/server/apiv1/controllers.common.ts
@@ -14,6 +14,16 @@ import type {
 } from "@ethereum-sourcify/lib-sourcify";
 import type { Match } from "../types";
 
+export const CHECK_ENDPOINTS_DEPRECATION_WARNING =
+  "DEPRECATED: This endpoint will be removed. Do not build new integrations against it. " +
+  "Use GET /v2/contract/{chain}/{address} for each address+chain combination instead. " +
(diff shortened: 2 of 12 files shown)


manuelwedler (approved):

MohammedYahyaHusseinAlZalab:
Great addition! Adding these warnings directly in the response body is a very effective way to ensure developers and AI agents catch the deprecation early. This will definitely help smooth the transition to v2."

PR #2684, opened 2026-03-09
fix(server): add index on contract_deployments.chain_id for /v2/contracts pagination

### LINKED ISSUE #2111: `/v2/contracts/{chainId}` endpoint hangs beyond certain `matchId`
A user in our public chat reported an issue where the `/v2/contracts/{chainId}` endpoint hangs after reaching a certain `matchId`.

I investigated and found that running the query below takes a long time, but if you set LIMIT 1, it returns quickly. This slowdown happens because there’s a large gap of non-137 chain contracts before _matchId 4403388_, causing a delay before returning results.

```sql
SELECT
      sourcify_matches.id, sourcify_matches.creation_match, sourcify_matches.runtime_match, nullif(concat('0x', encode(contract_deployments.address, 'hex')), '0x') as address, to_char(sourcify_matches.created_at, 'YYYY-MM-DD"T"HH24:MI:SS"Z"') as verified_at
    FROM public.sourcify_matches
    JOIN public.verified_contracts ON verified_contracts.id = sourcify_matches.verified_contract_id
    JOIN public.contract_deployments ON 
        contract_deployments.id = verified_contracts.deployment_id
    WHERE sourcify_matches.id < 4403388 and contract_deployments.chain_id = 137
    ORDER BY sourcify_matches.id DESC
    LIMIT 2;
```

The user later identified the exact gap:
> seems like the issue is fixed now.  After '1152336', it jumped to '4403507' (quite a big jump), but its working now.⁦

### LINKED ISSUE #2647: Switch pg_dump version to 15 instead of 16 in the CI
In https://github.com/argotorg/sourcify/pull/2644 we noticed we are using a pg_dump version that does not match out Postgres version 15. 

Switch to 15 also in pg_dump in CI runs

kuzdogan:
Fixing both the `/v2/contracts/{chainId}` endpoint issue as well as changing the `pg_dump` version to 15 as we discussed before. I changed the validation step to also exclude pg_cron lines and other whitespaces, comments etc. 

I'll test on staging after approval.

## Summary

- **Fix slow `/v2/contracts/{chainId}` pagination** (fixes #2111): adds a database index on `contract_deployments(chain_id)` and moves the chain filter from the JOIN `ON` clause to `WHERE` so the query planner can use the index independently of the cursor condition
- **Fix CI pg_dump version mismatch** (fixes #2647): switches `postgresql-client-16` → `postgresql-client-15` in the schema validation job to match the `postgres:15-alpine` server image

## Problem (pagination hang)

The `getSourcifyMatchesByChain` query paginates on `sourcify_matches.id` but filters by `chain_id` via a JOIN to `contract_deployments`. PostgreSQL scans `sourcify_matches` in descending ID order and applies the chain filter per row. When there is a large gap of non-matching-chain IDs (e.g. IDs 1,152,336–4,403,388 are all non-chain-137 contracts), it must scan ~3.25M rows to find the next page — causing the endpoint to hang.

See investigation comment: https://github.com/argotorg/sourcify/issues/2111#issuecomment-4021686175

## Problem (CI pg_dump version)

The `validate-database-schema` job was installing `postgresql-client-16` while running against a `postgres:15-alpine` server. pg_dump 16.10+ also emits `\restrict`/`\unrestrict` wrappers with a random key on each run, making schema diffs non-reproducible across machines. Switching to pg_dump 15 matches the server version and avoids the `\restrict` noise entirely.

See: https://github.com/argotorg/sourcify/issues/2647

## Test plan

- [ ] Apply migration and verify `contract_deployments_chain_id` index exists
- [ ] Run `EXPLAIN ANALYZE` on the slow query from the issue and confirm the planner uses the new index
- [ ] Hit `/v2/contracts/137?afterMatchId=4403388` and confirm it returns quickly
- [ ] Verify CI schema validation job passes with pg_dump 15

### DIFF
--- a/.circleci/new_branch.yml
+++ b/.circleci/new_branch.yml
@@ -160,14 +160,11 @@ jobs:
             # Define cleanup function to remove:
             # - SQL comments (lines starting with --)
             # - psql meta-commands (\restrict, \unrestrict)
-            # - pg_cron extension lines (optional extension, not available in all environments)
             # - blank lines (pg_dump formatting varies across OS/distributions)
             clean_sql() {
               sed -e '/^--/d' \
                   -e '/^\\restrict/d' \
                   -e '/^\\unrestrict/d' \
-                  -e '/^CREATE EXTENSION.*pg_cron/d' \
-                  -e '/^COMMENT ON EXTENSION pg_cron/d' \
                   -e '/^$/d' \
                   "$1"
             }


.circleci/new_branch.yml
```diff
@@ -157,9 +157,19 @@ jobs:
...
+            # - psql meta-commands (\restrict, \unrestrict)
+            # - pg_cron extension lines (optional extension, not available in all environments)
+            # - blank lines (pg_dump formatting varies across OS/distributions)
             clean_sql() {
-              sed -e '/^--/d' -e '/^\\restrict/d' -e '/^\\unrestrict/d' "$1"
+              sed -e '/^--/d' \
+                  -e '/^\\restrict/d' \
+                  -e '/^\\unrestrict/d' \
+                  -e '/^CREATE EXTENSION.*pg_cron/d' \
+                  -e '/^COMMENT ON EXTENSION pg_cron/d' \
```
manuelwedler: I think we shouldn't ignore pg_cron lines. We removed pg_cron lines from the dump on purpose. See https://github.com/argotorg/sourcify/issues/2639

manuelwedler:
Just to note. This is not true:

> pg_dump 16.10+ also emits \restrict/\unrestrict wrappers with a random
> key on each run, which causes schema diffs to be non-reproducible across
> machines. Using pg_dump 15 avoids this entirely.

pg_dump 15 also has the restrict feature from minor version 15.14.

.circleci/new_branch.yml
kuzdogan: Why is ignoring not fine?

In that case, have we been manually removing pg_cron lines before we commit? IMO that's error prone.

kuzdogan:
> Just to note. This is not true:
> 
> > pg_dump 16.10+ also emits \restrict/\unrestrict wrappers with a random
> > key on each run, which causes schema diffs to be non-reproducible across
> > machines. Using pg_dump 15 avoids this entirely.
> 
> pg_dump 15 also has the restrict feature from minor version 15.14.

Are we using the restrict in any way? Before we were stripping off those lines

.circleci/new_branch.yml
manuelwedler: I think we should either decide to have it inside the dump or not. I tend to say we should simply exclude it because it is not worth installing if you also don't have the job data. The cron job data won't be exported.

I don't have pg_cron installed locally, and therefore I don't have to remove lines manually.

manuelwedler:
> Are we using the restrict in any way? Before we were stripping off those lines

No, it's a security feature not applicable to our case in my understanding.

Issue #1654, opened 2024-09-23
Adding `transientStorageLayout` to outputs


kuzdogan:
In addition to the `storageLayout` the Solidity compiler also now has a `transientStorageLayout` output. 

We need to look into this and potentially incorporate this into our and Verifier Alliance's database schemas.

kuzdogan:
Related: https://github.com/verifier-alliance/database-specs/issues/31

manuelwedler:
all done

PR #2689, opened 2026-03-12
Add transientStorageLayout to compiler output and store in db

manuelwedler:
See #1654

### DIFF
--- a/services/server/test/integration/verification-cases/generate-test-case.ts
+++ b/services/server/test/integration/verification-cases/generate-test-case.ts
@@ -33,6 +33,7 @@ import type {
 import {
   Verification,
   SolidityCompilation,
+  DEFAULT_OUTPUT_SELECTION,
 } from "@ethereum-sourcify/lib-sourcify";
 import type { VerificationTestCase } from "./verification-cases.spec";
 import { toMatchLevel } from "../../../src/server/services/utils/util";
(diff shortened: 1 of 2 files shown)


services/server/test/integration/verification-cases/generate-test-case.ts
```diff
@@ -396,6 +397,28 @@ async function main() {
     console.log("-".repeat(80));
     console.log("Compiling verification contract...");
 
+    verificationStdJsonInput.settings.outputSelection = {
```
kuzdogan: Instead of duplicating, should we export the outputSelection in packages/lib-sourcify/src/Compilation/SolidityCompilation.ts and import here?

Also if we are adding this here now, how were outputSelections being set before? By the JSON provided when generating the tests?

manuelwedler: sounds good :+1: 

yeah output selection came from the json input.

manuelwedler:
> Also wondering if we should document somehow that `transientStorage` is `null` for unsupported contracts pre 0.8.28 (right?). Is this the same for the `storageLayout` too? Not too important though

We could document it for the other fields too, like metadata, userdoc, devdoc. The versions should be here already: https://github.com/argotorg/sourcify/issues/2296

Just not sure where we should put it.

PR #2692, opened 2026-03-17
feat: add Fe language support for smart contract verification

kuzdogan:
## Summary

- **New language**: Fe (argotorg/fe) — Rust-inspired EVM smart contract language with CLI-only compiler
- **Compiler wrapper** (`feCompiler.ts`): downloads Fe binary per-version, scaffolds ingot project structure (`fe.toml` + `src/`), runs `fe build`, reads `out/*.bin`/`out/*.runtime.bin`
- **`FeCompilation` class**: extends `AbstractCompilation`; `AuxdataStyle.FE` (no CBOR metadata → always partial match); no ABI/immutables/libraries
- **Server pipeline**: `FeLocal` compiler, Piscina worker support, `compilation.ts` and `database-util.ts` Fe cases
- **API middleware** (`validateAndNormalizeFeInput`): normalizes source keys to `src/` prefix, validates `contractIdentifier` format (`src/path.fe:Name` required)
- **OpenAPI spec** updated with Fe language, examples, and `contractIdentifier` format docs
- **Tests**: `FeCompilation.spec.ts` (6 unit tests), `FeVerification.spec.ts` (3 verification tests on local hardhat), `verify.fe.spec.ts` (2 server e2e tests)
- **Pre-commit hook** (husky + lint-staged): runs ESLint and Prettier on all staged files before every commit. Added to ensure AI agents write correctly styled code — a pre-commit hook was chosen over a Claude-specific instruction so the enforcement applies globally to all contributors, not just Claude.

## API usage

Single-file:
```json
POST /v2/verify/{chainId}/{address}
{
  "compilerVersion": "26.0.0-alpha.10",
  "contractIdentifier": "src/lib.fe:Counter",
  "stdJsonInput": {
    "language": "Fe",
    "sources": {
      "src/lib.fe": { "content": "pub contract Counter { ... }" }
    }
  }
}
```

Multi-file ingot (contract in non-`lib.fe` file):
```json
{
  "contractIdentifier": "src/counter.fe:Counter",
  "stdJsonInput": {
    "language": "Fe",
    "sources": {
      "src/lib.fe": { "content": "use ingot::counter::Counter" },
      "src/counter.fe": { "content": "pub contract Counter { ... }" }
    }
  }
}
```

Sources without `src/` prefix are automatically normalized. Match is always **partial** (Fe emits no CBOR metadata hash).

## Test plan

- [ ] `cd packages/lib-sourcify && npm test` — 150 tests passing (includes 6 FeCompilation + 3 FeVerification)
- [ ] `cd services/server && npm run test-local` — full integration suite
- [ ] Manual: POST to `/v2/verify` with a deployed Fe contract

# TODOs
- [x] https://github.com/argotorg/sourcify/issues/2693
- [ ] Add Fe verification docs


kuzdogan:
## How a Fe contract is stored in `compiled_contracts`

Fe is a new compilation type with some workarounds in how it plugs into Sourcify's existing framework (no official std-JSON interface, no CBOR metadata, no ABI). Since this is the first language we add that lacks most of the fields Solidity/Vyper produce, I wanted to highlight exactly how these values get saved in `compiled_contracts`.

**`compiled_contracts` row:**

| Column | Value | Description |
|--------|-------|-------------|
| `language` | `"fe"` | Lowercased from `"Fe"` |
| `compiler` | `"fe"` | Fe CLI compiler |
| `version` | `"26.0.0-alpha.10"` | Compiler version as passed in the request |
| `name` | `"Counter"` | Contract name from `contractIdentifier` |
| `fully_qualified_name` | `"src/lib.fe:Counter"` | Full `path:name` from `contractIdentifier` |
| `compiler_settings` | `{}` | Fe alpha has no configurable settings |
| `compilation_artifacts` | `{"abi": null, "userdoc": null, "devdoc": null, "storageLayout": null, "transientStorageLayout": null, "sources": null}` | Fe does not emit ABI, NatSpec, storage layouts, or source IDs |
| `creation_code_artifacts` | `{"sourceMap": null, "linkReferences": null, "cborAuxdata": {}}` | Fe emits no source maps, has no library linking, and embeds no CBOR metadata |
| `runtime_code_artifacts` | `{"sourceMap": null, "linkReferences": null, "immutableReferences": null, "cborAuxdata": {}}` | Same as creation, plus no immutables |

**Match type:** always `"match"` (partial) for both creation and runtime. Fe embeds no CBOR metadata in its bytecode, so there is no metadata hash to achieve an exact/perfect match.

All of the above is verified by `"should store and return correct Fe-specific fields after verification"` in `services/server/test/integration/apiv2/verification/verify.fe.spec.ts`, which checks every field in this table through two layers:
1. **API response** — `GET /v2/contract/:chainId/:address?fields=compilation,abi,metadata,storageLayout,sources,creationBytecode,runtimeBytecode`
2. **Direct DB query** — raw `compiled_contracts` row via `serverFixture.sourcifyDatabase`

cburgdorf:
Very nice! I'll prioritize a few things so that we can close some gaps (Most importantly the missing JSON ABI)

packages/compilers-types/src/FeTypes.ts
```diff
@@ -0,0 +1,61 @@
...
+  settings?: Record<string, never>;
+}
+
+export interface FeOutputContract {
+  /** ABI is not emitted by `fe build` — null when not available */
+  abi: null;
+  /** Fe does not emit userdoc */
+  userdoc?: undefined;
+  /** Fe does not emit devdoc */
+  devdoc?: undefined;
```
manuelwedler:
```suggestion
  /** Fe does not emit userdoc */
  userdoc?: never;
  /** Fe does not emit devdoc */
  devdoc?: never;
```

or remove these fields

packages/compilers-types/src/FeTypes.ts
```diff
@@ -0,0 +1,61 @@
...
+  evm: {
+    bytecode: {
+      /** Hex string without 0x prefix */
+      object: string;
+    };
+    deployedBytecode: {
+      /** Hex string without 0x prefix */
+      object: string;
+      /** Fe does not emit source maps */
+      sourceMap?: undefined;
```
manuelwedler:
```suggestion
      sourceMap?: never;
```

same

packages/compilers/src/lib/feCompiler.ts
```diff
@@ -0,0 +1,228 @@
+import path from 'path';
```
manuelwedler, line 1: would be good to have some basic tests for `feCompiler` as we do for the other compilers.

packages/bytecode-utils/src/lib/bytecode.ts
```diff
@@ -158,6 +159,11 @@ export const splitAuxdata = (
   validateBytecode(bytecode);
   bytecode = ensureHexPrefix(bytecode);
 
+  // Fe has no CBOR metadata — return the full bytecode with no auxdata
+  if (auxdataStyle === AuxdataStyle.FE) {
+    return [bytecode];
+  }
```
manuelwedler, line 165: small test case for this would also be good

packages/lib-sourcify/src/Verification/VerificationTypes.ts
```diff
@@ -130,7 +130,11 @@ export interface VerificationExport {
     immutableReferences?: ImmutableReferences;
     metadata?: Metadata;
     jsonInput: {
-      settings: SoliditySettings | VyperSettings;
+      settings:
+        | SoliditySettings
+        | VyperSettings
+        | Record<string, never>
```
manuelwedler: Rather add a FeSettings type to compilers-types like this:

```ts
export type FeSettings = Record<string, never>;
```

and use the `FeSettings` type on the VerificationExport

services/server/test/integration/apiv2/verification/verify.fe.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+import { LocalChainFixture } from "../../../helpers/LocalChainFixture";
+import { ServerFixture } from "../../../helpers/ServerFixture";
+import path from "path";
+import fs from "fs";
+import { assertJobVerification } from "../../../helpers/assertions";
+import sinon from "sinon";
+
+chai.use(chaiHttp);
+
+describe("POST /v2/verify/:chainId/:address - Fe contracts", function () {
```
manuelwedler: I think all of this should actually be tested through the new verification-cases test suite. Just don't use the generation script and instead write the json file manually.

packages/lib-sourcify/test/Compilation/FeCompilation.spec.ts
```diff
@@ -0,0 +1,140 @@
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+import path from 'path';
+import fs from 'fs';
+import { FeCompilation } from '../../src/Compilation/FeCompilation';
+import { feCompiler } from '../utils';
+
+describe('FeCompilation', () => {
+  describe('Basic Verification', () => {
```
manuelwedler:
```suggestion
  describe('Basic Compilation', () => {
```

I found this confusing.

packages/lib-sourcify/test/utils.ts
```diff
@@ -345,6 +361,26 @@ export async function deployCompiledContract(
...
+ * Deploys a Fe contract from its creation bytecode.
+ * Fe contracts have no ABI — we deploy using an empty-ABI ContractFactory.
+ * @param creationBytecode - raw hex string WITHOUT 0x prefix (from Fe artifact.json)
+ */
+export async function deployFeContract(
+  signer: JsonRpcSigner,
+  creationBytecode: string,
+): Promise<{ contractAddress: string; txHash: string }> {
+  const factory = new ContractFactory([], creationBytecode, signer);
+  const deployment = await factory.deploy();
```
manuelwedler: nitpicky, but why use an ethers factory for this case if you can just send a transaction with the creation bytecode:

```
  const tx = await signer.sendTransaction({
    data: bytecode,
  });
```

like in server's deployFromBytecodeForCreatorTxHash

packages/lib-sourcify/test/Verification/FeVerification.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+              `Expected SourcifyLibError with code 'no_match' but got '${e.code}'`,
+            );
+          }
+        } else {
+          throw e;
+        }
+      }
+      if (!threw) {
+        throw new Error('Expected verification to throw but it did not');
+      }
```
manuelwedler: This check is overly complicated. There is a `.to.throw` assertion in chai that should be used.

services/server/src/server/apiv2/middlewares.ts
```diff
@@ -224,6 +226,60 @@ export function validateMetadata(
...
+  }
+
+  // If no keys have src/ prefix: add it to all keys
+  if (withSrc.length === 0) {
+    const normalized: Record<string, { content: string }> = {};
+    for (const [k, v] of Object.entries(sources)) {
+      normalized[`src/${k}`] = v;
+    }
+    req.body.stdJsonInput = { ...stdJsonInput, sources: normalized };
+  }
```
manuelwedler, line 258: I'm missing context here. Why do files need to be under `src/`?

services/server/src/server/apiv2/middlewares.ts
```diff
@@ -224,6 +226,60 @@ export function validateMetadata(
...
+  // - Path must start with "src/" and end with ".fe"
+  const ci: string | undefined = req.body.contractIdentifier;
+  if (ci) {
+    const colonIdx = ci.lastIndexOf(":");
+    if (colonIdx === -1) {
+      throw new InvalidParametersError(
+        'For Fe contracts, contractIdentifier must include the source file path, e.g. "src/lib.fe:Counter" or "src/counter.fe:Counter".',
+      );
+    } else {
+      const contractPath = ci.slice(0, colonIdx);
```
manuelwedler: I find the normalization logic a bit weird in general, but if we do it for sources, shouldn't we do it for the contractIdentifier too?

services/server/src/server/services/utils/database-util.ts
```diff
@@ -893,12 +896,19 @@ export function prepareCompilerSettingsFromVerification(
...
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const { outputSelection, ...restSettings } =
-    verification.compilation.jsonInput.settings;
-  return restSettings;
+    verification.compilation.jsonInput.settings ?? {};
+  return restSettings as Omit<
+    SoliditySettings | VyperSettings,
+    "outputSelection"
+  >;
 }
```
manuelwedler: You can add a FeSettings type to the return type and don't do the cast on the return type.

services/server/src/apiv2.yaml
```diff
@@ -6,7 +6,7 @@ paths:
         - Verify Contracts
       summary: Verify Contract (Standard JSON)
       description: |-
-        Submit a contract for verification via the [Solidity standard JSON input](https://docs.soliditylang.org/en/latest/using-the-compiler.html#input-description) or [Vyper JSON input](https://docs.vyperlang.org/en/stable/compiling-a-contract.html#input-json-description).
+        Submit a contract for verification via the [Solidity standard JSON input](https://docs.soliditylang.org/en/latest/using-the-compiler.html#input-description), [Vyper JSON input](https://docs.vyperlang.org/en/stable/compiling-a-contract.html#input-json-description), or Fe JSON input (a Sourcify-defined format — Fe has no official compiler JSON interface).
```
manuelwedler, line 9: Not part of this PR, but we should add a docs page for Fe support and especially describe how the json input should be created for a Fe contract.

services/server/test/helpers/ServerFixture.ts
```diff
@@ -94,6 +100,9 @@ export class ServerFixture {
           solcRepoPath: config.get("solcRepo"),
           solJsonRepoPath: config.get("solJsonRepo"),
           vyperRepoPath: config.get("vyperRepo"),
+          feRepoPath: config.has("feRepo")
+            ? config.get("feRepo")
+            : "/tmp/server-test-fe-repo",
```
manuelwedler: I think the fallback here is not needed anymore. Since config files have been updated.

services/server/test/helpers/ServerFixture.ts
```diff
@@ -81,6 +82,11 @@ export class ServerFixture {
         chains: fixtureOptions_?.chains || sourcifyChainsMap,
         solc: new SolcLocal(config.get("solcRepo"), config.get("solJsonRepo")),
         vyper: new VyperLocal(config.get("vyperRepo")),
+        fe: new FeLocal(
+          config.has("feRepo")
+            ? config.get("feRepo")
+            : "/tmp/server-test-fe-repo",
+        ),
```
manuelwedler: same here, fallback not needed

services/server/test/unit/VerificationService.spec.ts
```diff
@@ -107,6 +107,9 @@ describe("VerificationService", function () {
         solcRepoPath: config.get("solcRepo"),
         solJsonRepoPath: config.get("solJsonRepo"),
         vyperRepoPath: config.get("vyperRepo"),
+        feRepoPath: config.has("feRepo")
+          ? config.get("feRepo")
+          : "/tmp/server-test-fe-repo",
```
manuelwedler: throughout this file the same

services/server/src/server/apiv2/middlewares.ts
kuzdogan: Fe's file structure is as such: https://fe-lang.org/ingots/project-structure/

services/server/src/server/apiv2/middlewares.ts
kuzdogan: True, check the fix

services/server/src/apiv2.yaml
kuzdogan: We don't have docs pages for other languages. Isn't the example  here enough?

packages/compilers/src/lib/feCompiler.ts
kuzdogan: I've added some but it's quite the duplicate of lib-sourcify Compilation tests

kuzdogan:
The Blockscout scraping test is failing which will be removed in https://github.com/argotorg/sourcify/pull/2672

services/server/src/apiv2.yaml
manuelwedler: yeah, probably it's enough. Some note somewhere that we support Fe would still be good. We list languages on the intro page, maybe just add it there.

manuelwedler:
@kuzdogan lib-sourcify tests also fail:

```
@ethereum-sourcify/lib-sourcify:  Exception during run: test/Compilation/FeCompilation.spec.ts(22,9): error TS2345: Argument of type '{ language: "Fe"; sources: { "src/lib.fe": { content: string; }; }; }' is not assignable to parameter of type 'FeJsonInput'.
@ethereum-sourcify/lib-sourcify:   Property 'settings' is missing in type '{ language: "Fe"; sources: { "src/lib.fe": { content: string; }; }; }' but required in type 'FeJsonInput'.
@ethereum-sourcify/lib-sourcify: test/Compilation/FeCompilation.spec.ts(74,9): error TS2345: Argument of type '{ language: "Fe"; sources: { "src/lib.fe": { content: string; }; }; }' is not assignable to parameter of type 'FeJsonInput'.
@ethereum-sourcify/lib-sourcify:   Property 'settings' is missing in type '{ language: "Fe"; sources: { "src/lib.fe": { content: string; }; }; }' but required in type 'FeJsonInput'.
@ethereum-sourcify/lib-sourcify: test/Compilation/FeCompilation.spec.ts(109,9): error TS2345: Argument of type '{ language: "Fe"; sources: { 'src/lib.fe': { content: string; }; 'src/counter.fe': { content: string; }; }; }' is not assignable to parameter of type 'FeJsonInput'.
@ethereum-sourcify/lib-sourcify:   Property 'settings' is missing in type '{ language: "Fe"; sources: { 'src/lib.fe': { content: string; }; 'src/counter.fe': { content: string; }; }; }' but required in type 'FeJsonInput'.
```

kuzdogan:
@manuelwedler fixed

cburgdorf:
>What's the status on the ABI output?

It was merged and will be included in the next release. I'm currently off but I can try to coordinate with the team to cut a new alpha release so you can test out the integration.

cburgdorf:
You can run `feup` to get the latest version which has json abi support.

```
$ fe new foobar
Created ingot at /somepath/foobar
cburgdorf at hackbot in /somepath
$ cd foobar/
cburgdorf at hackbot in /somepath/foobar
$ fe build
Wrote /somepath/foobar/out/Counter.bin
Wrote /somepath/foobar/out/Counter.runtime.bin
Wrote /somepath/foobar/out/Counter.abi.json
cburgdorf at hackbot in /somepath/foobar
$ cat /somepath/foobar/out/Counter.abi.json
[
  {
    "type": "constructor",
    "inputs": [],
    "stateMutability": "nonpayable"
  },
  {
    "type": "function",
    "name": "increment",
    "inputs": [],
    "outputs": [],
    "stateMutability": "nonpayable"
  },
  {
    "type": "function",
    "name": "get",
    "inputs": [],
    "outputs": [
      {
        "name": "",
        "type": "uint256"
      }
    ],
    "stateMutability": "view"
  }
]
```

manuelwedler:
> You can run `feup` to get the latest version which has json abi support.

Awesome! Thanks for this!

@kuzdogan let's update this PR and make the new release the minimum supported Fe version. We can also remove the assumption that abi can be null from the code.

kuzdogan:
## Bug discovered: Signature duplication in API response for multi-source contracts

While adding ABI support for Fe, I discovered a pre-existing bug in the contract lookup query that affects **all** multi-source contracts (Solidity, Vyper, Fe) when the API response includes both `sources` and `signatures` (e.g. `?fields=all`).

See the current bug on prod now: https://sourcify.dev/server/v2/contract/8453/0x3fC91A3afd70395Cd496C647d5a6CC9D4B2b7FAD?fields=all

### Root cause

The lookup query in `Database.ts` (lines 240-251) JOINs both `compiled_contracts_signatures` and `compiled_contracts_sources` on the same `compiled_contracts.id`. For a contract with N source files and M signatures, this creates a Cartesian product of N×M rows before aggregation. The `json_agg` in `generateSignaturesSelector` collected all rows without deduplication, so each signature appeared N times.

Example: a 2-source contract with 3 function signatures returned 6 signature entries instead of 3.

### Why it wasn't caught before

The only existing multi-source test cases either had no ABI/signatures stored (Fe with null ABI) or were single-source Solidity contracts. No test combined multiple sources + signatures + `?fields=all`.

### Fix

Switched from `json_agg(json_build_object(...))` to `jsonb_agg(DISTINCT jsonb_build_object(...))` in `generateSignaturesSelector` (`database-util.ts`). PostgreSQL supports `DISTINCT` on `jsonb` values but not `json`.

Added a multi-source Solidity verification test case (`multi_source_solidity.json`) that reproduces the bug — 2 source files with 3 function signatures.

Commit: ab92338f

kuzdogan:
Can you shortly check the last commits? @manuelwedler

PR #2709, opened 2026-03-26
feat: fetch chain config from remote URL and remove HTML scraping

kuzdogan:
## Summary

This PR makes two related changes to how Sourcify manages chain configuration:

### 1. Remote chain config

Removes the bundled `chains.json`, `sourcify-chains-default.json`, and `extra-chains.json`
from the server, replacing them with a remote fetch at startup from
[sourcifyeth/sourcify-chains](https://github.com/sourcifyeth/sourcify-chains).

- Server now fetches `sourcify-chains-default.json` from the remote URL on startup
- Local `sourcify-chains.json` override still supported for self-hosted instances
- 3-attempt retry with 3s delay if the fetch fails
- `chains.remoteUrl` in `default.js` points to the published chain list
- Chain configuration is now maintained and auto-generated in a dedicated repo with
  nightly CI that pulls from QuickNode, dRPC, Etherscan, and Blockscout APIs

### 2. Remove blockscoutScrape and etherscanScrape

Removes both HTML-scraping methods for contract creation transaction lookup:

- **blockscoutScrape**: all chains migrated to `blockscoutApi`, deprecated, or moved to
  `additional-chains.json` (Bitkub Chain, Velas, Ozone Chain — low activity, no v2 API)
- **etherscanScrape**: defined in the type but never used in any chain configuration

With both removed, the entire scraping infrastructure is also deleted:
`getScrapeContractCreationFetcher`, `getCreatorTxByScraping`, regex constants.

## Test changes

- `chain-tests.spec.ts`: replaced static `sourcify-chains-default.json` import with async
  `initializeSourcifyChains()` + `sourcifyChainsMap`; uses Mocha `--delay` flag
- `ServerFixture`: uses `LOCAL_CHAINS` as default chains so integration tests don't need a remote URL
- `contract-creation-util.spec.ts`: removed two scrape tests (Blockscout old/new regex)
- `sourcify-chains.spec.ts`: new unit tests for `initializeSourcifyChains` covering the
  local override path, remote fetch (happy path, HTTP errors, retries, exhaustion), and
  chain building edge cases (no RPCs, deprecated chains, re-initialization)

## Other fixes

- **`.husky/pre-commit` executable bit**: the hook was committed as `100644` (non-executable)
  so git was silently skipping `npx lint-staged` on every commit in this repo. Fixed by
  `git update-index --chmod=+x .husky/pre-commit` so the hook now actually runs.


kuzdogan:
@claude review

services/server/src/config/default.js
```diff
@@ -45,4 +45,10 @@ module.exports = {
   // verify-deprecated endpoint used in services/database/scripts.mjs. Used when recreating the DB with deprecated chains that don't have an RPC.
   verifyDeprecated: false,
   replaceContract: false,
+  chains: {
+    // URL to fetch the chains config (sourcify-chains.json format) from at server startup.
+    // To use a different set of chains, create a sourcify-chains.json override file — it takes priority.
+    remoteUrl:
+      "https://raw.githubusercontent.com/sourcifyeth/sourcify-chains/main/sourcify-chains-default.json",
```
manuelwedler, line 52: I think we shouldn't fetch the config from GitHub but instead have it in a GCP bucket. If GitHub goes down or is partly not reachable, we won't be able to scale and start up new instances. GitHub downtimes happen too often IMO.

services/server/Dockerfile
```diff
@@ -20,6 +20,9 @@ FROM node:22.22.0-trixie-slim as production
 # Fe compiler (v26.0.0-alpha.12+) requires libssl3 and glibc >= 2.39
 RUN apt-get update && apt-get install -y --no-install-recommends libssl3 && rm -rf /var/lib/apt/lists/*
 
+# Fe compiler (v26.0.0-alpha.12+) is dynamically linked against libssl3
```
manuelwedler: these lines were duplicated

services/server/src/sourcify-chains-default.json
```diff
@@ -1,2571 +0,0 @@
-{
```
manuelwedler, line 1: We should also adapt the `pull_request_template.md`.

services/server/src/chains.json
manuelwedler, line 1: the `update-chains` npm script should also be removed, as well as the `updateChains.mjs`.

services/server/src/server/cli.ts
```diff
@@ -85,149 +88,157 @@ export const fe = new FeLocal(feRepoPath);
...
-      repositoryPath: config.get("repositoryV1.path"),
+
+// Start Server — async because chain config is fetched at startup
+(async () => {
+  logger.info("Starting server with config", {
+    config: JSON.stringify(config.util.toObject(), null, 2),
+  });
+
+  // Load chain config first so getEtherscanApiKeyForEachChain() sees the populated map
+  await initializeSourcifyChains();
```
manuelwedler: I tend to say that this function should return the `sourcifyChainsMap` instead of modifying a shared object. The current pattern seems error-prone to me.

services/server/src/sourcify-chains.ts
```diff
@@ -239,95 +202,111 @@ function buildCustomRpcs(
...
+          );
+          await new Promise((resolve) => setTimeout(resolve, retryDelayMs));
+        }
+      }
+    }
+    if (!chainsExtensions!) {
+      throw new Error(
+        `Failed to fetch chains config after ${maxAttempts} attempts: ${lastError?.message}`,
+      );
     }
```
manuelwedler, line 271: nitpick: the logic of the else branch could be refactored into a function for improved readability.

services/server/test/helpers/ServerFixture.ts
```diff
@@ -62,6 +66,8 @@ export class ServerFixture {
     this.repositoryV1Path = config.get<string>("repositoryV1.path");
 
     before(async () => {
+      await initializeSourcifyChains();
+
```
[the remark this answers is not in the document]
manuelwedler: Agreed. For testing, we shouldn't call this function and instead have some hardcoded chain map solely for testing.

services/server/src/sourcify-chains.ts
```diff
@@ -1,19 +1,16 @@
...
   SourcifyRpc,
 } from "@ethereum-sourcify/lib-sourcify";
 import { SourcifyChain } from "@ethereum-sourcify/lib-sourcify";
-import chainsRaw from "./chains.json";
-import extraChainsRaw from "./extra-chains.json";
-import rawSourcifyChainExtentions from "./sourcify-chains-default.json";
 import logger from "./common/logger";
 import fs from "fs";
 import path from "path";
+import config from "config";
```
manuelwedler, line 13: I think it would be better to not make this file importing config, and leave any config resolution to the `cli.ts` file in order to reduce side-effects.

services/server/src/sourcify-chains.ts
```diff
@@ -239,95 +202,111 @@ function buildCustomRpcs(
...
+      );
     }
   }
 
-  if (chainId in sourcifyChainsExtensions) {
-    const sourcifyExtension = sourcifyChainsExtensions[chainId];
+  // Clear the map before populating (allows re-initialization)
+  for (const key of Object.keys(sourcifyChainsMap)) {
+    delete sourcifyChainsMap[key];
+  }
```
manuelwedler: not necessary when you return an object from this function and not write to a shared one.

services/server/test/unit/verificationWorker.spec.ts
```diff
@@ -35,7 +38,9 @@ describe("verificationWorker", function () {
   const chainFixture = new LocalChainFixture();
   const piscinaSandbox = Sinon.createSandbox();
 
-  before(() => {
+  before(async () => {
+    await initializeSourcifyChains();
+
```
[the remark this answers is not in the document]
manuelwedler: can we simply use `LOCAL_CHAINS` ?

services/server/test/chains/chain-tests.spec.ts
```diff
@@ -55,213 +58,220 @@ if (process.env.NEW_CHAIN_ID) {
...
-    const createXDeployments: { chainId: string }[] = await createXRes.json();
-    const multicall3Deployments: { chainId: string }[] =
-      await multicall3Res.json();
-
-    createXChainIds = new Set(
-      createXDeployments.map((d) => d.chainId.toString()),
+// Chains config is loaded async; use Mocha's --delay + run() to defer test registration
+(async () => {
+  await initializeSourcifyChains();
+
```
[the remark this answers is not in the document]
manuelwedler: Not 100 % sure, but this sounds right.

manuelwedler (changes requested):
Will review the new chain repo next.

manuelwedler:
REMINDER: do a release shortly before merging this.

kuzdogan:
Reminder: Add chain https://github.com/argotorg/sourcify/pull/2755

packages/compilers-types/CHANGELOG.md
```diff
@@ -1,5 +1,3 @@
-
```
manuelwedler, line 1: This shouldn't be here and would trigger a new package release for an untouched package.

This is the second time I see this from Claude commits. don't know how this happens, but this was the reason why we had to make an empty release once.

packages/compilers/CLAUDE.md
```diff
@@ -11,12 +11,12 @@ Sourcify uses two compiler paths:
...
 For old soljsons that predate Standard JSON, the `solc` npm package (`wrapper.js`) translates Standard JSON input/output to the old format. It detects available C functions and uses the highest-priority one available:
 
-| Priority | C function | Introduced | Input format |
-|---|---|---|---|
-| 1 | `solidity_compile` | v0.5.0 | Standard JSON (new libsolc API, replaces all older functions) |
-| 2 | `compileStandard` | v0.4.11 | Standard JSON |
-| 3 | `compileJSONCallback` | v0.2.1 | `{"sources":{"f.sol":"..."}}` + import callback |
-| 4 | `compileJSONMulti` | v0.1.6 | `{"sources":{"f.sol":"..."}}` |
-| 5 | `compileJSON` | v0.1.0 | Raw source string (single file only) |
+| Priority | C function            | Introduced | Input format                                                  |
```
manuelwedler: unrelated change that would trigger a new package release.

packages/compilers/CHANGELOG.md
```diff
@@ -1,5 +1,3 @@
-
```
manuelwedler, line 1: unrelated change that would trigger a new package release.

services/4byte/CHANGELOG.md
```diff
@@ -1,5 +1,3 @@
-
```
manuelwedler, line 1: unrelated change that would trigger a new package release.

services/server/package.json
```diff
@@ -102,7 +97,6 @@
     "copyfiles": "2.4.1",
     "hardhat": "2.28.6",
     "mocha": "10.8.2",
-    "mochawesome": "7.1.4",
```
manuelwedler, line 105: I'm missing a change in `package-lock.json` for this

services/server/src/sourcify-chains-default.json
manuelwedler: This is not resolved.

services/server/src/chains.json
manuelwedler: Not done

services/server/test/integration/apiv1/verification-handlers/verify.stateless.spec.ts
```diff
@@ -17,13 +17,25 @@ import {
...
 chai.use(chaiHttp);
 
 describe("/", function () {
   const chainFixture = new LocalChainFixture();
-  const serverFixture = new ServerFixture();
+  const serverFixture = new ServerFixture({
+    chains: {
+      ...Object.fromEntries(LOCAL_CHAINS.map((c) => [c.chainId.toString(), c])),
+      "5": new SourcifyChain({
+        name: "Goerli (deprecated stub)",
```
manuelwedler: why do we need a deprecated chain in this test file?

services/server/test/helpers/etherscanResponseMocks.ts
```diff
@@ -18,6 +18,16 @@ export const mockEtherscanApi = (
       : process.env[sourcifyChain.etherscanApi.apiKeyEnvName || ""] ||
         process.env.ETHERSCAN_API_KEY ||
         "";
+  const customUrl = sourcifyChain.etherscanApi?.url;
+  if (customUrl) {
+    return nock(customUrl)
+      .get(
+        `/api?module=contract&action=getsourcecode&address=${contractAddress}&apikey=${apiKey}`,
```
manuelwedler: why is a custom url called without chain id parameter?

services/server/test/helpers/mockRpcServer.ts
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Real local HTTP server that responds to JSON-RPC (eth_getCode,
+ * eth_getTransactionByHash, eth_getTransactionReceipt) and GET /api
+ * (Etherscan-style action=getcontractcreation) for a given chain.
+ *
+ * Sourced from etherscanBytecodes.json. Must be a real server (not nock)
+ * because v2 verification runs in Piscina worker threads and nock's
+ * HTTP-module patches don't propagate across thread boundaries.
+ */
+export const startMockRpcServer = (
```
manuelwedler: Why do we need this?

services/server/test/integration/apiv2/verification/verify.etherscan.spec.ts
```diff
@@ -21,25 +22,87 @@ import {
...
   }
 
+  const MOCK_RPC_PORT = 18546;
+  let mockRpcServer: http.Server;
+  before(async () => {
+    mockRpcServer = await startMockRpcServer("1", MOCK_RPC_PORT);
+  });
+  after(async () => {
+    await stopMockRpcServer(mockRpcServer);
+  });
```
manuelwedler: I don't see the point of running and maintaining a mock rpc server. Can we deploy the contracts to the local  hardhat chain and call the etherscan endpoints with the hardhat chain id?

manuelwedler (changes requested):

services/server/test/helpers/etherscanResponseMocks.ts
kuzdogan: The mock mirrors what the production code does — `chainid` is only included when hitting the multi-chain endpoint at `api.etherscan.io/v2`. A custom URL is treated as a per-chain endpoint (self-hosted explorer, Polygonscan, etc.) where the chain is implicit in the URL.

**`fetchFromEtherscan` (sourcecode lookup):**

https://github.com/argotorg/sourcify/blob/72a35f463adbb99c39a667fdc1e8d24a82a66335/packages/lib-sourcify/src/utils/etherscan/etherscan-util.ts#L255-L257

**Creator-tx fetcher:**

https://github.com/argotorg/sourcify/blob/72a35f463adbb99c39a667fdc1e8d24a82a66335/services/server/src/server/services/utils/contract-creation-util.ts#L45-L46

Default (multi-chain) template:

https://github.com/argotorg/sourcify/blob/72a35f463adbb99c39a667fdc1e8d24a82a66335/services/server/src/server/services/utils/contract-creation-util.ts#L6

Both production functions split on `customBaseUrl` and only the multi-chain path carries `chainid`. The mock does the same for `getsourcecode`.

packages/compilers-types/CHANGELOG.md
kuzdogan: The reason is we added a pre-commit hook to run Prettier (`.husky/pre-commit`) I guess we need to run a full linting once. But it says it's "lint-staged" so this shouldn't have been linted?

services/server/test/helpers/etherscanResponseMocks.ts
manuelwedler: I think it was missed to change the code here earlier. Since Etherscan API v2 is out, we never call this function with a cutom url in production. IMO, we can remove this option.

packages/compilers-types/CHANGELOG.md
manuelwedler: The thing is other PRs didn't have this problem. Maybe it happened while merging staging into this branch?

services/server/test/integration/apiv2/verification/verify.etherscan.spec.ts
kuzdogan: Looked into this. The mock server *can't* be fully removed, but it can shrink significantly. Here's why:

The v2 import flow splits across two threads:

- **Parent thread** (request handler): fetches `action=getsourcecode` from Etherscan. nock works here, so `etherscanResponseMocks.ts` handles it cleanly.
- **Piscina worker thread** (`verificationWorker.ts`): calls `getCreatorTx` → `action=getcontractcreation`, plus the chain RPC (`eth_getCode`, `eth_getTransactionByHash`, `eth_getTransactionReceipt`). The worker has its own `http` module — nock's parent-thread patches don't apply, so any HTTP it makes has to hit a real local server.

That's why `mockRpcServer.ts` exists today — it's the only way to mock the worker's HTTP calls. It currently serves both the chain RPC *and* the Etherscan getcontractcreation endpoint.

**What your suggestion buys us:** the JSON-RPC half goes away. We can deploy each test contract to the local hardhat chain by sending its `creationBytecode` (already present in `etherscanBytecodes.json`) as a raw deployment tx — same EVM result, same runtime bytecode, real `eth_getCode` / `eth_getTransactionByHash` from hardhat. The deployed address replaces the hardcoded mainnet one in the Etherscan response (the source code, compiler version, and constructor args stay the same, so partial-vs-exact-match characteristics are preserved). The `mainnetStub` then points its rpcs at hardhat instead of the mock.

**What still needs the mock:** the `action=getcontractcreation` endpoint, because the worker calls it. So `mockRpcServer.ts` can drop to ~40 lines (just that one Etherscan endpoint) instead of ~190.

Trade-off: a few hours of work to redo the deployment + address-substitution wiring, in exchange for ~140 lines of test infra deleted and the test setup mirroring how the chain actually works. Worth doing in this PR, or fold it in as a follow-up so we can land the chain-config change first?

services/server/test/helpers/etherscanResponseMocks.ts
kuzdogan: We don't call this *yet*. This was added in anticipation of the Battlechain testnet's explorer, check out the related commit's comment https://github.com/argotorg/sourcify/pull/2732/changes/384c4451f460541b010115c670961d1cc96ba0aa

services/server/test/integration/apiv2/verification/verify.etherscan.spec.ts
manuelwedler: This doesn't sound right to me. The rpc won't be called from the worker when it calls `action=getcontractcreation`. I think AI is mixing things here.

Hardhat is already a separated process that provides all rcp methods if we deploy the needed contract there.

kuzdogan: It says the RPC will work but Etherscan `action=getcontractcreation` will not, what you're suggesting is the other way around. 

This here in the VerificationService, is in the worker, no?
https://github.com/argotorg/sourcify/blob/9c49980d35ca2bf9f9a3db9fdfe8334a77150e7d/services/server/src/server/services/VerificationService.ts#L235-L239

This was causing the tests to fail when developing this and this fix made it work really and the explanation made sense to me.

manuelwedler: I now got what is wrong here. The naming is off. This does not only mock a JSON RPC server but also the Etherscan API in the same place. I didn't see it mocks Etherscan because I assumed it would just be JSON RPC due to the naming.

I am still sure we can get rid of it fully. You can deploy the same contract as in the Etherscan mock responses. You don't need the Etherscan API inside verification worker. Instead let it hit the binary search and it would only rely on the RPC calls to get the creation tx, which is not costly on a Hardhat chain since it should have just one or two blocks in the test. The only issue with that is the current Etherscan response mocks (at least partially) use contracts that have constructors which fail if not another contract is deployed (I tried locally with the contract from single response).

So I see multiple options to fix this now:

1. Get new Etherscan response mocks for which we can deploy the contracts without failing constructors and do as I proposed in the paragraph before (most clean result, but probably too much work)
1. Just mock RPCs in the mock server and don't mock Etherscan API. Disable contract creation fetching via Etherscan and rely on binary search calls which will got to the mock rpc. Don't couple the mock server strictly to the Etherscan bytecodes, so it can be reused as a mock RPC in case we need it.
1. Give the mock server a proper name that is a general purpose mock http server. Make it clear in the file name and folder structure that the mock server is just needed for etherscan tests.

services/server/test/helpers/etherscanResponseMocks.ts
manuelwedler: ah sorry didn't know about this

kuzdogan:
Ready for review

manuelwedler (approved):
Finally ready :)

Looks clean now. Didn't expect you go the route with adding new Etherscan mocks, but pretty happy with the result.

PR #2725, opened 2026-03-31
Handle Etherscan's response with malformed version

marcocastignoli:
See #2716


marcocastignoli:
The problem is bigger than what's covered by this PR:
- versions without `nightly` mapping to a nightly version
- incomplete commit hashes

In order to properly address this issue we must implement a more advanced mechanism

PR #2727, opened 2026-03-31
fix(ci): increase build job resource class to xlarge to fix OOM

kuzdogan:
## Summary
- Docker build jobs were OOM-killed (exit 137) due to insufficient memory
- Upgrade `resource_class` from `large` (8 GB RAM) to `xlarge` (16 GB RAM) for all Docker build jobs in both `config.yml` and `continue_config.yml`
- Applies to both amd64 (`large` → `xlarge`) and arm64 (`arm.large` → `arm.xlarge`) builds

## Test plan
- [ ] Verify build jobs complete without OOM errors on next CI run

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,7 +148,7 @@ aliases:
   - &build-base
     docker:
       - image: cimg/base:2024.07
-    resource_class: large
+    resource_class: xlarge
     parameters:
       run-build:
         type: boolean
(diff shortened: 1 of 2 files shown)


manuelwedler (approved):

manuelwedler:
We could think of running the tests of individual packages in separate, parallel CI jobs. Would speed up things and we should be able to use smaller resource classes.

PR #2732, opened 2026-04-07
Handle etherscan Solidity malformed version

marcocastignoli:
See https://github.com/argotorg/sourcify/issues/2716

This PR introduces a mechanism that finds the closest solidity version to a malformed one coming from Etherscan using the [Solidity list file](https://binaries.soliditylang.org/bin/list.txt). In Vyper we already use a similar mechanism in which when Etherscan returns a malformed Vyper version a file containing all the Vyper releases is fetched and the closest version is used, but for the Solidity case we don't need to fetch the file on the fly because the only affected Solidity versions are old ones.

The resolver handles the following malformation patterns:
- **Old date-based format**: `v0.3.2-2016-04-18-81ae2a7` — extracts the version and truncated commit hash, matches against the list to find `0.3.2+commit.81ae2a78`
- **Truncated commit hashes**: `0.3.2+commit.81ae2a7` (7 chars instead of 8) — prefix-matches against the list
- **Missing commit hash**: `0.8.17` — resolves to the full version with commit hash
- **Nightly versions with wrong date or unknown commit**: falls back to the stable release for that base version (e.g. `0.6.0-nightly.2019.3.11+commit.4704ef84` → `0.6.0+commit.26b70077`)


packages/lib-sourcify/src/utils/etherscan/etherscan-util.ts
```diff
@@ -97,6 +98,88 @@ export const getVyperCompilerVersion = async (
   return found;
 };
 
+export function resolveSolidityVersion(version: string): string {
+  // Only resolve if the version doesn't match the standard format
+  if (/^\d+\.\d+\.\d+\+commit\.[0-9a-f]{7,8}$/.test(version)) {
+    // Check if it's an exact match in the list (handles truncated hashes)
```
kuzdogan: The comment "handles truncated hashes" on line 104 is slightly misleading. I was trying to understand how a 7 char hash is being completed to 8 char but it actually doesn't, since there is no 7 char hashes in the list file. 7 char always gets skipped here and gets resolved later in:

```ts
    if (
      entryCommit.startsWith(commitPrefix) ||
      commitPrefix.startsWith(entryCommit)
    ) {
```

kuzdogan (commented):
I'm not fully comfortable with the workaround here:

- We shouldn't keep a `solidity-bin-list.ts` file IMO. We need to keep this up to date constantly. 
- Should we move this logic to where we are fetching solc binaries? 
  - Try fetching the full binary URL
  - If it fails, handle date format
  - If it fails fetch `list.txt` on the fly and apply the resolution logic here. It should be ok to  fetch the list on the fly since this will be rare?


Or maybe we should keep a shorter local `solidity-bin-list.ts` that will only apply to earlier versions that will clearly not be kept up to date, and we keep this in the etherscan utils.

marcocastignoli:
> Or maybe we should keep a shorter local solidity-bin-list.ts that will only apply to earlier versions that will clearly not be kept up to date, and we keep this in the etherscan utils.

That was exactly my idea while working on this, as I wrote in the description:

> In Vyper we already use a similar mechanism in which when Etherscan returns a malformed Vyper version a file containing all the Vyper releases is fetched and the closest version is used, but for the Solidity case we don't need to fetch the file on the fly because the only affected Solidity versions are old ones.

As you mentioned this file can actually be shorter but I don't know where to cut it, we should investigate when is the last occurrence of a malformed version, do you think it's worth doing this investigation? At the end the file is not that big and the process is basically instant, it's not a performance bottleneck.

marcocastignoli:
> Should we move this logic to where we are fetching solc binaries?  [...] If it fails fetch list.txt on the fly and apply the resolution logic here. It should be ok to fetch the list on the fly since this will be rare?

I think if we move it to the solidity binary fetching then it's because we want to support bad malformed versions even in other APIs, then it makes sense to include all versions and fetch it on the fly at start.
But I think this really makes no sense, we've always forced the user to pass the specific version, what I implemented in this PR it's just a workaround for malformed versions coming from Etherscan.

packages/lib-sourcify/src/utils/etherscan/etherscan-util.ts
marcocastignoli: Ups, I totally missed this wrong comment, thanks :)

kuzdogan:
> As you mentioned this file can actually be shorter but I don't know where to cut it, we should investigate when is the last occurrence of a malformed version, do you think it's worth doing this investigation? At the end the file is not that big and the process is basically instant, it's not a performance bottleneck.

Ok fine for me. Can you just make it very clear the purpose of the `solidity-bin-list.ts` file in the comment so that it won't be confused with ordinary binary fetching?

kuzdogan (approved):

marcocastignoli:
Ups I realized this just after merging. Is this intended to be here? 

@kuzdogan https://github.com/argotorg/sourcify/pull/2732/changes/384c4451f460541b010115c670961d1cc96ba0aa

PR #2735, opened 2026-04-07
fix(lib-sourcify): don't misdiagnose extra_file_input_bug for pre-0.4.7 contracts

### LINKED ISSUE #2729: v0.1.7 extra_file_input_bug: Standard JSON recompilation produces different bytecode
## Contract

- **Address:** `0xd0af1e9919f6321f1dfef2d9cbb60ef7a10d6c99` (Ethereum mainnet)
- **Compiler:** `v0.1.7+commit.b4e666cc`
- **Optimizer:** Enabled (runs: 200)
- **Contract name:** SmartShipp
- **Creation TX:** `0x4302d8f517f126352b23223bc0e46aa199eb5d7b8bb96d2d3aa3b5c316b3d33a`

## Problem

Submitting via the v2 verify endpoint with `stdJsonInput` returns `extra_file_input_bug`:

> It seems your contract's metadata hashes match but not the bytecodes.

The contract compiles correctly with the old `compileJSON()` API (exact creation TX match confirmed), but Standard JSON input produces different bytecode for this compiler version.

## Source

```solidity
contract SmartShipp {
    string public name;
    string public symbol;
    uint8 public decimals;
    mapping (address => uint256) public balanceOf;
    event Transfer(address indexed from, address indexed to, uint256 value);

    function SmartShipp(uint256 initialSupply, string tokenName, string tokenSymbol, uint8 decimalUnits) {
        if (initialSupply == 0) initialSupply = 1000000;
        balanceOf[msg.sender] = initialSupply;
        name = tokenName;
        symbol = tokenSymbol;
        decimals = decimalUnits;
    }

    function transfer(address _to, uint256 _value) {
        if (balanceOf[msg.sender] < _value) throw;
        if (balanceOf[_to] + _value < balanceOf[_to]) throw;
        balanceOf[msg.sender] -= _value;
        balanceOf[_to] += _value;
        Transfer(msg.sender, _to, _value);
    }
}
```

## Context

This is a December 2015 token contract based on the ethereum.org token tutorial. We verified it byte-for-byte using the old `compileJSON()` API as part of the [EthereumHistory](https://ethereumhistory.com) project. The verification repo is at [cartoonitunes/smartshipp-verification](https://github.com/cartoonitunes/smartshipp-verification).

Related to #618.

kuzdogan:
## Summary

- `handleSolidityExtraFileInputBug()` was incorrectly returning `EXTRA_FILE_INPUT_BUG` for pre-0.4.7 Solidity contracts (e.g. v0.1.7) whenever bytecodes didn't match and the optimizer was enabled
- Root cause: pre-0.4.7 contracts have no CBOR metadata, so `splitAuxdata()` returns a single-element array — the auxdata at index `[1]` is `undefined` for both onchain and recompiled bytecodes. Since `undefined === undefined` is `true`, the check falsely triggered
- Fix: guard that both auxdata values must be defined (i.e. CBOR metadata actually exists in both bytecodes) before diagnosing the extra file input bug
- Added some documentation in compilers CLAUDE.md about the findings during the investigation. I don't know the best place to put this.

Fixes #2729

> **Note:** This fixes the misdiagnosis. The deeper issue (Standard JSON producing different bytecode than the old `compileJSON()` API for very old compiler versions) is tracked separately.

### DIFF
--- a/packages/compilers/CLAUDE.md
+++ b/packages/compilers/CLAUDE.md
@@ -1,0 +1,22 @@
+## Solc Compiler Internals
+
+### soljson vs native binary
+
+Sourcify uses two compiler paths:
+
+- **Native binary** — invoked via `solc --standard-json`. Available from v0.4.10 (Linux), v0.3.6 (macOS), v0.4.1 (Windows).
+- **soljson** — Emscripten-compiled JavaScript build. Used for all versions below the native binary thresholds. Downloaded from `binaries.soliditylang.org/bin/`.
+
+### solc npm wrapper — C function fallback chain
+
+For old soljsons that predate Standard JSON, the `solc` npm package (`wrapper.js`) translates Standard JSON input/output to the old format. It detects available C functions and uses the highest-priority one available:
+
+| Priority | C function | Introduced | Input format |
+|---|---|---|---|
+| 1 | `solidity_compile` | v0.5.0 | Standard JSON (new libsolc API, replaces all older functions) |
+| 2 | `compileStandard` | v0.4.11 | Standard JSON |
+| 3 | `compileJSONCallback` | v0.2.1 | `{"sources":{"f.sol":"..."}}` + import callback |
+| 4 | `compileJSONMulti` | v0.1.6 | `{"sources":{"f.sol":"..."}}` |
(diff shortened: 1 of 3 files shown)


manuelwedler (approved):
lgtm

PR #2739, opened 2026-04-09
feat: support Vyper storage_layout_overrides, storage layout, and creation sourceMap

### LINKED ISSUE #2002: Support `storage_layout_overrides` of Vyper's json input
Vyper added an option `storage_layout_overrides` in the json input which can change the bytecode output (https://github.com/vyperlang/vyper/pull/4370). We need to support it for Verification after completing #2001.

## Tasks
- [ ] Pass the new field to the vyper compiler from APIs
- [ ] After successful verification, store the field in the database inside `additional_inputs` of `compiled_contracts`

manuelwedler:
## Summary

- Support Vyper's `storage_layout_overrides` top-level JSON input field through the full verification pipeline (API input → compilation → verification export → DB storage → API v2 response)
- Capture Vyper's `layout` output (storage layout, available since 0.4.1) and `evm.bytecode.sourceMap` (creation sourceMap, available since 0.4.0) which were previously ignored
- Define `VyperStorageLayout` type and expose `additionalInput` as a queryable field in API v2

## Changes

**Types:**
- Add `VyperStorageLayout` type, reuse for `storage_layout_overrides` input and `layout` output
- Add `layout` and `evm.bytecode.sourceMap` to `VyperOutputContract`
- Add `storageLayoutOverrides` to `VerificationExport`
- Add `additionalInput` field to API v2 `VerifiedContract` type

**Compilation & Verification:**
- Request `layout` (always) and `evm.bytecode.sourceMap` (>= 0.4.0) in Vyper output selection
- Export `storageLayoutOverrides` and Vyper storage layout from `Verification.export()`
- Extract Vyper `layout.storage_layout` as `storageLayout` in DB columns

**Database:**
- Store `storage_layout_overrides` in `additional_input` column of `compiled_contracts`
- Include `additional_input` in `std_json_input` SQL reconstruction via jsonb merge
- Add `additionalInput` as queryable field in API v2

**API:**
- Add Vyper `storage_layout_overrides` example to API v2 verify endpoint
- Document `additionalInput` field in API v2 schema

## Test plan

- [x] lib-sourcify: compilation test proving `storage_layout_overrides` changes bytecode
- [x] lib-sourcify: verification test asserting `storageLayoutOverrides` in export
- [x] lib-sourcify: verification test asserting exported storage layout and sourceMaps
- [x] Server: verification-cases test with DB + API v2 assertions for `storage_layout_overrides`
- [x] Server: updated Vyper test data (0.4.0, 0.4.1, 0.4.1b1) with creation sourceMaps and layout
- [x] All 153 lib-sourcify tests pass
- [x] All 296 server tests pass (1 pre-existing veChainApi failure)

Closes #2002


manuelwedler:
Again, good findings by copilot. Fixed them in the last commit

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
```diff
@@ -103,21 +103,29 @@ export class VyperCompilation extends AbstractCompilation {
   public compilerVersionCompatibleWithSemver: string;
 
   initVyperJsonInput() {
+    const outputs = [
+      'abi',
+      'ast',
+      'interface',
+      'ir',
+      'layout',
```
kuzdogan, line 111: Shouldn't we also version filter this after 0.4.1 if this is available as such as mentioned in the PR description? It's inconsistent with sourceMap

packages/compilers-types/src/VyperTypes.ts
```diff
@@ -73,15 +76,23 @@ export interface VyperSourceMap {
...
   devdoc: Devdoc;
   ir: string;
+  layout?: {
+    storage_layout: VyperStorageLayout;
+  };
   evm: {
     bytecode: {
       object: string;
       opcodes: string;
+      sourceMap?: VyperSourceMap;
```
kuzdogan, line 95: So VyperSourceMap was already defined but we forgot to add it here?

packages/compilers-types/src/VyperTypes.ts
```diff
@@ -73,15 +76,23 @@ export interface VyperSourceMap {
   pc_pos_map_compressed: string;
 }
```
kuzdogan, line 77: `breakpoints` and `pc_breakpoints` types only accept empty arrays. Is this expected?

services/server/src/server/services/utils/database-util.ts
```diff
@@ -425,11 +431,20 @@ export const STORED_PROPERTIES_TO_SELECTORS = {
   devdoc: "compiled_contracts.compilation_artifacts->'devdoc' as devdoc",
   source_ids:
     "compiled_contracts.compilation_artifacts->'sources' as source_ids",
-  std_json_input: `json_build_object(
-    'language', INITCAP(compiled_contracts.language), 
-    'sources', ${sourcesAggregation},
-    'settings', compiled_contracts.compiler_settings
-  ) as std_json_input`,
+  additional_input: "compiled_contracts.additional_input",
+  std_json_input: `CASE
```
kuzdogan: Can we deduplicate this as 

```
json_build_object('language', ...)::jsonb 
|| COALESCE(compiled_contracts.additional_input, '{}'::jsonb)
as std_json_input
```

kuzdogan (commented):
- [x] I guess we need to add this to repo UI too. Meaning we need to handle the Vyper storageLayout format as well as display when there are storageLayoutOverrides in the additionalInput.
- [x] If Vyper has different storageLayout format in the output maybe small comment in the apiv2.yaml about this.
- [x] I guess part of this documentation is missing in Vyper docs? While we have the LLM context, should we open a PR to the Vyper docs?

packages/compilers-types/src/VyperTypes.ts
manuelwedler: I think it will never be emitted actually, because the relevant code is dead according to a comment. Claude pointed me to this source file: https://github.com/vyperlang/vyper/blob/master/vyper/evm/assembler/symbols.py

Theoretically, it is a `number[]`. So I changed the type.

packages/compilers-types/src/VyperTypes.ts
manuelwedler: We already had it for `deployedBytecode.sourceMap`. It seems like `bytecode.sourceMap` was added in a later Vyper version.

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
manuelwedler: I intentionally kept it like this. I tried if it errors on older versions, but it does not. Some more fields we ask for, are not supported by older versions too, I think. IMO, if old versions ignore output selections, it is fine to just always ask for them to make the code less bulky. Only `evm.bytecode.sourceMap` requires it, because the older vyper versions were erroring. I assume this is due to the `.` notation for nested access.

manuelwedler:
> [x] I guess part of this documentation is missing in Vyper docs? While we have the LLM context, should we open a PR to the Vyper docs?

PR for Vyper is here: https://github.com/vyperlang/vyper/pull/4913

PR for repo will follow.

manuelwedler:
PR for repo: https://github.com/sourcifyeth/repo.sourcify.dev/pull/53

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
kuzdogan: Ah ok. So if you ask for `evm.bytecode.sourceMap` and it's not in the version it throws? Then ok

kuzdogan (approved):

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
manuelwedler: exactly

PR #2743, opened 2026-04-13
refactor: move v1 verification edge case tests to v2 and verification-cases

### LINKED ISSUE #2640: Refactor relevant test cases from API v1 tests
_Follow up from #2049_ 

The `verify.stateless.spec.ts` test file has a bunch of tests that doesn't test the behavior of the stateless endpoints directly, but instead tests the behavior of our verification logic.

Here we test verification edge cases, which should be moved to `verification-cases.spec.ts`:

https://github.com/argotorg/sourcify/blob/f16ad55943aa7b9f34ae88f4b196d7163a7106eb/services/server/test/integration/apiv1/verification-handlers/verify.stateless.spec.ts#L1031-L1218

The following tests check for match upgrades, which we don't test on API v2 at all:

https://github.com/argotorg/sourcify/blob/f16ad55943aa7b9f34ae88f4b196d7163a7106eb/services/server/test/integration/apiv1/verification-handlers/verify.stateless.spec.ts#L459-L467

Since API v1 will be removed eventually, we should check all API v1 tests if we would like to refactor some of them and test via the new API. There also might be more inside `verify.stateless.spec.ts`. I didn't check it fully.

### Tasks

- [x] Refactor the above mentioned test cases
- [ ] Check the other API v1 tests and refactor the relevant cases to other files

manuelwedler:
## Summary

- Move verification edge case tests from API v1 `verify.stateless.spec.ts` to their proper locations
- Match upgrade tests (null→perfect, partial→perfect) → `apiv2/verify.json.spec.ts`
- `extraFilesBytecodeMismatch` tests → `apiv2/verify.metadata.spec.ts`
- Wrong metadata storage correctness → `apiv2/verify.metadata.spec.ts`
- Pre-0.5.0 non-keccak library placeholders with call protection → `apiv2/verify.metadata.spec.ts`
- Factory-deployed contracts (2 tests) → new `verification-cases/factories.spec.ts`
- Deprecated chain (Goerli) test → `private.stateless.spec.ts`
- Remove unused `testPartialUpgrade` helper

## Test plan
- [ ] Run `verify.json.spec.ts` — match upgrade tests pass
- [ ] Run `verify.metadata.spec.ts` — extraFilesBytecodeMismatch, wrong metadata, and library tests pass
- [ ] Run `factories.spec.ts` — factory-deployed contract tests pass
- [ ] Run `private.stateless.spec.ts` — deprecated chain test passes
- [ ] Run `verify.stateless.spec.ts` — remaining v1 tests still pass

Closes #2640


kuzdogan:
Can you fix the lint? I can review this

manuelwedler:
should be fixed, but wait until the tests ran fine

manuelwedler:
now it's good for review

kuzdogan (commented):
(removing malformed comment)

kuzdogan (commented):
`metadataExpected: false` is unexpected here — this flag is never used as `false` anywhere else in the test suite (checked `assertVerification` and `assertJobVerification` across all integration tests). Passing `false` silently skips the repo-v1 filesystem metadata check in `assertContractSaved`. Is it expected that factory-deployed contracts don't get their metadata persisted to the filesystem? If so, please add a comment explaining why — otherwise this could be masking a storage regression.

kuzdogan (commented):
**Correction to my earlier comment about `metadataExpected: false`**: my previous multi-line grep missed it — the flag IS used in `vyper.stateless.spec.ts` (lines 94, 176, 223, 274), so it's not unique to this PR.

However, the concern still stands, just reframed: the Vyper usage makes sense because Vyper contracts don't produce a solc-style `metadata.json` for filesystem storage. But the factory tests here are verifying **Solidity** contracts (`FactoryImmutable`, `FactoryImmutableWithoutConstrArg`), which should produce metadata normally. Why does factory-deployment change that? If factory-deployed Solidity contracts really do skip metadata storage, please add a comment explaining the mechanism; otherwise this may still be masking a gap.

kuzdogan (commented):
FYI — I had my agent scan the remaining v1 tests for cases that test verification logic (not stateless-endpoint behavior) and aren't already covered elsewhere. Most of what I found is already redundant with `verification-cases/testdata/` or covered implicitly by v2 tests, but three cases look like unique coverage that would be lost if v1 is eventually removed:

1. **`should fetch a missing file that is accessible via ipfs`** (`verify.stateless.spec.ts:154`) — tests that the verifier fetches missing sources from IPFS. Grepping `apiv2/` and `verification-cases/` for `ipfs`/`IPFS` returned no source-fetching tests anywhere — this behavior appears to be entirely untested outside of v1.

2. **`should have an auxdata transformation for a contract with a wrong cbor auxdata on Vyper 0.3.8`** (`vyper.stateless.spec.ts:194`) — `verification-cases/testdata/vyper/auxdata-0.3.8.json` covers the happy path, but not the wrong-auxdata replacement scenario.

3. **`should have an auxdata transformation for a contract with a wrong cbor auxdata on Vyper 0.3.4`** (`vyper.stateless.spec.ts:245`) — same as above for 0.3.4; `auxdata-0.3.4.json` covers the happy path only.

Posting this as FYI — feel free to decide yourself if it's worth acting on, or feed it into your agent.

manuelwedler:
Moved the additional tests and removed the `metadataExpected: false` flag.

kuzdogan (approved):

PR #2749, opened 2026-04-14
refactor: return SourcifyChainMap from initializeSourcifyChains

kuzdogan:
## Summary

Addresses Manuel's review comment on #2709 (`cli.ts:99`):
> I tend to say that this function should return the `sourcifyChainsMap` instead of modifying a shared object. The current pattern seems error-prone to me.

And the follow-up on `sourcify-chains.ts:277`:
> not necessary when you return an object from this function and not write to a shared one. (referring to the map-clearing loop)

## What changed

- **`src/sourcify-chains.ts`**: Remove the exported `sourcifyChainsMap` singleton. `initializeSourcifyChains()` now returns `Promise<SourcifyChainMap>`. The map-clearing loop is gone (each call builds a fresh local map). The unusual `if (!chainsExtensions!)` non-null assertion is cleaned up to `let chainsExtensions: ... | undefined` + `if (!chainsExtensions)`. All chainId keys are now consistently strings (`chainId.toString()`).
- **`src/server/cli.ts`**: Capture the returned map with `const sourcifyChainsMap = await initializeSourcifyChains()`. `getEtherscanApiKeyForEachChain` now accepts the map as a parameter rather than closing over the module-level export.
- **`test/helpers/ServerFixture.ts`**: Store the returned map on `this.sourcifyChainsMap` (public field). Integration tests access it via `serverFixture.sourcifyChainsMap[...]`.
- **Unit tests** (`sourcify-chains.spec.ts`, `verificationWorker.spec.ts`, `contract-creation-util.spec.ts`): Use the returned value; remove the `beforeEach` map-clearing loop; rewrite the re-initialization test as two independent-map assertions.
- **Chain tests** (`chain-tests.spec.ts`, `etherscan-instances.spec.ts`): Use the returned value in the async IIFEs. Add `.catch()` to both IIFEs so a chain-init failure exits instead of hanging Mocha's `--delay` mode. `etherscan-instances.spec.ts` is now wrapped in the same `--delay`/`run()` pattern as `chain-tests.spec.ts` (was previously relying on synchronous access to the old module-level map).
- **`test/chains/deployContracts.ts`**: Call `initializeSourcifyChains()` inside `main()` and use the returned map.

## Out of scope

Other review comments from #2709 (GCP-hosted URL, not importing `config` in `sourcify-chains.ts`, Dockerfile dedup, etc.) will be addressed separately.

### DIFF
--- a/services/server/test/helpers/ServerFixture.ts
+++ b/services/server/test/helpers/ServerFixture.ts
@@ -30,13 +30,14 @@ export class ServerFixture {
   readonly repositoryV1Path: string;
   readonly testS3Path: string = testS3Path;
   readonly testS3Bucket: string = testS3Bucket;
-  // Assigned in the before() hook below; safe to read inside it/beforeEach.
-  sourcifyChainsMap!: SourcifyChainMap;
 
   private _server?: Server;
 
   // Getters for type safety
   // Can be safely accessed in "it" blocks
+  get sourcifyChainsMap(): SourcifyChainMap {
+    return this.server.chainRepository.sourcifyChainMap;
+  }
   get sourcifyDatabase(): Pool {
     // sourcifyDatabase is just a shorter way to get databasePool inside SourcifyDatabaseService
     const _sourcifyDatabase = (
@@ -64,7 +65,11 @@ export class ServerFixture {
     this.repositoryV1Path = config.get<string>("repositoryV1.path");
 
     before(async () => {
-      this.sourcifyChainsMap = await initializeSourcifyChains();
+      // fixtureOptions_?.chains takes priority; fall back to fetching from the
+      // remote/local config. The same map is passed to both serverOptions.chains
+      // and sourcifyChainMap so both are always consistent.
+      const sourcifyChainsMap =
+        fixtureOptions_?.chains || (await initializeSourcifyChains());
 
       process.env.SOURCIFY_POSTGRES_PORT =
         process.env.DOCKER_HOST_POSTGRES_TEST_PORT || "5431";
@@ -83,7 +88,7 @@ export class ServerFixture {
         port: fixtureOptions_?.port || config.get<number>("server.port"),
         maxFileSize: config.get<number>("server.maxFileSize"),
         corsAllowedOrigins: config.get<string[]>("corsAllowedOrigins"),
-        chains: fixtureOptions_?.chains || this.sourcifyChainsMap,
+        chains: sourcifyChainsMap,
         solc: new SolcLocal(config.get("solcRepo"), config.get("solJsonRepo")),
         vyper: new VyperLocal(config.get("vyperRepo")),
         fe: new FeLocal(config.get("feRepo")),
@@ -96,7 +101,7 @@ export class ServerFixture {
       this._server = new Server(
         serverOptions,
         {
-          sourcifyChainMap: this.sourcifyChainsMap,
+          sourcifyChainMap: sourcifyChainsMap,
           solcRepoPath: config.get("solcRepo"),
           solJsonRepoPath: config.get("solJsonRepo"),
           vyperRepoPath: config.get("vyperRepo"),
@@ -186,7 +191,7 @@ export class ServerFixture {
   }
 
   resetChainHealthStates(): void {
-    const chains = this.server.chainRepository.sourcifyChainMap;
+    const chains = this.sourcifyChainsMap;
     for (const chain of Object.values(chains)) {
       for (const rpc of chain.rpcs) {
         if (rpc.health) {
--- a/services/server/test/helpers/ServerFixture.ts
+++ b/services/server/test/helpers/ServerFixture.ts
@@ -3,10 +3,7 @@ import { resetDatabase } from "../helpers/helpers";
 import type { ServerOptions } from "../../src/server/server";
 import { Server } from "../../src/server/server";
 import config from "config";
-import {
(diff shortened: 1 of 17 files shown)


services/server/test/unit/verificationWorker.spec.ts
```diff
@@ -39,7 +36,7 @@ describe("verificationWorker", function () {
   const piscinaSandbox = Sinon.createSandbox();
 
   before(async () => {
-    await initializeSourcifyChains();
+    const sourcifyChainsMap = await initializeSourcifyChains();
```
manuelwedler, line 39: this still involves the network call, right? will this be addressed separately?

services/server/test/helpers/ServerFixture.ts
```diff
@@ -33,6 +30,8 @@ export class ServerFixture {
   readonly repositoryV1Path: string;
   readonly testS3Path: string = testS3Path;
   readonly testS3Bucket: string = testS3Bucket;
+  // Assigned in the before() hook below; safe to read inside it/beforeEach.
+  sourcifyChainsMap!: SourcifyChainMap;
```
manuelwedler: Actually not necessary to introduce this property. The chain map can also be accessed via `fixture.server.chainRepository.sourcifyChainMap`. Maybe a bit bulky to write this everywhere, so also fine to keep the property. You can decide.

services/server/test/helpers/ServerFixture.ts
```diff
@@ -84,7 +83,7 @@ export class ServerFixture {
         port: fixtureOptions_?.port || config.get<number>("server.port"),
         maxFileSize: config.get<number>("server.maxFileSize"),
         corsAllowedOrigins: config.get<string[]>("corsAllowedOrigins"),
-        chains: fixtureOptions_?.chains || sourcifyChainsMap,
+        chains: fixtureOptions_?.chains || this.sourcifyChainsMap,
```
manuelwedler: not a problem of this PR, but a problem in general:

below we always pass the `this.sourcifyChainsMap` to the verification service, but it should also prioritize the fixtureOptions.chains.

I think in general it would be better to remove the sourcifyChainMap from the VerificationServiceOptions and instead pass a separate argument from the Server constructor to the VerificationService in order to avoid such mistakes.

services/server/test/unit/utils/contract-creation-util.spec.ts
```diff
@@ -4,19 +4,18 @@ import {
...
-} from "../../../src/sourcify-chains";
+import { initializeSourcifyChains } from "../../../src/sourcify-chains";
 import { ChainRepository } from "../../../src/sourcify-chain-repository";
 import type { FetchContractCreationTxMethod } from "@ethereum-sourcify/lib-sourcify";
 import sinon from "sinon";
 import { SourcifyChain } from "@ethereum-sourcify/lib-sourcify";
 import { findContractCreationTxByBinarySearch } from "../../../src/server/services/utils/contract-creation-util";
 
 describe("contract creation util", function () {
+  let sourcifyChainsMap: Awaited<ReturnType<typeof initializeSourcifyChains>>;
```
manuelwedler: The type of this variable should just be `SourcifyChainMap`.

services/server/test/unit/verificationWorker.spec.ts
kuzdogan: Yes

kuzdogan: But also we have a lot of integration tests like Etherscan tests or some that involve verifying a contract on a specific chain. How do we go about it? Our tests overall are not too isolated

services/server/test/helpers/ServerFixture.ts
kuzdogan: Do you think it should be in this PR? The change should be straightforward?

services/server/test/helpers/ServerFixture.ts
kuzdogan: See https://github.com/argotorg/sourcify/pull/2749/changes/c16794e84dd5a393c5b56cfb180ac790279deb6a

services/server/test/helpers/ServerFixture.ts
kuzdogan: See https://github.com/argotorg/sourcify/pull/2749/changes/c16794e84dd5a393c5b56cfb180ac790279deb6a about prioritizing `fixtureOptions.chains`

Above comment for removing the `sourcifyChainMap` from `VerificationServiceOptions`

services/server/test/helpers/ServerFixture.ts
manuelwedler: Now it's clean :)

services/server/test/unit/verificationWorker.spec.ts
manuelwedler: Most of them should be in a separate suite like the contract creation fetching one. But generally I think we could improve in that sense a lot. Maybe something for the near future if we want to speed up our tests.

This case here only would affect all tests that use the ServerFixture and is easily solvable.

services/server/test/helpers/ServerFixture.ts
manuelwedler: doesn't have to be in this PR, but should be straight forward, yes. I would simply add the chain map as a separate argument to the VerificationService constructor.

manuelwedler (approved):

PR #2754, opened 2026-04-22
Fix OOM error in CircleCI

marcocastignoli:
This PR fixes the out of memory error in CI server-test runs. It addresses three separate problems:

1. **Clean `VerificationService` after each test in its tests.** This is the real core of the problem. Each unit test creates a new VerificationService() which spins up a Piscina worker pool (minThreads=2), but never calls close(). After all 4 tests, 8 zombie worker threads remain alive for the rest of the suite, accumulating hundreds of MBs and eventually OOM-killing the CI container (exit 137). The fix calls verificationService.close() in afterEach to destroy the pool and free the threads.
2. **Gracefully close storage services with proper shutdown order.**  We've always missed gracefully closing the database storage services. I also make sure to close the Sourcify database storage service as last since other services are writing to it.
3. **Mock contracts insertion in should handle pagination.** In `file.spec.ts` we deploy and verify 15 contracts just to test pagination, I replaced it with inserting into database mocked contracts

See [how the latest CircleCI](https://app.circleci.com/pipelines/github/argotorg/sourcify/10925/workflows/bc98252d-4a76-4567-9e9b-a67d042991fe/jobs/63973/resources) run has a very low impact on memory.

### DIFF
--- a/services/server/src/server/services/StorageService.ts
+++ b/services/server/src/server/services/StorageService.ts
@@ -352,13 +352,26 @@ export class StorageService {
 
   public async close() {
     logger.info("Gracefully closing storage services");
+
+    // Close all services except SourcifyDatabase first because
+    // it is used by other services and should be closed last
+    // to avoid errors during the closing process
+    const sourcifyDatabase =
+      this.rwServices[RWStorageIdentifiers.SourcifyDatabase];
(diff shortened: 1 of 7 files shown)


manuelwedler (approved):
Looks good. Thanks for finding this!

PR #2779, opened 2026-05-08
feat(server): optional remote chain registry to augment sourcify-chains

mirshko:
Adds an opt-in HTTP source for chain configuration that is merged with the
bundled chains.json + sourcify-chains-default.json at server boot. Static
config always wins; remote entries are only added for chainIds that aren't
already known locally. The /chains endpoint shape is unchanged.

- New chain-registry.ts: fetchChainRegistry() using native fetch, hand-written
  type guards, winston logger, AbortSignal.timeout. Graceful failure: returns
  empty additions on network/parse/non-2xx errors. Accepts a chainlist.org-
  shaped payload with an optional sourcifyExtension field for full parity
  with sourcify-chains-default.json (etherscanApi, fetchContractCreationTxUsing,
  typed RPCs).
- sourcify-chains.ts: extracted assembleSourcifyChainsMap() from the existing
  top-level merge logic. Exported buildSourcifyChainsMap(opts) merges remote
  additions before assembly. The synchronous sourcifyChainsMap export is
  preserved so existing test/integration imports keep working.
- cli.ts: wraps Server construction in async bootstrap(); awaits the chain
  map and threads it through getEtherscanApiKeyForEachChain(map).
- config/default.js: chainRegistry.{url, required, timeoutMs}. Bearer token
  lives in process.env.CHAIN_REGISTRY_AUTH_TOKEN.
- .env.dev: documents CHAIN_REGISTRY_URL and CHAIN_REGISTRY_AUTH_TOKEN.
- 9 unit tests in chain-registry.spec.ts using mocha+chai+nock.

### DIFF
--- a/services/server/.env.dev
+++ b/services/server/.env.dev
@@ -48,6 +48,17 @@ NODE_LOG_LEVEL=
 # Configuration for the Piscina worker pool
 # How long one worker can be idle before it is shutdown in seconds. If not set, uses 30 s as default.
 WORKER_IDLE_TIMEOUT=
+
+# Optional remote chain registry. When set, the server fetches additional
+# chains at boot and merges them with the bundled sourcify-chains-default.json.
+# Static chains/extensions always take precedence over remote entries.
+# Payload format is chainlist.org-style: a JSON object keyed by chainId where
+# each entry has { chainId, rpc[], name?, hidden?, ... } and may optionally
+# include a `sourcifyExtension` field with full sourcify-chains-default.json
+# semantics (etherscanApi, fetchContractCreationTxUsing, typed RPCs, etc.).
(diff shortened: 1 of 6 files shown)


mirshko:
Hi @kuzdogan 

Would love your thoughts on this PR, the idea is that for running this ourselves we'd need a way to provide a non-static ingestion of supported chains as they are added / removed every so often, instead of manually maintaining the json, we could tell the API to restart and re-ingest the supported chains from an optionally auth'd endpoint.

lemme know, happy to discuss further.

kuzdogan:
Hey thanks! this looks very similar to #2709 :) right? Merging that PR today

kuzdogan:
Also curious about the use case. Can you reach out on telegram? Same username as here

PR #2780, opened 2026-05-11
feat: add hidden chain config option to filter chains from public listings

mirshko:
## Summary

Adds an optional `hidden` boolean field to `SourcifyChainExtension` that can be set in either the hardcoded `sourcify-chains.json` or the remote chains config (now supported via `chains.remoteUrl`).

When a chain is marked `hidden: true`:
- It is omitted from `GET /chains`
- Its entries are filtered out of `GET /v2/contract/all-chains/:address`

Verification and per-chain lookups by explicit `chainId` continue to work — `hidden` only suppresses inclusion in listings that return chains without an explicit chainId request.

## Changes

- `packages/lib-sourcify`: added `hidden?: boolean` to `SourcifyChainExtension`; `SourcifyChain` now exposes `readonly hidden: boolean` (defaults to `false`)
- `services/server/src/sourcify-chains.ts`: passes `hidden` through from local/remote config when constructing each `SourcifyChain`
- `services/server/src/server/routes.ts`: `GET /chains` filters out hidden chains
- `services/server/src/server/apiv2/lookup/lookup.handlers.ts`: `GET /v2/contract/all-chains/:address` filters out results from hidden chains

## Endpoints reviewed

Other endpoints that touch chains either accept explicit `chainId` parameters (verification routes, `/v2/contracts/:chainId`, `/v2/contract/:chainId/:address`, v1 `/check-by-addresses?chainIds=...`, etc.) or are internal-only logging — none required changes.

### DIFF
--- a/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
+++ b/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
@@ -47,6 +47,12 @@ export class SourcifyChain {
   /** Whether the chain supports tracing, used for fetching the creation bytecode for factory contracts */
   readonly traceSupport?: boolean;
   readonly supported: boolean;
+  /**
+   * When true, the chain is hidden from public listings such as the /chains
+   * endpoint and the /v2/contract/all-chains/{address} response. Verification
+   * for the chain still works when its chainId is explicitly requested.
+   */
+  readonly hidden: boolean;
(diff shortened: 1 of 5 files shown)


kuzdogan (approved):
lgtm :) thanks. 

Tests are failing because with the new external chains json fetching, these chains have DRPC and QUICKNODE API keys and those are skipped for  external PRs. Will do a separate fix for it.

PR #2783, opened 2026-05-11
Upgrade lerna and hardhat to major versions and override packages in package.json

marcocastignoli:
## Summary

* Bumped hardhat to 3.4.5. Added a new internal package `@ethereum-sourcify/test-helpers` that centralizes the hardhat runner config (`hardhat-runner/`) and the shared `startHardhatNetwork` / `stopHardhatNetwork` helpers, so each other package (lib-sourcify, monitor, server) no longer duplicates them.
* Bumped lerna to 9.0.7
* Patched transitive vulnerabilities via the `package.json` `overrides` field (axios, yaml, brace-expansion, follow-redirects, diff).


packages/test-helpers/hardhat-runner/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "sourcify-hardhat-runner",
```
manuelwedler, line 2: Can you explain why we need this package inside the test-helper package? Can't we put the config simply one level up?

marcocastignoli: From Claude:

> Tried the refactor and it doesn't work in Hardhat 3, sadly. The .mjs extension isn't enough — Hardhat 3 walks up to the project's package.json and refuses to run unless that file has "type": "module". With the config at packages/test-helpers/ root it fails with `Hardhat only supports ESM projects. Please make sure you have "type": "module" in your package.json.`
> Adding "type": "module" to test-helpers/package.json itself isn't viable: the helper is consumed via require from CJS test files (lib-sourcify, monitor, server tests run under ts-node/register in CJS mode), and a require() of an ESM module errors with ERR_REQUIRE_ESM.
> The nested hardhat-runner/ directory exists purely to confine the ESM marker to the dir Hardhat treats as the project root, so the helper itself stays CJS. Happy to revisit if there's a cleaner trick I'm missing, but this seems to be the least-bad workaround for Hardhat 3's ESM-project enforcement.

manuelwedler: okay sounds fine to me

manuelwedler (approved):

PR #2797, opened 2026-05-21
Release new package versions

kuzdogan:
🚨 **CRITICAL: USE MERGE COMMIT, NOT SQUASH** 🚨

**IMPORTANT:** This PR contains release commits with git tags that point to specific commit hashes in this branch. 

**YOU MUST USE "CREATE A MERGE COMMIT" WHEN MERGING THIS PR.**

❌ **DO NOT** use "Squash and merge" - this will break the git tags and release history.
✅ **DO** use "Create a merge commit" to preserve the commit history and tag references.

The tags in this branch point to specific commits, and squashing would invalidate these references.

### DIFF
--- a/services/server/CHANGELOG.md
+++ b/services/server/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file.
 
+## sourcify-server@3.15.0 - 2026-05-21
+
+- feat: fetch chain config from remote URL and remove HTML scraping
+  #2709
+- Refactor Etherscan tests to fully mock the requests
+- fix(server): throw on chains with no usable RPCs, add DRPC_API_KEY
+  #2782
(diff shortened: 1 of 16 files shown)


services/server/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file.
 
+## sourcify-server@3.15.0 - 2026-05-21
```
manuelwedler, line 5: this one is missing:

fix(server): validate afterMatchId as numeric in v2 contracts list (#2793)

manuelwedler (approved):
you can add the missing changelog line if you like. still approving.

services/server/CHANGELOG.md
kuzdogan: I don't think it's significant

PR #2803, opened 2026-05-26
clear deployer and blockNumber when creationTransactionHash validation fails

marcocastignoli:
When a user provides a `creationTransactionHash` that doesn't match the contract address, `deployer` and `blockNumber` from that wrong transaction were still stored in the database. This happened because these fields were assigned before the address validation in `getContractCreationBytecodeAndReceipt`, and the catch block only cleared `creatorTxHash`.

The fix clears `blockNumber` and `deployer` in the catch block alongside `creatorTxHash`. A regression test is added that asserts all three fields are null in `contract_deployments` when a wrong creation tx hash is provided.


manuelwedler (approved):
looks good

PR #2805, opened 2026-05-27
Add chain_id column to sourcify_matches

marcocastignoli:
## Summary

First step toward fixing the slow `/v2/contracts/{chainId}` pagination reported in [#2111](https://github.com/argotorg/sourcify/issues/2111). The root cause is that pagination orders by `sourcify_matches.id` but filters by `chain_id` two joins away on `contract_deployments`; with no way to drive the scan from `chain_id`, PostgreSQL walks millions of rows per page on chains with sparse matches.

The plan is to denormalize `chain_id` onto `sourcify_matches` and add a composite `(chain_id, id DESC)` index. This PR lays the groundwork:

- New migration `20260527081526_add_chain_id_to_sourcify_matches.sql` adds the column as a nullable `bigint`. Brief AccessExclusive lock, no table rewrite (PG ≥ 11), no inline backfill — safe on the ~30M-row production table.
- `insertSourcifyMatch` and `updateSourcifyMatch` now write `chain_id` on every call, sourcing it from `verification.chainId` (already in scope at both call sites in `SourcifyDatabaseService.ts`).
- Type updates: `Tables.SourcifyMatch.chain_id` (server) and `PgSourcifyMatch.chain_id` (database pg-types).

## Follow-up PRs

1. Standalone batched backfill script in `services/database/scripts/` for the existing ~30M rows (each batch its own short transaction, idempotent, resumable).
2. `CREATE INDEX CONCURRENTLY` migration on `(chain_id, id DESC)`.
3. Switch `getSourcifyMatchesByChain` to filter on `sourcify_matches.chain_id`.
4. Promote the column to `NOT NULL` via `CHECK NOT VALID` + `VALIDATE CONSTRAINT`.

## Notes

- `sourcify-database.sql` schema dump is **not** regenerated in this PR — needs to be done by running `npm run migrate:up` locally and committing the result alongside.
- Existing rows remain NULL until the backfill script lands; the read path still filters via `contract_deployments.chain_id` join, so this PR is functionally a no-op for users.

### DIFF
--- a/services/database/migrations/20260527081526_add_chain_id_to_sourcify_matches.sql
+++ b/services/database/migrations/20260527081526_add_chain_id_to_sourcify_matches.sql
@@ -1,0 +1,14 @@
+-- migrate:up
+
+-- Add chain_id column to sourcify_matches to enable a composite index
+-- (chain_id, id DESC) for efficient pagination of /v2/contracts/{chainId}.
+-- See: https://github.com/argotorg/sourcify/issues/2111
+--
+-- Nullable for now: existing rows are backfilled via
+-- services/database/scripts/backfill-sourcify-match-chain-id.mjs.
+-- A subsequent migration will set NOT NULL after the backfill completes.
+ALTER TABLE sourcify_matches ADD COLUMN chain_id bigint;
+
+-- migrate:down
+
+ALTER TABLE sourcify_matches DROP COLUMN chain_id;
--- a/services/database/pg-types.ts
+++ b/services/database/pg-types.ts
@@ -1,7 +1,7 @@
 // Type definitions for rows fetched from the database with the pg client
 // E.g. pg fetches numbers as strings
 
-import {
+import type {
   Transformation,
   TransformationValues,
 } from "@ethereum-sourcify/lib-sourcify";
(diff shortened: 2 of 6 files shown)


services/database/migrations/20260527081526_add_chain_id_to_sourcify_matches.sql
```diff
@@ -0,0 +1,14 @@
+-- migrate:up
+
+-- Add chain_id column to sourcify_matches to enable a composite index
+-- (chain_id, id DESC) for efficient pagination of /v2/contracts/{chainId}.
+-- See: https://github.com/argotorg/sourcify/issues/2111
+--
+-- Nullable for now: existing rows are backfilled via
+-- services/database/scripts/backfill-sourcify-match-chain-id.mjs.
+-- A subsequent migration will set NOT NULL after the backfill completes.
+ALTER TABLE sourcify_matches ADD COLUMN chain_id bigint;
```
manuelwedler, line 10: Would it be cheaper to add the index already here?

manuelwedler: probably wouldn't matter much since we use `concurrently` in the other PR

manuelwedler (approved):

PR #2806, opened 2026-05-27
Add backfill script for sourcify_matches.chain_id

marcocastignoli:
## Summary

- Adds `services/database/schema-updates/post-v2.12-to-v2.13-upgrade.mjs`, a one-time batched UPDATE that fills the new `sourcify_matches.chain_id` column from `contract_deployments.chain_id`.
- Companion to #2805, which adds the column itself (nullable). Run this after upgrading to `sourcify-database@2.13.x` and before applying the follow-up `NOT NULL` migration.

## Why a separate script instead of inline in the migration?

The production `sourcify_matches` table has ~30M rows. An inline UPDATE inside a dbmate migration would run in a single transaction, holding row locks and pinning the cluster's xmin horizon for the full duration (10–60 min). Splitting the backfill into many short autocommit batches keeps locks brief, lets autovacuum keep up, and is resumable on interrupt. See [#2111](https://github.com/argotorg/sourcify/issues/2111) for the full context.

## Usage

```bash
cd services/database/schema-updates
node post-v2.12-to-v2.13-upgrade.mjs --batch-size=50000 --sleep-ms=50

### DIFF
--- a/services/database/schema-updates/post-v2.12-to-v2.13-upgrade.mjs
+++ b/services/database/schema-updates/post-v2.12-to-v2.13-upgrade.mjs
@@ -1,0 +1,246 @@
+/**
+ * Backfills sourcify_matches.chain_id from contract_deployments.chain_id.
+ *
+ * Required after upgrading to sourcify-database@2.13.x, which adds the new
+ * chain_id column to sourcify_matches but leaves it nullable so existing rows
+ * can be filled out-of-band. Once this script has finished and no NULL rows
+ * remain, the follow-up migration that promotes the column to NOT NULL is
+ * safe to apply.
+ *
+ * The script is idempotent and resumable: it only touches rows where
+ * chain_id IS NULL and walks the table in primary-key ranges, so each batch
+ * runs in its own short transaction. Safe to Ctrl+C and re-run.
+ *
+ * See: https://github.com/argotorg/sourcify/issues/2111
+ *
+ * Environment Variables Required:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: services/database/schema-updates/post-v2.12-to-v2.13-upgrade.mjs [hunk 1]
```diff
@@ -0,0 +1,246 @@
+/**
```
manuelwedler, line 1: We should document in the README that it is required to run this after the related migration

marcocastignoli: Do you think README is the best place? I was thinking the Release description / changelog would be better. I don't think sourcify maintainers will read the readme

manuelwedler: I think it should be in both places. I don't think you will read through all release notes when you upgrade like half a year later. also we might need to add more such scripts, so it makes sense to have a central place where they are documented.

marcocastignoli: Done :)

manuelwedler: I added a small note in the database migrations section to make the new section more easily findable.

manuelwedler (approved):
looks good

PR #2817, opened 2026-06-06
fix(lib-sourcify): support legacy Vyper immutable tails

banteg:
## Summary

- Extend `VyperCompilation.immutableReferences` so Vyper `0.3.1 <= version < 0.3.10` can produce the same synthetic tail reference shape that Sourcify already uses for Vyper `>= 0.3.10`.
- Derive legacy immutable tail length from Vyper compiler IR instead of reconstructing source-level type sizes in Sourcify.
- Keep verification on one immutable path: `extractImmutablesTransformation` receives one `ImmutableReferences` object and applies the existing Vyper insert transformation only after runtime tail sanity checks pass.
- Apply those Vyper tail checks uniformly to legacy and modern Vyper immutables.

## Why

Vyper added `immutable(...)` in `0.3.1`, but it did not add the newer initcode/runtime layout metadata, including immutable section size, until `0.3.10` via [vyperlang/vyper#3584](https://github.com/vyperlang/vyper/pull/3584).

For affected legacy contracts, the deployed runtime can be:

```text
recompiled_runtime || immutable_tail
```

The old Vyper runtime metadata trailer is still inside `recompiled_runtime`, and the immutable payload is appended after it. Because pre-`0.3.10` metadata does not tell Sourcify how large that payload should be, these contracts fall through to plain bytecode comparison and fail with a deployed-vs-recompiled mismatch.

Vyper `>= 0.3.10` immutables were already supported through CBOR auxdata decoding. This PR keeps that modern path and adds legacy reference inference for `0.3.1` through `0.3.9`.

Concrete failing Vyper `0.3.7` examples from the observed bytecode-mismatch set:

- `0x0164df8d03e9e337497b3ce49d7736696054b92b`
- `0x19198c82f57f4613ef3b7fea2d268b44a4fc33a9`
- `0x1d43e85ff803938cc8f9d13b8c194a27bb59f7ef`
- `0x1d47ccdd400e56ec3517ac3f14a01601e4192ae6`
- `0x20c8602f5df730f93e298a0492777412935f37aa`
- `0x2106c2d7d321e434c1e37697a22575b497c32272`

For example, `0x0164df8d03e9e337497b3ce49d7736696054b92b` recompiles to a `248` byte runtime with Vyper `0.3.7`, while the deployed runtime is `280` bytes. The extra `32` bytes are the appended immutable value:

```text
0x000000000000000000000000216ce6e49e2e713e41383ba4c5d84a0d36189640
```

The local corpus also has exact `0.3.1` sources with active `immutable(...)` declarations, so this PR keeps the Vyper feature floor instead of dropping `0.3.1` support.

## Chosen Approach

Compilation now produces immutable references for both Vyper eras:

1. For `>= 0.3.10`, Sourcify keeps decoding `immutableSize` from Vyper creation-bytecode CBOR auxdata.
2. For `0.3.2 <= version < 0.3.10`, Vyper Standard JSON emits structured IR. The deploy node contains the compiler-resolved immutable size at `deploy[2]`; contracts without immutable data use `0`.
3. For `0.3.1`, Vyper Standard JSON emits old LLL text. The compiler-generated constructor return has the stable shape:

```text
[return, <memory_start>, [add, <immutable_tail_bytes>, _lllsz]]
```

4. Sourcify accepts a legacy immutable size only when exactly one positive, `32`-byte-aligned size signal is found.
5. `VyperCompilation.immutableReferences` returns one synthetic reference keyed by `"0"`, with `start` equal to the compiler runtime length and `length` equal to the compiler-derived immutable tail size.
6. Verification uses that normal reference object and appends the observed on-chain immutable bytes before bytecode comparison.

Using IR avoids hand-maintaining Vyper type-size logic in Sourcify. It also handles compiler-resolved constants in immutable array bounds, structs, arrays, and other supported immutable layouts because Vyper has already lowered the final immutable section size.

## Safety Conditions

The IR-derived legacy size must be positive, word-aligned, and unambiguous. Then, during verification, all Vyper immutable references are accepted only when:

- The deployed runtime is strictly longer than the populated recompiled runtime.
- The deployed runtime starts with the populated recompiled runtime.
- The appended byte length exactly equals the total compiler-derived immutable reference length.
- The references describe a contiguous tail starting at the end of the recompiled runtime.

If any check fails, Sourcify records no immutable transformation and keeps the existing bytecode comparison behavior.

This also addresses false-positive risk. A no-immutable structured IR signal (`deploy[2] = 0`) is ignored. Even if a future compiler shape accidentally produced a positive false signal, an already matching runtime has no appended tail, so the verifier records no phantom immutable. A prefix drift or wrong tail length also records no transformation, and the final bytecode comparison remains the correctness backstop.

## Limitations

- This handles append-only legacy Vyper immutable tails. It does not normalize unrelated legacy Vyper codegen instability, such as function/helper ordering differences in some `0.3.7` outputs. See [vyperlang/vyper#3369](https://github.com/vyperlang/vyper/issues/3369) for one known source of ordering instability.
- The immutable transformation accepts deployed immutable bytes as observed runtime data, matching Sourcify's existing Vyper runtime-side semantics. It does not prove constructor assignment semantics.
- The `0.3.1` path necessarily parses text LLL because that compiler version does not expose structured IR through Standard JSON. The parser is intentionally narrow and fails closed on ambiguity.

## Checks

- Focused Vyper immutable tests: `19 passing`
- `Transformations.spec.ts`: `17 passing`
- Prettier check on touched files
- ESLint on touched files
- `lerna run build --scope=@ethereum-sourcify/lib-sourcify --include-dependencies`

### DIFF
--- a/packages/lib-sourcify/src/Compilation/VyperCompilation.ts
+++ b/packages/lib-sourcify/src/Compilation/VyperCompilation.ts
@@ -88,7 +88,9 @@ function isAstNode(node: unknown): node is VyperAstNode {
   return node !== null && typeof node === 'object';
 }
 
-function getImmutableTypeAnnotation(node: VyperAstNode): VyperAstNode | undefined {
+function getImmutableTypeAnnotation(
+  node: VyperAstNode,
+): VyperAstNode | undefined {
   const annotation = node.annotation;
   if (!isAstNode(annotation)) {
     return undefined;
@@ -174,7 +176,9 @@ function getTypeByteLength(
     if (subtypeByteLength === undefined) {
       return undefined;
     }
-    return DYNAMIC_ARRAY_OVERHEAD_WORDS * WORD_SIZE + maxLength * subtypeByteLength;
+    return (
+      DYNAMIC_ARRAY_OVERHEAD_WORDS * WORD_SIZE + maxLength * subtypeByteLength
+    );
   }
 
   const length = getSubscriptLength(annotation);
@@ -209,7 +213,6 @@ export function returnLegacyVyperImmutableReferences(
 
   const structs = collectStructDefinitions(ast);
   const runtimeByteLength = runtimeBytecode.substring(2).length / 2;
-  const immutableReferences: ImmutableReferences = {};
   let immutableOffset = 0;
 
   for (const node of ast.body.filter(isAstNode)) {
@@ -219,25 +222,29 @@ export function returnLegacyVyperImmutableReferences(
     }
 
     const immutableByteLength = getTypeByteLength(annotation, structs);
-    const immutableName = node.target?.id;
     if (
       immutableByteLength === undefined ||
       immutableByteLength <= 0 ||
-      typeof immutableName !== 'string'
+      typeof node.target?.id !== 'string'
     ) {
       return {};
     }
 
-    immutableReferences[immutableName] = [
-      {
-        length: immutableByteLength,
-        start: runtimeByteLength + immutableOffset,
-      },
-    ];
     immutableOffset += immutableByteLength;
   }
 
-  return immutableReferences;
+  if (immutableOffset === 0) {
+    return {};
+  }
+
+  return {
+    '0': [
+      {
+        length: immutableOffset,
+        start: runtimeByteLength,
+      },
+    ],
+  };
 }
 
 export function returnImmutableReferences(
--- a/packages/lib-sourcify/src/Compilation/VyperCompilation.ts
+++ b/packages/lib-sourcify/src/Compilation/VyperCompilation.ts
@@ -74,329 +74,94 @@ export function returnAuxdataStyle(
   return AuxdataStyle.VYPER;
 }
 
-type VyperAstNode = Record<string, any>;
-type VyperStructDefinitions = Record<string, VyperAstNode[]>;
-type VyperIntegerConstants = Record<string, number>;
-
 const WORD_SIZE = 32;
-const DYNAMIC_ARRAY_OVERHEAD_WORDS = 1;
+const LEGACY_VYPER_TEXT_IR_IMMUTABLE_RETURN =
+  /\[return,\s*(\d+),\s*\[add,\s*(\d+),\s*_lllsz\]\]/g;
 
-function ceil32(value: number): number {
-  return Math.ceil(value / WORD_SIZE) * WORD_SIZE;
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return value !== null && typeof value === 'object' && !Array.isArray(value);
 }
 
-function isAstNode(node: unknown): node is VyperAstNode {
-  return node !== null && typeof node === 'object';
-}
-
-function getImmutableTypeAnnotation(
-  node: VyperAstNode,
-): VyperAstNode | undefined {
-  const annotation = node.annotation;
-  if (!isAstNode(annotation)) {
-    return undefined;
-  }
-
-  const immutableCall =
-    annotation.ast_type === 'Call' && annotation.func?.id === 'immutable';
-
-  if (immutableCall && isAstNode(annotation.args?.[0])) {
-    return annotation.args[0];
-  }
-
-  if (node.is_immutable === true) {
-    return annotation;
-  }
-
-  return undefined;
-}
-
-function collectStructDefinitions(ast: VyperAstNode): VyperStructDefinitions {
-  const structs: VyperStructDefinitions = {};
-  for (const node of ast.body ?? []) {
-    if (
-      isAstNode(node) &&
-      node.ast_type === 'StructDef' &&
-      typeof node.name === 'string' &&
-      Array.isArray(node.body)
-    ) {
-      structs[node.name] = node.body.filter(isAstNode);
-    }
-  }
-  return structs;
-}
-
-function getAstIntegerValue(node: VyperAstNode): number | undefined {
-  const value = node.value ?? node.n;
-  if (Number.isSafeInteger(value)) {
-    return value;
-  }
-  return undefined;
-}
-
-function evaluateIntegerExpression(
-  node: unknown,
-  constants: VyperIntegerConstants,
-): number | undefined {
-  if (!isAstNode(node)) {
-    return undefined;
-  }
-
-  const integerValue = getAstIntegerValue(node);
-  if (integerValue !== undefined) {
-    return integerValue;
-  }
-
-  if (node.ast_type === 'Name' && typeof node.id === 'string') {
-    return constants[node.id];
-  }
-
-  if (node.ast_type === 'UnaryOp') {
-    const operand = evaluateIntegerExpression(node.operand, constants);
-    if (operand === undefined) {
-      return undefined;
-    }
-    if (node.op?.ast_type === 'USub') {
-      return -operand;
-    }
-    if (node.op?.ast_type === 'UAdd') {
-      return operand;
-    }
-    return undefined;
-  }
-
-  if (node.ast_type !== 'BinOp') {
-    return undefined;
-  }
-
-  const left = evaluateIntegerExpression(node.left, constants);
-  const right = evaluateIntegerExpression(node.right, constants);
-  if (left === undefined || right === undefined) {
-    return undefined;
-  }
-
-  let result: number | undefined;
-  switch (node.op?.ast_type) {
-    case 'Add':
-      result = left + right;
-      break;
-    case 'Sub':
-      result = left - right;
-      break;
-    case 'Mult':
-      result = left * right;
-      break;
-    case 'Pow':
-      result = left ** right;
-      break;
-    case 'Div':
-    case 'FloorDiv':
-      if (right === 0 || left % right !== 0) {
-        return undefined;
-      }
-      result = left / right;
-      break;
-    default:
-      return undefined;
-  }
-
-  if (Number.isSafeInteger(result)) {
-    return result;
-  }
-  return undefined;
-}
-
-function getConstantTypeAnnotation(
-  node: VyperAstNode,
-): VyperAstNode | undefined {
-  const annotation = node.annotation;
-  if (!isAstNode(annotation)) {
-    return undefined;
-  }
-
-  if (node.is_constant === true) {
-    return annotation;
-  }
-
-  const constantCall =
-    annotation.ast_type === 'Call' && annotation.func?.id === 'constant';
-  if (constantCall && isAstNode(annotation.args?.[0])) {
-    return annotation.args[0];
-  }
-
-  return undefined;
-}
-
-function collectIntegerConstantDefinitions(
-  ast: VyperAstNode,
-): VyperIntegerConstants {
-  const constants: VyperIntegerConstants = {};
-  let pendingConstants: VyperAstNode[] = ast.body
-    .filter(isAstNode)
-    .filter(
-      (node: VyperAstNode) =>
-        typeof node.target?.id === 'string' &&
-        getConstantTypeAnnotation(node) !== undefined &&
-        isAstNode(node.value),
-    );
-
-  while (pendingConstants.length > 0) {
-    const remainingConstants: VyperAstNode[] = [];
-    let madeProgress = false;
-
-    for (const node of pendingConstants) {
-      const value = evaluateIntegerExpression(node.value, constants);
-      if (value === undefined) {
-        remainingConstants.push(node);
-        continue;
-      }
-
-      constants[node.target.id] = value;
-      madeProgress = true;
-    }
-
-    if (!madeProgress) {
-      break;
-    }
-    pendingConstants = remainingConstants;
-  }
-
-  return constants;
-}
-
-function getSubscriptLength(
-  node: VyperAstNode,
-  constants: VyperIntegerConstants,
-): number | undefined {
-  const value = node.slice?.value;
-  if (typeof value?.value === 'number') {
-    return value.value;
-  }
-  if (typeof value?.n === 'number') {
-    return value.n;
-  }
-  return evaluateIntegerExpression(value, constants);
-}
-
-function getTypeByteLength(
-  annotation: VyperAstNode,
-  structs: VyperStructDefinitions,
-  constants: VyperIntegerConstants,
-): number | undefined {
-  if (annotation.ast_type === 'Name' && typeof annotation.id === 'string') {
-    const structMembers = structs[annotation.id];
-    if (structMembers !== undefined) {
-      let structByteLength = 0;
-      for (const member of structMembers) {
-        const memberByteLength = getTypeByteLength(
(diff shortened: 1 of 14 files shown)


kuzdogan:
Thanks for this PR, and for the really thorough write-up and live examples. The approach of recovering the immutable size from the `ir` output for pre-`0.3.10` contracts is genuinely clever — we weren't aware this was even possible for those versions, so this is a nice find.

The implementation itself looks solid and the safety conditions are carefully thought through. My main feedback is about **where** this logic lives rather than whether it works.

Effectively, what this introduces is a way to infer the **`immutableReferences`** — i.e. the *positions/lengths* of the immutables (not the values) — for legacy contracts. For non-legacy Vyper we already produce that in the **Compilation** layer (`VyperCompilation.returnImmutableReferences`), reading the size from the CBOR auxdata. As far as I can tell, the immutables sit in the same place for both legacy and non-legacy (a single region appended at the end of the runtime); the only difference is that non-legacy contracts expose the size explicitly while legacy ones require the IR heuristic to recover it.

Given that, I'd prefer to keep this inference in the Compilation layer too — right alongside the existing `returnImmutableReferences` logic — so that `VyperCompilation.immutableReferences` returns the same shape of `ImmutableReferences` object whether or not the contract is legacy. If we do that, the **transformation logic can stay identical** for legacy and non-legacy, and we wouldn't need the separate `legacyImmutableReferences` path, the `effectiveImmutableReferences` branch, or the extra params threaded into `extractImmutablesTransformation`.

On the length checks inside `inferLegacyVyperImmutableReferences` (expected vs. actual immutable length) — I like those and I think we should keep them, but applied uniformly to **all** Vyper immutables rather than legacy-only. One thing to note: that check compares against the *on-chain* tail length, so it needs the on-chain bytecode, which the Compilation layer doesn't have. So the natural split would be: Compilation produces the references (from CBOR or IR), and any on-chain sanity check lives at the verification layer and runs for every Vyper contract. (The final bytecode comparison already backstops a wrong inference — it just won't match — so the explicit check is mostly about keeping the recorded transformations honest rather than correctness.)

The bigger open question for me: **how reliable is reading the immutable size from the IR?** Concretely — is `deploy[2]` (structured IR) and the `[add, N, _lllsz]` return shape (0.3.1 text IR) a correct and stable signal across the full `0.3.1`–`0.3.9` range? In particular, can it ever **false-positive** — infer a non-zero immutable region for a contract that actually has none — and if so, do we need to guard against that explicitly? If a false positive is possible, I'd want to understand what happens to a no-immutable contract that otherwise matches exactly, and whether we'd end up recording a phantom immutable. That feeds directly into whether the on-chain length/sanity check is "nice to have" or "required."

If you'd like, I'm happy to implement these changes (moving the inference into `VyperCompilation` and collapsing the two reference paths into one) and you can have a look at it — most of the moving parts already exist, it's mostly relocation. Whatever's easier for you.

banteg:
Thanks again for the detailed feedback. I pushed a follow-up in `54d9bb90` that applies the split you suggested:

- Moved legacy Vyper immutable reference inference into the compilation layer. `VyperCompilation.immutableReferences` now returns the same synthetic tail-reference shape for legacy IR-derived sizes and modern CBOR-derived sizes.
- Removed the separate `legacyImmutableReferences` verification path and the extra params threaded through `extractImmutablesTransformation`.
- Kept the on-chain sanity checks in verification, but now applies them uniformly to all Vyper immutable references: runtime must be longer, prefix-identical, have an exact tail-length match, and references must describe a contiguous tail from the end of the recompiled runtime.
- Tightened the false-positive story: zero structured IR size is ignored, ambiguous/invalid IR signals fail closed, and even a hypothetical positive false signal records no phantom immutable if the bytecodes already match or the prefix/tail length does not line up.

I also updated the PR description to reflect the new architecture and added tests for the no-phantom, zero-size, oversized-tail, and prefix-drift cases.

kuzdogan:
Quick follow-up on these commits:

The extra validation step (`getValidatedVyperImmutableReferences`) turned out not to be needed. The final bytecode comparison already rejects a wrong body or wrong length, so the only thing actually worth guarding is the phantom case — where the IR heuristic tells us there's an immutable but the on-chain code doesn't actually have the tail. A simple `immutableValue.length` check covers that, so I replaced the whole validation step with it.

With that, the Vyper Transformations and immutables flow stays mostly the same as before — the only real addition is one extra branch in `returnImmutableReferences` for `>=0.3.1` (legacy) when extracting the immutable values.

I also pulled the legacy immutable helpers out into their own file (`legacyVyperImmutablesHelpers.ts`) just for cleanliness.

kuzdogan (approved):
lgtm after small changes, thank you! Will still need another eye from the team

packages/lib-sourcify/src/Compilation/legacyVyperImmutablesHelpers.ts
```diff
@@ -0,0 +1,113 @@
...
+function getLegacyVyperImmutableLengthFromIr(
+  ir: VyperIROutput | undefined,
+): number | undefined {
+  if (ir === undefined) {
+    return undefined;
+  }
+  const lengths =
+    typeof ir === 'string'
+      ? extractTextIrImmutableLengths(ir)
+      : collectStructuredIrImmutableLengths(ir);
```
marcocastignoli: I had to read through the whole code to understand that this is the place in which there is a difference between 0.3.1 and 0.3.2

We should make it clear with a comment.
```
const lengths =
  typeof ir === 'string'
    ? extractTextIrImmutableLengths(ir)        // ← 0.3.1: text LLL
    : collectStructuredIrImmutableLengths(ir); // ← 0.3.2–0.3.9: structured IR tree
```

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
```diff
@@ -79,12 +84,20 @@ export function returnImmutableReferences(
   creationBytecode: string,
   runtimeBytecode: string,
   auxdataStyle: AuxdataStyle,
+  compilerOutput?: VyperOutput,
+  compilationTarget?: CompilationTarget,
```
manuelwedler, line 88: I don't like how this function changes its behavior dependent if these parameters are passed or not. I think they should be required and we should adapt `PreRunCompilation`.

marcocastignoli (changes requested):
Looks good, thanks! I think we should add this small comment to make things more clear. An alternative would be to separate completely the concerns of 0.3.1 vs 0.3.2–0.3.9 directly in `returnFixedVyperVersion` by passing the right function or creating a specific function for 0.3.1 vs 0.3.2–0.3.9.

But the comment is good enough.

```
  if (gte(compilerVersion, '0.3.10')) {
    try {
      const { immutableSize } = decode(
        creationBytecode,
        auxdataStyle,
      ) as VyperDecodedObject;
      if (
        immutableSize !== undefined &&
        isValidImmutableLength(immutableSize)
      ) {
        immutableReferences = {
          '0': [
            {
        creationBytecode: creationBytecode,
      });
    }
   } else if (gte(compilerVersion, '0.3.2') && compilationTarget !== undefined) {
    immutableReferences = returnLegacyVyperImmutableReferences(
      collectStructuredIrImmutableLengths, // here
      compilerOutput,
      compilationTarget,
      runtimeBytecode,
    );
  } else if (gte(compilerVersion, '0.3.1') && compilationTarget !== undefined) {
    immutableReferences = returnLegacyVyperImmutableReferences(
      extractTextIrImmutableLengths, // here
      compilerOutput,
      compilationTarget,
      runtimeBytecode,
    );
  }
```

packages/lib-sourcify/src/Compilation/VyperCompilation.ts
kuzdogan: Hmm yes it's in fact passed always `get immutableReferences()` here. So we should make it required.

But then for the PreRunCompilation how would it work? I guess it can't handle that case because we don't store the `.ir` output from the Vyper compiler.

marcocastignoli (approved):

banteg:
Follow-up pushed in `c9783aa1` after the Claude Fable 5 review surfaced a replay/storage gap.

What changed:
- Persist non-empty Vyper immutable references into runtime code artifacts instead of dropping them for Vyper. This lets DB-backed similarity/pre-run verification reuse the references later, even when reconstructed `std_json_output` does not include `ir`.
- Updated `PreRunCompilation` to prefer stored Vyper immutable references and fall back to the existing IR/CBOR derivation when full compiler output is available.
- Added Vyper deployed-bytecode `immutableReferences` to the Vyper output type so the stored artifact shape is typed.
- Added coverage for real Vyper `0.3.1` text IR, real `0.3.2` structured IR, ambiguous structured IR, pre-run outputs without IR, and server serialization of Vyper immutable references.
- Updated the existing Vyper constructor/immutables integration fixture to expect the now-stored runtime immutable reference.

Validation run locally:
- `packages/lib-sourcify/test/Compilation/VyperCompilation.spec.ts`: 31 passing outside the sandbox. The sandboxed run had the known PyInstaller semaphore failure on the existing Vyper 0.4.1 case; rerun outside the sandbox passed.
- `packages/lib-sourcify/test/Verification/Transformations.spec.ts`: 17 passing.
- `services/server/test/unit/utils/database-util.spec.ts`: 1 passing.
- TypeScript builds for `compilers-types`, `lib-sourcify`, and `services/server` passed.
- Prettier, ESLint, and `git diff --check` passed.

banteg:
Ran an end-to-end validation of this PR on the highest-profile contract it unblocks: **crvUSD** (`0xf939E0A03FB07F59A73314E73794Be0E57ac1b4E`, vyper 0.3.7), which has been stuck at `creationMatch: match, runtimeMatch: null` on production since 2025-01-08.

Driving `Verification.verify()` from this branch directly against mainnet (stock `0.3.7+commit.6020b8bb` binary, creator tx `0x5f501b4e420485ac3a18be3b88ad73413ac17ee45c776bb48cfd26902585e9fd`):

```
status: {"runtimeMatch":"partial","creationMatch":"partial"}
```

with every byte of both on-chain bytecodes accounted for:

- **runtime**: `{type: insert, reason: immutable, offset: 3284}` — the recovered immutable values decode to `"Curve.Fi USD Stablecoin"`, `"crvUSD"`, the cached EIP-712 domain separator, and chain id 1;
- **creation**: `{type: insert, reason: constructorArguments, offset: 3720}` — ABI-encoded name/symbol, internally consistent with the immutables.

Two supporting notes from the related research (argotorg/sourcify#2825):

- crvUSD is provably **immune** to the 0.3.7 call-graph ordering nondeterminism (its call graph admits exactly one valid topsort), so this PR alone is sufficient to fully verify it — no layout work involved.
- In a corpus of 105 observed 0.3.7 "bytecode don't match" failures, **99 are immutable-tail-only**, i.e. unblocked by this PR. Once it deploys, those contracts (and crvUSD-class ones already partially verified) will need a re-verification pass to pick up their runtime matches.

crvUSD would also make a strong integration-test fixture for this PR: a real mainnet contract exercising the immutable tail and constructor args simultaneously, with stable expected transformation values.

kuzdogan:
Thanks for this PR — and especially for surfacing something we'd completely missed: it turns out we've **never been storing `immutableReferences` for Vyper contracts at all**. cc @manuelwedler @marcocastignoli

A quick check on production confirms every Vyper row has `runtime_code_artifacts -> 'immutableReferences'` as `null`, even for `>= 0.3.10` contracts that genuinely have immutables (we computed them and used them for the runtime transformation, but dropped them at the DB layer). The old `constructor_args_immutables.json` fixture even asserted `null`, so the behavior was effectively locked in. This PR fixes it and adds proper coverage. 🙏

**Backfill:** existing rows stay `null` until re-verified, so we'll need to either re-run verifications ourselves or backfill the table with a script. For `>= 0.3.10` this should be straightforward — fetch all Vyper contracts, compile, and write the `immutableReferences`. (Legacy `< 0.3.10` needs the full compile path for the IR-derived size, but the same re-verification covers it.) This doesn't need to block this PR — let's track it in a separate issue.

**On the `PreRunCompilation` fallback:** since we'll be storing the references after this fix, the IR-derivation fallback there is effectively dead weight — in a pre-run the reconstructed `std_json_output` has no `ir`, so for legacy it just returns `{}`, and the stored-refs branch already covers everything once persisted. We should either **remove that useless part entirely** or **add a comment** explaining why it can't derive from IR in pre-run. Not a blocker either way, since storing the references makes it moot.

banteg:
Addressed the `PreRunCompilation` fallback note in `a688ae30`.

I kept the fallback in place because it can still recover modern Vyper `>=0.3.10` immutable references from CBOR for existing rows where stored refs are missing. The added comment clarifies that pre-run reconstructed Vyper output does not include `ir`, so legacy `<0.3.10` immutable references cannot be re-derived there and must come from the stored `immutableReferences` branch.

banteg:
Follow-up to my previous comment: after checking #2827, I tightened this further in `64147615`.

`PreRunCompilation` now prefers stored Vyper immutable references and calls the fallback without compiler output / compilation target. That means pre-run keeps the old `>=0.3.10` CBOR fallback for rows that have not been backfilled yet, but it no longer attempts legacy `<0.3.10` IR-derived inference there. Legacy pre-run references now come only from stored `immutableReferences`, which matches the backfill plan in #2827.

kuzdogan:
Ok great this is ready to go and should land early next week.

Please review the latest commits @marcocastignoli (since @manuelwedler is off on Monday)

marcocastignoli (changes requested):
As far as I understood, `PreRunCompilations` are always constructed by passing the output of `this.database.getSourcifyMatchByChainAddressWithProperties(... ["std_json_output", ...] ...)`. 

So the line `compilationTarget.evm.deployedBytecode.immutableReferences` in `PreRunCompilation.immutableReferences()` always returns whatever we store in `compiled_contracts.creation_code_artifacts->'linkReferences`.

So I don't think we need a fallback at all. In order to backfill all of the missing references we'll need to call `replaceContract` with a new REPLACE_METHOD `replace-vyper-immutables` in which we'll force a recompilation that will return the `immutableReference` correctly after the fixes in this PR.

After the script is executed we'll have `compiled_contracts.creation_code_artifacts->'linkReferences` values for each row, so `PreRunCompilation.immutableReferences()` will return the right values.

@kuzdogan tell me if this makes sense, if it does I can remove the fallback in `PreRunCompilation.immutableReferences()` and approve.

banteg:
Removed the PreRunCompilation Vyper fallback in f54531c4. The pre-run path now only reads stored deployedBytecode.immutableReferences and returns {} when they are missing. I kept the persistence of non-empty Vyper runtime immutableReferences for new/replaced verifications; the replace-vyper-immutables backfill remains a follow-up outside this PR. Checks: focused mocha grep for pre-run compiler outputs, and lib-sourcify tsc --noEmit.

marcocastignoli (approved):
lgtm

Issue #2751, opened 2026-04-15
Reduce json input on API to the fields we store


manuelwedler:
While working on Vyper's storage_layout_overrides, I came across this problem of the API: We currently take the standard json input from the user and pass it directly to the compiler as it is. This means a currently unsupported field like `interfaces` of Vyper (https://github.com/argotorg/sourcify/issues/2750) is given to the compiler and leading to the expected compilation output. However, we don't store this field in the database and this makes it impossible to recompile the contract later from the data we have in the database.

To address this, I propose to reduce the user given json input to the fields we actually support when passing it to the compiler. This enforces a failed verification for unknown fields which modify compiler behavior but we don't store in the database. The reasoning is we cannot control when a compiler decides to add a new top level json input field that we have to add support for.

kuzdogan:
How would you like to ensure consistency? A shared validation function across the API and the StorageService? 

The stripping down at the storage step happens implicitly, because we assign JSON input fields to database fields explicitly. So I'm wondering how we can make sure this step and the user input validation can be consistent. Maybe something like a JSON schema check?

manuelwedler:
At the moment, it is actually explicitly stripped down to `storage_layout_overrides` in lib-sourcify and server. Maybe we should simply pass all json input fields (apart from language, sources and settings) to the `additional_input` column and let the database constraint throw?

Still, some validation middleware on the API would be good for better UX. So, I think we will end up with defining the supported additional input fields in the db constraint and in the API layer.

Issue #2750, opened 2026-04-15
Add support for Vyper interfaces


manuelwedler:
While working on storage_layout_overrides, I found that Vyper has another top level field on standard json input that we don't store currently but changes the compiler behavior: `interfaces`. See https://docs.vyperlang.org/en/stable/compiling-a-contract.html#input-json-description

We should also store `interfaces` inside the `additional_input` column. If we don't store it, it is possible that we cannot recompile a verified contract that makes use of this feature.

PR #2823, opened 2026-06-09
Reduce standard JSON input to the fields we store

marcocastignoli:
See #2751

## Summary
- **lib-sourcify**: the `Verification` export now forwards *all* top-level standard JSON input fields other than `language`/`sources`/`settings` via `additionalInput` (previously it kept only `storage_layout_overrides`). Nothing used for compilation is silently dropped, which matters for the library's other consumers too.
- **server**: the `/v2/verify` API now rejects any unsupported top-level input field (anything outside `language`/`sources`/`settings` + `storage_layout_overrides`) with a clear `400`, instead of failing later with a generic `500` from the DB constraint.

## Why
Unsupported fields such as Vyper's `interfaces` (#2750) change the compiler output but aren't persisted, making the contract impossible to recompile from the database. The API allowlist mirrors the DB `validate_additional_input` CHECK constraint, which stays as the authoritative backstop — so no migration is needed.

### DIFF
--- a/packages/lib-sourcify/src/Verification/Verification.ts
+++ b/packages/lib-sourcify/src/Verification/Verification.ts
@@ -781,12 +781,8 @@ export class Verification {
         creationBytecodeCborAuxdata,
         immutableReferences: immutableReferences,
         metadata,
-        jsonInput: {
-          settings,
-          ...(Object.keys(additionalInput).length > 0
-            ? { additionalInput }
-            : {}),
-        },
+        jsonInput: { settings },
+        ...(Object.keys(additionalInput).length > 0 ? { additionalInput } : {}),
         compilationTime: this.compilation.compilationTime,
       },
     };
--- a/packages/lib-sourcify/src/Verification/Verification.ts
+++ b/packages/lib-sourcify/src/Verification/Verification.ts
@@ -33,7 +33,6 @@ import type {
 } from './VerificationTypes';
 import { SolidityBugType, VerificationError } from './VerificationTypes';
 import type {
-  VyperJsonInput,
   VyperOutputContract,
   ImmutableReferences,
   SolidityOutputContract,
@@ -725,6 +724,13 @@ export class Verification {
       // pass
     }
 
+    // Surface every top-level standard JSON input field used for compilation other than
+    // language/sources/settings (e.g. Vyper's `storage_layout_overrides`) so consumers can
+    // persist them.
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { language, sources, settings, ...additionalInput } =
+      this.compilation.jsonInput;
+
     return {
       address: this.address,
       chainId: this.chainId,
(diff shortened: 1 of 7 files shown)


packages/lib-sourcify/src/Verification/VerificationTypes.ts
```diff
@@ -134,7 +134,10 @@ export interface VerificationExport {
     metadata?: Metadata;
     jsonInput: {
       settings: SoliditySettings | VyperSettings | FeSettings;
-      storageLayoutOverrides?: VyperJsonInput['storage_layout_overrides'];
+      // Top-level standard JSON input fields used for compilation other than language/sources/settings.
+      additionalInput?: {
+        storage_layout_overrides?: VyperJsonInput['storage_layout_overrides'];
+      };
```
manuelwedler: I'd rather move `additionalInput` one level up as `additionalInput` is not a field of standard json input. Alternatively, you could just return the full json input and give the `SolidityJsonInput | VyperJsonInput` type to the `jsonInput`. But then the question is if we want to remove the sources and language fields from `compilation`.

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -725,6 +724,13 @@ export class Verification {
       // pass
     }
 
+    // Surface every top-level standard JSON input field used for compilation other than
+    // language/sources/settings (e.g. Vyper's `storage_layout_overrides`) so consumers can
+    // persist them.
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
manuelwedler, line 730: Let's get rid of this eslint-disable line and simply use the language, sources and settings variables in the return statement.

services/server/test/integration/apiv2/verification/verify.json.spec.ts
```diff
@@ -461,6 +461,37 @@ describe("POST /v2/verify/:chainId/:address", function () {
...
   });
 
+  it("should return a 400 if the standard json input contains an unsupported top-level field", async () => {
+    const jsonInput = JSON.parse(
+      JSON.stringify(chainFixture.defaultContractJsonInput),
+    );
+    // `interfaces` is a Vyper top-level field that Sourcify does not store. Such fields
+    // can change the compiler output and would make the contract impossible to recompile
+    // from stored data, so they must be rejected at the API.
+    jsonInput.interfaces = { "IFoo.json": { abi: [] } };
```
manuelwedler: we should use some other field name in this test. We will add support for `interfaces` so this test would need to get adapted later.

manuelwedler (changes requested):
Looks mostly good, just some minor changes that would improve it. 

Also I added a commit to reduce the redundancy of additional input field definitions. Let me know if you disagree with the change.

packages/lib-sourcify/src/Verification/Verification.ts
marcocastignoli: where would you use those values?

We are already using `this.compilation.language` and we cannot replace it since it's set as a class property. If you meant in `jsonInput`, I would rather not add additional fields already present in the db.

Same for sources

manuelwedler: ah nevermind. I confused `this.compilation.jsonInput.language` with `this.compilation.language`

manuelwedler (approved):
lgtm

kuzdogan:
Couple things when looking at this:
1. Should we implement #2750 too while we are on it? It can include a test that shows the verification indeed doesn't work without the `interface` field but will verify when included.
2. We are filtering top level JSON fields but can you inject arbitrary fields inside "settings", "sources" etc. and we still store them? I'm guessing not because I expect the compilers to throw with unsupported field but just wanted to make sure.

marcocastignoli:
> Should we implement https://github.com/argotorg/sourcify/issues/2750 too while we are on it?

Let's just open another PR for this

> We are filtering top level JSON fields but can you inject arbitrary fields inside "settings", "sources" etc. and we still store them?

Actually this is a real case documented in here: https://github.com/sourcifyeth/private-issues/issues/64#issuecomment-4545470942, and again I'd say let's open another PR if we want to fix this.

PR #2824, opened 2026-06-10
fix(server): don't send global ETHERSCAN_API_KEY to custom Etherscan explorers

kuzdogan:
## Problem

A chain can opt into Etherscan support via a custom explorer URL (`etherscanApi.url`) — used for Etherscan-compatible explorers that aren't on etherscan.io (e.g. BattleChain's `https://block-explorer-api.testnet.battlechain.com`).

In both the source-import path (`etherscan-util.ts`) and the contract-creation-tx path (`contract-creation-util.ts`), the API key was resolved as:

```
process.env[etherscanApi.apiKeyEnvName || ""] || process.env.ETHERSCAN_API_KEY || ""
```

When such a chain has no chain-specific `apiKeyEnvName`, this falls back to the global `ETHERSCAN_API_KEY` and appends it as `&apikey=<key>` to a request sent to the chain's own explorer host. That leaks our etherscan.io API key to a third-party server.

Separately, the custom explorer URL was never surfaced in the `/chains` API response, so the verification UI (verify.sourcify.dev) had no way to know a chain uses a non-canonical explorer and would always import from `api.etherscan.io`.

## Fix

**1. Don't leak the global key to custom explorers.** The global `ETHERSCAN_API_KEY` fallback now applies only to canonical etherscan.io chains (those without a custom `url`). Custom Etherscan-compatible explorers only receive a key explicitly configured for them via their own `apiKeyEnvName`; otherwise no key is sent (`&apikey=`).

Behavior for existing etherscan.io chains is unchanged.

**2. Expose the custom explorer URL in `/chains`.** The serialized chain object now includes an optional `etherscanApiUrl` field (sourced from `etherscanApi.url`), omitted for canonical Etherscan chains. This lets the UI import contracts from the correct explorer. OpenAPI spec updated accordingly.

> UI counterpart: sourcifyeth/verify.sourcify.dev#47 reads this `etherscanApiUrl` field and passes it through `EtherscanUtils.fetchFromEtherscan(..., customBaseUrl)`.

## Tests

- `contract-creation-util.spec.ts` (mocked `fetch`):
  - custom-`url` chain → outbound URL does not contain the global key (ends with `&apikey=`)
  - canonical chain (no `url`) → outbound URL does contain the global key
- `status-handlers.spec.ts`: asserts the `/chains` response exposes `etherscanAPI` as a boolean and `etherscanApiUrl`, when present, as a non-empty string.

### DIFF
--- a/services/server/src/server/services/utils/etherscan-util.ts
+++ b/services/server/src/server/services/utils/etherscan-util.ts
@@ -75,12 +75,18 @@ export const fetchFromEtherscanOrThrowError = async (
         : new BadRequestError(errorMessage);
     }
 
-    // Derive API key precedence: user -> chain-specific env -> global -> ''
-    const apiKey =
-      userApiKey ||
-      process.env[sourcifyChain.etherscanApi.apiKeyEnvName || ""] ||
-      process.env.ETHERSCAN_API_KEY ||
-      "";
+    // Derive API key precedence: user -> chain-specific env -> global -> ''.
+    // The global ETHERSCAN_API_KEY is only used for canonical etherscan.io
+    // chains. For custom Etherscan-compatible explorers (those with a `url`),
+    // we never fall back to it — that would leak our etherscan.io key to a
+    // third-party server. Such explorers only get a key explicitly configured
+    // for them via their own apiKeyEnvName.
+    const chainSpecificKey =
+      process.env[sourcifyChain.etherscanApi.apiKeyEnvName || ""];
+    const globalKey = sourcifyChain.etherscanApi.url
+      ? undefined
+      : process.env.ETHERSCAN_API_KEY;
+    const apiKey = userApiKey || chainSpecificKey || globalKey || "";
 
     return await EtherscanUtils.fetchFromEtherscan(
       sourcifyChain.chainId,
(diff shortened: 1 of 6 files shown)


services/server/src/server/services/utils/etherscan-util.ts
```diff
@@ -59,6 +59,23 @@ function mapLibError(err: any, throwV2Errors: boolean): never {
...
+  userApiKey?: string,
+): string => {
+  const chainSpecificKey =
+    process.env[sourcifyChain.etherscanApi?.apiKeyEnvName || ""];
+  const globalKey = sourcifyChain.etherscanApi?.url
+    ? undefined
+    : process.env.ETHERSCAN_API_KEY;
+  return userApiKey || chainSpecificKey || globalKey || "";
+};
+
```
marcocastignoli, line 78: @kuzdogan FYI since the etherscan key extraction was duplicated I added this method, that is reused in `contract-creation-util.ts`

marcocastignoli (approved):
lgtm

services/server/src/server/services/utils/etherscan-util.ts
kuzdogan: Nice thanks

Issue #2837, opened 2026-06-16
Update monitor to use Sourcify's server API v2


marcocastignoli:
CI is failing during brownouts: https://app.circleci.com/pipelines/github/argotorg/sourcify/11260/workflows/722d0ae0-fd3c-4209-b540-870daf4e794e/jobs/66800

marcocastignoli:
closed by https://github.com/argotorg/sourcify/pull/2840

PR #1561, opened 2024-08-14
Add sourcify request retry mechanism in Monitor

marcocastignoli:
Some of the /verify requests in the monitor running on GCP are failing with ECONNRESET error. We already tried debugging the problem but with no luck. Until we find a better solution we need to implement a retry mechanism in Monitor.

This PR implements a retry mechanism in Monitor configurable with: 

```
  sourcifyRequestOptions: {
    // Maximum number of retry attempts for contract verification requests after encountering an error
    maxRetries: 3,
    // Delay in milliseconds between each retry attempt for verification requests to Sourcify
    retryDelay: 30000,
  },
```

PR #2840, opened 2026-06-16
feat(monitor): submit contracts via Sourcify API v2

marcocastignoli:
See #2837

## Summary

Upgrades the Sourcify Monitor's contract submission from the v1 API to the v2 API: `PendingContract.sendToSourcifyServer` now POSTs to `/v2/verify/metadata/{chainId}/{address}` with the v2 metadata payload (`sources`, `metadata`, `creationTransactionHash`) instead of the v1 `{ chainId, address, files, creatorTxHash }` body.

## Notes

- v2 verification is asynchronous: the endpoint returns `202` with a `verificationId` (logged) and processes the job in the background, so the monitor submits and does not poll for the result.
- A `409 already_verified` response is treated as a benign terminal state — logged at info level and skipped, rather than surfaced as an error.
- Removes the retry mechanism (`sourcifyRequestOptions` / `maxRetries` / `retryDelay`) entirely: it was added in #1561 to work around ECONNRESET on the old synchronous v1 verify requests, and v2's fast `202` submission removes that long-lived-connection failure mode, so a failed submission is now simply logged and skipped.
- Network errors from `fetch` are no longer re-wrapped, preserving the original error's type/code/stack when logged, and the non-ok-status error message was trimmed to drop context the caller already logs.
- The existing monitor test now exercises the v2 endpoint; the retry test was removed along with the retry mechanism.

### DIFF
--- a/services/monitor/README.md
+++ b/services/monitor/README.md
@@ -66,12 +66,6 @@ The structure of the file is as such:
   },
   // Sourcify instances to verify the contracts on. Can be multiple
   sourcifyServerURLs: ["https://sourcify.dev/server/", "http://localhost:5555/"],
-  sourcifyRequestOptions: {
-    // Maximum number of retry attempts for contract verification requests after encountering an error
-    maxRetries: 3,
-    // Delay in milliseconds between each retry attempt for verification requests to Sourcify
-    retryDelay: 30000,
-  },
   defaultChainConfig: {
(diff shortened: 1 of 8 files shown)


kuzdogan (changes requested):
Let's have proper test coverage and fix codecov/patch. It's mostly the "already verified" path and error handling path that are uncovered.

marcocastignoli:
@kuzdogan I added tests for PendingContract's missing tested branches

kuzdogan (approved):

PR #2842, opened 2026-06-18
Update major dependency versions

manuelwedler:


manuelwedler:
I looked into the code, the socket alert seems like a false-positive to me

marcocastignoli (approved):
lgtm, I'm wondering are there any functionalities that are not tested by CI? such as debugging?

manuelwedler:
gcp-cli isn't because it is an circleCI orb. But I checked the changelog and the breaking change does not seem relevant to us.

Let's still keep this PR in mind, if we find anything not working.

PR #2844, opened 2026-06-22
Delete script schema support

marcocastignoli:
Currently the contract deletion script doesn't support schemas different than `public` (on VerA we use `vX`).


services/monitor/src/PendingContract.ts
```diff
@@ -141,46 +141,60 @@ export default class PendingContract {
       formattedSources[sourceUnitName] = source.content;
     }
 
+    const baseURL = sourcifyServerURL.replace(/\/+$/, "");
```
manuelwedler, line 144: Unrelated changes on the monitor

manuelwedler (changes requested):

services/monitor/src/PendingContract.ts
marcocastignoli: Sorry I opened the PR from an already existing branch

marcocastignoli: Here new PR: https://github.com/argotorg/sourcify/pull/2845

PR #2845, opened 2026-06-22
feat(scripts): add support for configurable PostgreSQL schema in delete-sourcify-match script

marcocastignoli:
Currently the contract deletion script doesn't support schemas different than `public` (on VerA we use `vX`).


manuelwedler (approved):
lgtm

PR #2855, opened 2026-06-29
feat(server): backfill missing Vyper immutableReferences

### LINKED ISSUE #2827: Backfill missing immutableReferences for Vyper contracts
### Context
While reviewing #2817 we discovered that we have **never been storing `immutableReferences` for Vyper contracts** — every Vyper row in `compiled_contracts` has `runtime_code_artifacts -> 'immutableReferences'` set to `null`, even for `>= 0.3.10` contracts that genuinely have immutables. The references were computed during verification and used for the runtime transformation, but dropped at the DB layer.

#2817 fixes this going forward (new verifications persist the references), but **existing rows stay `null` until re-verified.**

### Task
Backfill the missing `immutableReferences` for already-verified Vyper contracts, either by re-running verifications or with a one-off script.

For `>= 0.3.10` this should be straightforward:
1. Fetch all verified Vyper contracts.
2. Recompile each.
3. Write the resulting `immutableReferences` into `runtime_code_artifacts`.

Legacy (`< 0.3.10`) contracts need the full compile path to recover the IR-derived size (added in #2817), but the same re-verification pass covers them.

### Why
Without this, DB-backed pre-run / similarity verification can't reuse the references for existing contracts, and their stored artifacts remain incomplete.

Depends on #2817.

marcocastignoli:
## Summary

Vyper `immutableReferences` have never been persisted, every Vyper row in `compiled_contracts` has `runtime_code_artifacts -> 'immutableReferences'` set to `null`, even for contracts that genuinely have immutables. #2817 fixes this going forward; this PR backfills the existing rows. It adds a new `replace-vyper-immutable-references` custom replace method to the private `/replace-contract` endpoint, plus a `massive-replace-script` config that drives the backfill over all affected Vyper contracts.

## How it works

- `replaceVyperImmutableReferences` recompiles the contract (the endpoint is called with `forceCompilation: true`, required to recover the IR-derived size for legacy `< 0.3.10` contracts), reads `verification.compilation.immutableReferences`, and updates only the `immutableReferences` key of the existing `compiled_contracts.runtime_code_artifacts` via `jsonb_set`. The shared `compiled_contracts` row is updated, so every verified contract reusing that compilation is backfilled at once.
- Vyper-only: it throws for non-Vyper contracts (Solidity `immutableReferences` have always been stored), and is a no-op when a contract has no immutables (stays `null`, matching how new verifications store them).
- `config-backfill-vyper-immutable-references.js` selects Vyper contracts with null `immutableReferences`, rebuilds the std-json input from stored sources, and calls the endpoint with the new method. It skips `< 0.3.1` (Vyper had no immutables before then).

## Changes

- `customReplaceMethods.ts`: new method + registry entry.
- `private.stateless.paths.yaml`: OpenAPI doc for the new method.
- `private.stateless.spec.ts` + `test/sources/vyper/withImmutables/`: integration test and fixture (verify → null the field → backfill → assert restored, other artifacts untouched).
- `config-backfill-vyper-immutable-references.js`: the backfill config.
- `massive-replace-script.ts`: optional `PROCESS_ONLY_ONE` env flag (off by default) to process a single contract for verifying a config run.

Depends on #2817.
Closes https://github.com/argotorg/sourcify/issues/2827


services/server/src/server/apiv1/verification/private/stateless/customReplaceMethods.ts
```diff
@@ -170,7 +173,84 @@ export const replaceMetadata: CustomReplaceMethod = async (
...
+ * used together with `forceCompilation: true` so the references are recovered
+ * from a fresh compile (required for legacy `< 0.3.10` contracts whose immutable
+ * size is derived from the compiler IR). See issue #2827.
+ *
+ * Solidity is intentionally not supported: its `immutableReferences` have always
+ * been persisted, so this method throws for any non-Vyper contract.
+ *
+ * Only writes when there are references to write; contracts without immutables
+ * are left untouched (stays `null`), matching how new verifications store them.
+ */
```
kuzdogan, line 203: Just a small thing I noticed, also flagged by @manuelwedler in a different PR.

Recently Claude tends to overcomment on things and it really doesn't help, rather makes it very difficul to read things. Here example "Solidity is intentionally not supported", "Vyper `immutableReferences` were historically never persisted" are all context you provided and needed for this specific task but feels like an overkill to me.

Not a blocker etc. but something to keep in mind in general, and maybe we should include a proper prompt in our Claude Code settings to avoid this.

kuzdogan (approved):

PR #2857, opened 2026-06-30
feat(mcp): add an MCP server for looking up verified contracts

marcinciarka:
## Summary

Adds `services/mcp` (`sourcify-mcp`), a small Model Context Protocol - pulling verified-contract data straight from Sourcify. Six read-only tools, each a thin wrapper over the public v2 API:
`get_contract_abi`, `get_contract_metadata`, `get_source_files`, `check_verification_status`, `list_chain_contracts`, `list_supported_chains`.

## How it fits

Same pattern as `sourcify-monitor`: a standalone service that talks to the public HTTP API only. Tools take numeric chain IDs and distinguish "chain not supported" (400) from "contract not verified" (404 → `unverified`).

One non-obvious bit: the MCP SDK is ESM-only, so the package stays CJS and loads it via dynamic `import()` (isolated to two call sites, commented inline).

## Testing

Hermetic unit + protocol tests run in CI (new `test-mcp` job). There's also an opt-in live suite (`npm run test:live`). Smoke-tested end-to-end against production with real contracts on Ethereum, Base, and Arbitrum.

## Notes

I wasn't sure what to do about deployment, so I left it out — no Docker image, no publish wiring. It currently just runs locally over stdio (which is the normal way to use an MCP server), and the package is `private`. I'd really like people who know our infra to take it from here and figure out how/whether it should be hosted. Happy to work on it further.


marcinciarka:
Heads up for reviewers: this PR touches the CI config (`.circleci/continue_config.yml`) — I added a `test-mcp` job and an entry to the build/lint workspace paths. Worth a quick separate look at that diff on top of the code.

One thing I noticed: the `test-mcp` job actually ran and passed on CircleCI even though this is coming from my fork, which I wasn't expecting.p Since a fork PR can change the CI config and run its own code on your runners, you probably want to double-check that fork builds aren't getting any secrets/contexts — or that they require manual approval to run. Totally fine if that's already handled, just flagging since it surprised me.

kuzdogan:
Hi, thanks for the PR. Can you please explain a bit the actual use case here, especially vs. just calling the API? We thought about this a while ago but then didn't see much need or benefit to maintain an extra service. It's quite straightforward to use Sourcify services and the API is fairly well documented. 

Re the CI runs for the PRs, env vars and secrets are already held off already. Thanks for the heads up. Since we don't receive too many external PRs, it should be ok to keep automatic running on.

marcinciarka:
> Hi, thanks for the PR. Can you please explain a bit the actual use case here, especially vs. just calling the API? We thought about this a while ago but then didn't see much need or benefit to maintain an extra service. It's quite straightforward to use Sourcify services and the API is fairly well documented.
> 
> Re the CI runs for the PRs, env vars and secrets are already held off already. Thanks for the heads up. Since we don't receive too many external PRs, it should be ok to keep automatic running on.

It's really just convenience for AI agents. With the MCP server I can ask Claude Code or Cursor about a contract and it pulls the ABI/sources/status itself, correctly (instead of me pasting docs or it guessing the endpoints and chain-id format). One line of config and it works in any MCP client - that's basically the whole reason + since it's a thin read-only wrapper with no real logic, there's not much to maintain.

I'm happy for it to live in a separate repo if you'd rather not own it (or maybe create a npx package for this?)

EDIT: i also searched for the [`mcp`](https://github.com/search?q=repo:argotorg/sourcify%20mcp&type=code) but didnt find anything

marcocastignoli:
Thanks @marcinciarka, my two opposite solutions:
- implement this in a different repo, so then people can self host the mcp server (I wouldn't host it)
- use something like https://gofastmcp.com/integrations/openapi but in Typescript and integrate the mcp server directly in server (we already have OpenApi specifications)

@kuzdogan what's your favorite solution?

UPDATE: @manuelwedler pointed out that we already agreed to not serve an mcp server ourselves.

marcocastignoli:
Closing this because we want to go with a different approach, but this discussion is what made us change our mind. Thanks @marcinciarka.

The point is that we want to maintain it as little as possible, so we don't want to write and keep updated the tools by hand. Two options:

1) if we find a TypeScript library that generates an MCP server from an OpenAPI spec (like FastMCP does in Python), we generate it from our OpenAPI spec and put it directly in server
2) if not, we make a separate service generated from the same spec using FastMCP that we'll host

In both cases it's generated from the OpenAPI spec, this is why we prefer to not merge an implementation written by hand.

marcinciarka:
> Closing this because we want to go with a different approach, but this discussion is what made us change our mind. Thanks @marcinciarka.
> 
> The point is that we want to maintain it as little as possible, so we don't want to write and keep updated the tools by hand. Two options:
> 
> 1. if we find a TypeScript library that generates an MCP server from an OpenAPI spec (like FastMCP does in Python), we generate it from our OpenAPI spec and put it directly in server
> 2. if not, we make a separate service generated from the same spec using FastMCP that we'll host
> 
> In both cases it's generated from the OpenAPI spec, this is why we prefer to not merge an implementation written by hand.

Thats amazing news, thank you.

Issue #2792, opened 2026-05-20
Compiler subprocess invocations have no execution timeout


kuzdogan:
## Summary

None of the compiler subprocess invocations in `packages/compilers` set an
execution timeout. A compiler that hangs (infinite loop, deadlock, waiting
on a closed pipe, etc.) never terminates, so the wrapping Promise never
settles and the verification worker that started it is occupied indefinitely.

The verification worker pool (Piscina, `VerificationService.ts`) runs
`concurrentTasksPerWorker: 5` and only has an **idle** timeout — it does not
reclaim a worker stuck on a live-but-hung task. Enough hung compilations
therefore degrade and eventually stall verification capacity.

Noticed while reviewing #2788 (zksolc). The gap is pre-existing and applies
to every compiler, not just zksolc.

## Affected paths

| Compiler | Mechanism | Timeout | Output cap |
|---|---|---|---|
| solc (native) | `exec` via `asyncExec` (`common.ts`) | none | 250 MB `maxBuffer` |
| vyper | `exec` via `asyncExec` (`common.ts`) | none | 250 MB `maxBuffer` |
| fe | `spawnSync` (`feCompiler.ts`) | none | 250 MB `maxBuffer` |
| zksolc | async `spawn` (`zksolcCompiler.ts`) | none | none (unbounded) |

Notes:
- **fe is the worst case.** `spawnSync` is synchronous — a hung `fe build`
  blocks the worker thread's event loop entirely, freezing all 5 concurrent
  tasks on that thread rather than just the one.
- **zksolc additionally has no output cap.** `exec`/`spawnSync` enforce
  `maxBuffer` (250 MB -> `ENOBUFS`); zksolc's `spawn` accumulates stdout into
  an unbounded `Buffer[]`.

## Proposed fix

`spawn`, `exec`, and `spawnSync` all accept a built-in `timeout` (+
`killSignal`) option — Node kills the child when it fires. One option added
per call site:

- `asyncExec` (`common.ts`): add `timeout` + `killSignal` to the `exec` options.
- `feCompiler.ts`: add `timeout` + `killSignal` to the `spawnSync` options.
- `zksolcCompiler.ts`: add `timeout` + `killSignal` to the `spawn` options.

For zksolc, also add a `maxBuffer`-equivalent ceiling on the accumulated
stdout/stderr buffers.

### Choosing the timeout value

The default must account for legitimate long-running Solidity compilations.
[...]

Issue #2670, opened 2026-02-24
Investigate Large Contract Compilation Timeout


kuzdogan:
Investigate the issue from https://github.com/dhruvinparikh/skills/commit/413dd5e438e07af4dd86cc2d3c77a2472b05289e

### Large Contract Compilation Timeout

Contracts with many source files (e.g., 72 sources / ~320KB standard JSON) may cause the server to time out during compilation, resulting in:

```json
{
  "isJobCompleted": true,
  "error": {
    "customCode": "internal_error",
    "message": "Job timed out - container connection hung"
  }
}
```

Or the job may get permanently stuck with `"isJobCompleted": false`.

**Workaround**: Contact the chain's verification team to clear the stuck job or increase server compilation timeout. There is currently no client-side workaround when the metadata and similarity endpoints are not implemented.

kuzdogan:
Should be fixed with https://github.com/argotorg/sourcify/issues/2880

Issue #2669, opened 2026-02-24
Investigate not being able to repeat verification submissions


kuzdogan:
Investigate this issue from https://github.com/dhruvinparikh/skills/commit/413dd5e438e07af4dd86cc2d3c77a2472b05289e

### API Deduplication by Address

The API returns the **same `verificationId`** for repeated submissions to the same address. If a previous job is stuck/failed, new submissions get the old stuck ID — there is no way to force a fresh job from the client side.

kuzdogan:
Fixed with https://github.com/argotorg/sourcify/issues/2880

PR #2886, opened 2026-07-14
fix: prevent and auto-recover stuck v2 verification jobs (compiler OOM/hang) (#2880)

### LINKED ISSUE #2880: v2 verification jobs stuck in "running" for 10+ hours; stuck jobs block any resubmission for the address (`duplicate_verification_request`)
## Summary

Nine verification jobs on chain 43114 (Avalanche C-Chain) have been reporting `isJobCompleted: false` for 7+ hours. While a job is in this state, every new verification attempt for the same address is rejected with HTTP 429 `duplicate_verification_request`, and there is no API to cancel a job — so the nine addresses are locked out of verification entirely.

A tenth job from the same submission batch failed quickly and cleanly with `compiler_error` ("Command failed: solc-linux-amd64-v0.8.29 --standard-json\nKilled") — i.e. the compiler process was OOM-killed. We suspect the workers for the other nine died the same way, but without the job being marked failed.

## Likely root cause

The duplicate check in `services/server/src/server/services/VerificationService.ts` uses the in-memory map `activeVerificationsByChainIdAddress`, and the entry is only deleted in the `finally` of the verification promise. If a verification dies without settling that promise (e.g. the compiler process is OOM-killed), the entry — and with it the address lock — stays forever: there is no timeout, TTL, or reaper. The corresponding job row also never reaches a completed state, so `GET /v2/verify/{id}` reports `isJobCompleted: false` indefinitely.

(A side effect: restarting the server releases the locks since the map is in-memory, but the stale job rows remain, and nothing prevents the same wedge from happening again.)

## Timeline

- 2026-07-11 ~02:01 UTC: submitted 10 contracts via `POST /server/v2/verify/{chainId}/{address}` with `stdJsonInput` (full hardhat solc input, ~2.3 MB each — in hindsight too large; smaller metadata-route submissions for other addresses worked fine).
- 2026-07-11 ~02:06 UTC: one job (ConfigTimelockController, `e077981e-b531-4d69-a2b9-d76bcb108029`) completed as `compiler_error: ... Killed`.
- 2026-07-11 09:19 UTC (7.3 h later): the other nine still return `isJobCompleted: false` with `jobStartTime` 2026-07-11T02:01:41Z.
- Resubmission attempts for those addresses via `POST /v2/verify/metadata/{chainId}/{address}` return 429 `duplicate_verification_request` ("Contract ... is already being verified").

## Stuck jobs (chain 43114)

| verificationId | address | contract |
|---|---|---|
| b94ae26e-8729-42aa-be0e-c2883ac47f9f | 0x0B33EBA531e5a5A331a3Ff9F418B8205F01C2869 | AdlHandler |
| 43203591-6d39-4eee-9fda-f7d7c6c28b66 | 0x7BA7Ae61887F1aca28E0FE5aB1434ce85b6606aa | ClaimHandler |
| c22683a4-a1a6-4c94-932e-087a100a9205 | 0xC8470eBBb6960FD00Db7D4C9f71CAda27e8f6Bfa | ConfigSyncer |
| b8bfe805-42e7-445e-aa87-cb75144515a7 | 0xD3269C530fa732aec52a93dbC5A0863A71a4B56B | ExchangeRouter |
| ad6697b5-90ed-4ea8-8333-b62015d32a26 | 0xCFf2Afe6E2C3D4d21aDf0e5A5210B94A09546852 | GelatoRelayRouter |
| 93e84290-9e35-45db-b836-05f0865c76d2 | 0x2Dd99f39f58445CDDC57AA5E0DB2C367335BBD44 | GlvShiftHandler |
| 073f1d91-ab64-4420-9c45-4c6ce119b21d | 0x597A7898B0D31BefEE52488b3F046eAc870F94CF | JitOrderHandler |
| da390a6b-9f45-4468-87e8-8a24d3990391 | 0xFcc5520A1d42291F46c09BA7B5bcDa6044E1702F | LiquidationHandler |
| b3545cae-5c35-4805-abe4-cec1f4bb2df7 | 0xAb3EDf0f3eed6804BAe1bD9bF90109ccadFD262e | SimulationRouter |

Failed-but-correctly-marked sibling for comparison: `e077981e-b531-4d69-a2b9-d76bcb108029` (0x7E0c21527Bfd1FEFA727Aa79e4b722E73713ac04, ConfigTimelockController) — `compiler_error`, errorId `4aebdf07-0f4c-4984-829f-ccfa5fc25f9a`. That address was later verified successfully via the metadata endpoint once its job slot was free, so the lock behavior is the only thing keeping the nine above unverifiable.

## Asks

1. Clear or expire the nine stuck jobs so the addresses can be resubmitted (a server restart appears sufficient for the locks, given the in-memory map).
2. Add a timeout / reaper for entries in `activeVerificationsByChainIdAddress` (and mark the corresponding job failed) so a crashed verification cannot hold an address lock forever.
3. Consider handling compiler OOM on large standard-json inputs by failing the job (like the ConfigTimelockController case) instead of leaving it running.

## Environment

- Endpoint: https://sourcify.dev/server
- Chain: 43114
- Submissions were plain HTTP (no SDK); happy to provide full request payloads or reproduce.

kuzdogan:
## Background

Fixes #2880. Nine v2 verification jobs on Avalanche C-Chain (43114) got stuck at `isJobCompleted: false` for 7+ hours, each blocking resubmission for its address with `429 duplicate_verification_request`.

**Root cause:** a verification worker's native `solc` subprocess was OOM-killed mid-compile. The immediate trigger is memory (see below), but the reason the job *wedged* is a bug in `asyncExec` (`packages/compilers`): it had **no error handler on the child's `stdin`**, so a compiler killed mid-write of the standard-JSON input emitted `EPIPE` with no listener → an uncaught exception in the Piscina worker thread → the compile promise never rejected. So `completed_at` was never set. The v2 duplicate check is DB-based (`completed_at IS NULL` ⇒ "in progress"), so the address stays locked — and because the lock is a DB row, a server restart doesn't clear it.

Investigation against production found this is **not Avalanche-specific**: ~1,260 jobs are currently stuck with `completed_at IS NULL`, the oldest ~2 months old, across many chains and both `/v2/verify` and `/v2/verify/similarity`.

### Related previous issues

- **#2242** — Automate stale job recovery: stale jobs accumulate and need to be marked failed; proposes a periodic reaper.
- **#2792** — Compiler subprocess invocations have no execution timeout: a hung compiler occupies a worker forever (the "hang" sibling of this OOM case).
- **#2669** — Investigate not being able to repeat verification submissions: repeated submissions return the same stuck `verificationId`, with no way to force a fresh job.
- **#2670** — Investigate Large Contract Compilation Timeout: large standard-JSON inputs time out or get permanently stuck at `isJobCompleted: false`.
- **#2359** (closed issue; implemented in **#2613**) — Temporarily store standard json input for failed verifications: added S3 storage of the submitted input, but only on *settled* failures.

### The OOM trigger

We set solc `outputSelection` to `{"*":{"*":[…]}}` — i.e. **output every contract in every source file**, with heavy artifacts (`evm.legacyAssembly`, `evm.bytecode.generatedSources`, source maps, storage layout, metadata). For a 119-source contract that's ~40 MB of output and ~800 MB of solc Resident Set Size (RSS) per compile; several running concurrently exhaust the container and it gets OOM-killed.

Reproduced with the native linux-amd64 solc 0.8.29 on a real contract from the issue:

| outputSelection | output | solc peak RSS |
|---|---|---|
| current `"*":"*"` | 39.9 MB | 794 MB |
| scoped to target contract | 2.0 MB | 187 MB |

I recall the `"*":"*"` (all contracts) was originally needed for `extra-file-input` handling, but that shouldn't be necessary for newer compilers — so we can likely turn it off, or only enable the broad selection for that specific case. For the common path we should be able to safely output only the target contract.

## Fixes (multi-step)

- [x] **Scope `outputSelection` to the compilation target** (lib-sourcify) — prevents the OOM at the source: ~20× smaller output, ~4× less solc RSS. Also shrinks storage/network payloads for every verification.
- [x] **Compiler subprocess robustness + execution timeout** (#2792, compilers) — the deeper root cause. `asyncExec` now settles exactly once, rejects on `stdin` error / write failure (so a compiler killed mid-write no longer hangs), and enforces a wall-clock timeout (`SOLC_COMPILE_TIMEOUT_MS`, default **45 min**, `SIGKILL`). Failures are attributed with dedicated `compiler_timeout` / `compiler_out_of_memory` codes. Covers Solidity and Vyper.
- [x] **Stale-job reaper** (#2242) — a **pg_cron** scheduled job (every 15 min) marks `completed_at IS NULL` jobs older than **3 h** as failed with a new `job_abandoned` code, releasing the address lock. Follows the existing `refresh-signature-stats` pg_cron pattern, with the same graceful fallback when the extension isn't available (pg_cron is already enabled in production). Backed by a partial index `verification_jobs (started_at) WHERE completed_at IS NULL` — required because the table is ~77 M rows / ~50 GB with no timestamp index, turning the reaper query from a ~4.5-min full sequential scan into a sub-ms lookup (index built `CONCURRENTLY`).

### Considered but deferred

- **Compiler subprocess memory cap** — bounding the solc child's address space via `ulimit -v` (opt-in `SOLC_MAX_ADDRESS_SPACE_KB`, default-off) so solc dies as a clean `compiler_out_of_memory` *before* the container OOM-killer targets the Node server. Deferred: `ulimit -v` caps *virtual* (not resident) memory, needs ops tuning against real metrics, and risks false-failing legitimate large compiles. Straightforward to add later as an opt-in knob if the outputSelection reduction + timeout prove insufficient under load.

## Migration / deploy notes

- Two migrations: (1) `CREATE INDEX CONCURRENTLY` for the partial index (non-transactional), and (2) the `cron.schedule` for the reaper. The reaper relies on **pg_cron** being enabled (as it already is for `refresh-signature-stats`); on a stack without it the schedule is skipped gracefully and jobs can be reaped by running the same UPDATE manually.
- Regenerate the checked-in `services/database/sourcify-database.sql` dump after applying.
- [x] **One-time cleanup of the existing stuck jobs** — a separate operator-run SQL step (sets `completed_at` + `error_code = 'internal_error'` — recognized by the currently-deployed server — on jobs older than the threshold), releasing those locks immediately.

  **Re-measured against production on 2026-08-11:** the original ~1,260 backlog *was* cleared manually, but the jobs have come back — **472 are stuck again**, 469 of them older than the 3 h threshold and 454 older than a day, oldest `2026-08-03 06:44 UTC`:

  ```
   stuck_total | older_than_3h | older_than_1d |     oldest_started_at      |     newest_started_at
  -------------+---------------+---------------+----------------------------+---------------------------
           472 |           469 |           454 | 2026-08-03 06:44:27.438+00 | 2026-08-11 10:22:05.898+00
  ```

  That is ~59 new stuck jobs per day since the manual cleanup, each one an address that gets `429 duplicate_verification_request` until someone intervenes — the reason the automated reaper is needed rather than another manual pass.

### Open question: forcing a fresh job (#2669)

Should we let a client force a fresh job instead of returning the same stuck `verificationId`? IMO we shouldn't need this in the first place — an OOM-killed or otherwise dead process should never leave a job stuck (the reaper + timeouts above should guarantee that). Linking #2669 for discussion rather than committing to it.

Related: #2670 (large-contract compilation timeout).


services/server/src/server/apiv2/errors.ts
```diff
@@ -286,6 +287,12 @@ export function getVerificationErrorMessage(
       return "The server encountered an unexpected error.";
     case "no_similar_match_found":
       return "No similar verified contracts were found in the database.";
+    case "compiler_timeout":
+      return "The compiler timed out while compiling the contract. The compilation took too long and was aborted.";
+    case "compiler_out_of_memory":
+      return "The compiler process was killed unexpectedly, likely because it ran out of memory while compiling the contract.";
```
manuelwedler: These cases are duplicated across SourcifyLibError and here. Remove the occurrence here.

packages/compilers/src/lib/common.ts
```diff
@@ -1,6 +1,10 @@
 import { exec } from 'child_process';
 import { logDebug, logError, logSilly } from '../logger';
 import type { OutputError } from '@ethereum-sourcify/compilers-types';
+import {
+  COMPILER_TIMEOUT_CODE,
+  COMPILER_OOM_CODE,
+} from '@ethereum-sourcify/compilers-types';
```
manuelwedler, line 7: `@ethereum-sourcify/compilers-types` is only a **devDependency** of this package (and of `lib-sourcify`, which gets the same value imports in `AbstractCompilation.ts`). That was fine while every import was `import type` (erased at build), but these are value imports, so the built JS now contains a real `require('@ethereum-sourcify/compilers-types')`.

Both packages are published to npm, and npm doesn't install devDependencies of a dependency — external consumers will crash at load with `Cannot find module '@ethereum-sourcify/compilers-types'`. It only works in this repo because of workspace hoisting.

Please move `@ethereum-sourcify/compilers-types` to `dependencies` in both `packages/compilers/package.json` and `packages/lib-sourcify/package.json`.

packages/compilers/src/lib/common.ts
```diff
@@ -56,6 +60,22 @@ export async function fetchWithBackoff(
   throw new Error(`Failed fetching ${resource}`);
 }
 
+// Default wall-clock timeout for a single compiler invocation: 45 minutes.
+// Overridable via SOLC_COMPILE_TIMEOUT_MS. A genuinely hung compiler must be
+// killed so the verification job can fail instead of hanging forever (#2880).
+const DEFAULT_COMPILE_TIMEOUT_MS = 2_700_000;
```
manuelwedler: This timeout only covers `asyncExec`, i.e. the native-binary path. Two compile paths still have no wall-clock timeout:

- the soljson fallback (worker thread in `solidityCompiler.ts`), used for versions below the native-binary thresholds
- the Fe compiler (`spawnSync`, which also blocks synchronously)

A hung compile there still wedges a Piscina worker until the 3h reaper marks the job abandoned — so #2792 is only partially closed by this PR. Probably acceptable given soljson is only used for old versions, but worth stating explicitly whether that's intentional (or noting it in #2792).

packages/compilers/src/lib/common.ts
```diff
@@ -56,6 +60,22 @@ export async function fetchWithBackoff(
...
+
+function getCompileTimeoutMs(): number {
+  const fromEnv = process.env.SOLC_COMPILE_TIMEOUT_MS;
```
manuelwedler: This introduces the first functional env var in the published packages. Since #2058 removed `process.env` from lib-sourcify, all functional configuration in the packages flows explicitly through the API (`solcRepoPath`, `vyperRepoPath`, ...) — an ambient env read buried here in `asyncExec` breaks that convention, and library consumers won't discover it. (Browser compatibility isn't the concern here — `compilers` is inherently Node-only — just consistency of the config design.)

Consider making it an explicit `timeoutMs` parameter on `useSolidityCompiler`/`useVyperCompiler`, threaded down to `asyncExec`, and reading the env var in the server's config layer instead. The server already wraps these functions (`SolcLocal`/`VyperLocal`), so the parameter doesn't need to touch lib-sourcify. That would also solve two smaller issues with the current shape:

- the name: `SOLC_COMPILE_TIMEOUT_MS` also governs Vyper compiles since `asyncExec` is shared — the server config would own a compiler-agnostic name (e.g. `COMPILER_TIMEOUT_MS`) and document it next to the other settings (right now the variable isn't documented anywhere)
- the tests: `common.spec.ts` could pass `timeoutMs: 150` directly instead of mutating and restoring `process.env`

packages/compilers/src/lib/common.ts
```diff
@@ -64,36 +84,100 @@ export function asyncExec(
...
-      throw new Error('No stdin on child process');
+      settleReject(new Error('No stdin on child process'));
+      return;
     }
+    // If the compiler dies mid-write (e.g. OOM-killed while we stream a large
+    // input), the stdin pipe emits 'error' (EPIPE). Without a listener this is
+    // an uncaught exception in the Piscina worker thread and the compile promise
+    // never rejects -> the verification job hangs forever (#2880). Treat it as
+    // an unexpected process death (out of memory).
+    child.stdin.on('error', (err: NodeJS.ErrnoException) => {
```
manuelwedler, line 172: Attributing every stdin `'error'` to OOM can mislabel unrelated failures: any child that exits without reading its stdin — e.g. a missing/corrupt compiler binary (`exit 127`) — triggers `EPIPE` on a large input before the exec callback fires, so it surfaces as `compiler_out_of_memory` instead of the real spawn/exit error. That could mask e.g. a bad-binary incident as OOM.

An alternative that keeps the crash fix but improves attribution: don't settle here — just record that an stdin error happened (the listener alone prevents the uncaught exception). The exec callback always fires eventually and has better signals (`killed`, `signal`, exit code) to attribute the death; only reject from the recorded stdin error in the (practically impossible) case where the callback would otherwise resolve successfully.

services/database/migrations/20260724100000_schedule_stale_verification_job_reaper.sql
```diff
@@ -0,0 +1,57 @@
...
+        UPDATE public.verification_jobs
+        SET completed_at = NOW(),
+            error_code = 'job_abandoned',
+            error_id = gen_random_uuid(),
+            verified_contract_id = NULL,
+            compilation_time = NULL
+        WHERE completed_at IS NULL
+          AND started_at < NOW() - INTERVAL '3 hours';
+        $job$
+    );
```
manuelwedler, line 54: Consider extracting the UPDATE into a database function and scheduling a one-line call to it:

```sql
CREATE FUNCTION public.reap_stale_verification_jobs(stale_threshold interval DEFAULT '3 hours')
RETURNS integer
LANGUAGE sql
AS $$
  WITH reaped AS (
    UPDATE public.verification_jobs
    SET completed_at = NOW(),
        error_code = 'job_abandoned',
        error_id = gen_random_uuid(),
        verified_contract_id = NULL,
        compilation_time = NULL
    WHERE completed_at IS NULL
      AND started_at < NOW() - stale_threshold
    RETURNING 1
  )
  SELECT count(*)::integer FROM reaped;
$$;

-- in the DO block:
PERFORM cron.schedule('reap-stale-verification-jobs', '*/15 * * * *',
  'SELECT public.reap_stale_verification_jobs();');
```

Benefits:

- **Makes the reaper testable.** The test DB (`postgres:15-alpine`) has no pg_cron, so the schedule is skipped there and the reaper logic is currently untestable without duplicating the SQL. A plain `CREATE FUNCTION` lands in the test DB through the normal migration path, so a server integration test can exercise the production reaper SQL verbatim.
- **The manual fallback becomes a single call** returning the reaped count, instead of copy-pasting the UPDATE out of a migration file.
- **Matches the existing pattern**: `refresh-signature-stats` schedules a one-liner; the logic stays out of the cron entry. Changing the threshold later is a `CREATE OR REPLACE FUNCTION` instead of unschedule + reschedule.

**When addressing this, please also add the integration test:** create a pending job (`createMockJob`-style in `jobs.spec.ts`), run `SELECT reap_stale_verification_jobs('0 seconds')`, then assert that `GET /v2/verify/:id` returns `isJobCompleted: true` with `error.customCode: "job_abandoned"`, and that resubmitting the same chain/address no longer returns `429 duplicate_verification_request`. This also locks in the subtlety that `error_id` must be set for the API to build the error object at all (`SourcifyDatabaseService` requires both `error_code` and `error_id`).

(The down migration would gain a `DROP FUNCTION IF EXISTS`, and the function needs to be in the regenerated `sourcify-database.sql` dump.)

manuelwedler (changes requested):

services/server/src/server/apiv2/errors.ts
kuzdogan: Removed in `e9a8a5e8`. `compiler_timeout` and `compiler_out_of_memory` are lib-sourcify `CompilationErrorCode`s, so they already resolve through the `default` branch into `getErrorMessageFromCode`. `job_abandoned` stays here — the reaper sets it and lib-sourcify has no counterpart.

packages/compilers/src/lib/common.ts
kuzdogan: Moved to `dependencies` in both packages in `2e828b4c`. One thing to flag: the range is still `^1.3.1`, and the published 1.3.1 predates these constants. `lerna version` rewrites internal ranges at release time so the published packages will point at the new version, but worth a look when we cut the release.

packages/compilers/src/lib/common.ts
kuzdogan: Implemented rather than deferred, in `7e74eb53`. The soljson worker now has a timer that calls `worker.terminate()`, plus an `exit` guard so a worker that dies without posting a result rejects instead of leaving the promise pending. Fe uses `spawnSync`'s own `timeout`/`killSignal` and maps `ETIMEDOUT` to the same `COMPILER_TIMEOUT` code. All three paths report through one `createCompilerTimeoutError`, and both new paths have tests. So this PR now closes #2792 (compiler subprocess invocations have no execution timeout) fully rather than partially.

packages/compilers/src/lib/common.ts
kuzdogan: Done in `78fa8e7f` (plus `47147356` renaming the parameter to `timeoutMs`). `asyncExec` takes the timeout as an argument; `useSolidityCompiler`/`useVyperCompiler`/`useFeCompiler` thread it down; the server's `cli.ts` reads `COMPILER_TIMEOUT_MS` and passes it to `SolcLocal`/`VyperLocal`/`FeLocal` and through Piscina `workerData` to the verification workers. Documented in the compilers README and `.env.dev`, and the tests now pass `150` directly instead of mutating `process.env`.

packages/compilers/src/lib/common.ts
kuzdogan: Done in `db9659ee`, as suggested. The stdin listener stays (it's what prevents the uncaught `EPIPE`) but now only records the error; the exec callback settles the promise using `killed`/`signal`/exit code, and the recorded stdin error is used only if the callback would otherwise resolve successfully. Added a test asserting a failed spawn is not reported as out of memory.

services/database/migrations/20260724100000_schedule_stale_verification_job_reaper.sql
kuzdogan: Done in `766e909c`. `public.reap_stale_verification_jobs(stale_threshold interval DEFAULT '3 hours')` returns the reaped count, the cron entry is a one-liner, the down migration drops the function, and the dump is regenerated. Added the integration test you described — it creates a pending job, asserts the resubmit gets `429 duplicate_verification_request`, runs `SELECT reap_stale_verification_jobs('0 seconds')`, then asserts `isJobCompleted: true` with `customCode: "job_abandoned"` and that a new submission is accepted. Verified locally against `postgres:15-alpine` with no pg_cron: the function installs through the normal migration path and the test passes.

manuelwedler (approved):
looks good to me. Thanks!

PR #2888, opened 2026-07-15
Speed up delete-sourcify-match orphan checks and add verification_jobs FK index

marcocastignoli:
## Summary

Deleting a single contract with `delete-sourcify-match` could take ~50 minutes. Two fixes bring it down to seconds:

- **New index on `verification_jobs.verified_contract_id`**: every row deleted from `verified_contracts` triggered a full seq scan of `verification_jobs` (74M rows) for the FK integrity check. Measured: 180ms → 0.085ms per check on a 2M-row table.
- **`COUNT(*)` → `LIMIT 1` existence probes** in the orphan checks (signatures, sources, code): counting all references of a common signature like `transfer(address,uint256)` (~3M rows) took ~80s; an existence probe takes ~40ms.

### DIFF
--- a/scripts/delete-sourcify-match.ts
+++ b/scripts/delete-sourcify-match.ts
@@ -298,13 +298,15 @@ async function deleteCompiledContractSources(
   for (const row of sourcesResult) {
     const sourceHash = row.source_hash;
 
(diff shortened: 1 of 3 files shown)


manuelwedler (approved):
looks good to me

Issue #2887, opened 2026-07-15
Verification Failed: internal_error - Ethereum Mainnet (1) 0x4e59...956C


Argeric:
## Verification Job Failed

**Job ID:** `5629fcee-5e0d-43c2-a994-5c6c776f5da1`
**Chain:** Ethereum Mainnet (1)
**Contract Address:** `0x4e59b44847b379578588920cA78FbF26c0B4956C`
**Error Code:** `internal_error`
**Server URL:** `https://sourcify.dev/server`

### Error Details
```
The server encountered an unexpected error.
```

### Additional Information
- **Job Started:** 2026-07-15T03:40:27.000Z
- **Job Finished:** 2026-07-15T03:40:27.000Z

### User Input and Expectation
**Why do you expect this verification to succeed?**

**Language:**

**Compiler Version:**

**EVM Version:**

**Optimizer Settings:**

**Contract Identifier:**

**Files Used:**

```
<!-- Upload files directly to this issue -->
```

**Additional Context:**

---
*This issue was automatically generated from the Sourcify verification UI.*

This is a Yul contract; its repository is https://github.com/Arachnid/deterministic-deployment-proxy/tree/master#. Below is the JSON I submitted:
{
  module: 'contract',
  action: 'verifysourcecode',
  codeformat: 'solidity-standard-json-input',
  contractaddress: "0x4e59b44847b379578588920cA78FbF26c0B4956C",
  contractname: "deterministic-deployment-proxy.yul:Proxy",
  compilerversion: "v0.5.8+commit.23d335f2",
[...]

PR #2890, opened 2026-07-16
fix(lib-sourcify): throw specific error when compiler does not output deployedBytecode

marcocastignoli:
Verifying a Yul contract compiled with solc <0.6.9 (e.g. the deterministic deployment proxy, [#2887](https://github.com/argotorg/sourcify/issues/2887)) failed with `internal_error`: those compiler versions don't output `evm.deployedBytecode` for Yul, and the `runtimeBytecode` getter crashed with a TypeError that escaped as an unexpected error.

Compilations whose output has no `evm.deployedBytecode` now fail with a new `runtime_bytecode_not_found_in_compiler_output` error. The check lives in `AbstractCompilation.compileAndReturnCompilationTarget`, so every language's compile() enforces it, and the rest of the verification flow can keep relying on the field being present. Contracts with a present-but-empty runtime bytecode (abstract contracts) still fail with `compiled_bytecode_is_zero` as before.

### DIFF
--- a/packages/lib-sourcify/src/Compilation/AbstractCompilation.ts
+++ b/packages/lib-sourcify/src/Compilation/AbstractCompilation.ts
@@ -108,6 +108,19 @@ export abstract class AbstractCompilation {
     // We call contractCompilerOutput() before logging because it can throw an error
     const compilationTargetContract = this.contractCompilerOutput;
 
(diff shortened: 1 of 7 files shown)


manuelwedler (approved):
Looks good :)

PR #2901, opened 2026-07-27
fix(lib-sourcify): don't store sources unused by the compilation target

### LINKED ISSUE #2363: Reconsider which sources to store for a verified contract
## Problem Description

While investigating https://github.com/argotorg/sourcify/issues/2227#issuecomment-3269584348, I was thinking about ways how unexpected sources could end up in the database. I found that we store any sources included in the standard json input that a user provides. However, these sources can be totally unrelated to the deployed contract. If anyone wants to review the sources of a contract, such unrelated sources can hinder the review significantly. Unrelated sources could be added with a malicious intention to make reviews harder.

**IMO, we should only store the sources mentioned in the metadata output.**

## Example

I used this standard json to verify a contract:

```json
{
  "language": "Solidity",
  "sources": {
    "contracts/1_Source.sol": {
      "content": "// SPDX-License-Identifier: GPL-3.0\n\npragma solidity >=0.8.2 <0.9.0;\n\ncontract Storage {\n\n    uint256 sourceNumber;\n\n    function store(uint256 num) public {\n        sourceNumber = num;\n    }\n}"
    },
    "contracts/2_Source.sol": {
      "content": "// SPDX-License-Identifier: GPL-3.0\n\npragma solidity >=0.8.2 <0.9.0;\n\ncontract Storage {\n\n    uint256 maliciousNumber;\n\n    function store(uint256 num) public {\n        maliciousNumber = num;\n    }\n}"
    },
    "contracts/3_Source.sol": {
      "content": "// SPDX-License-Identifier: GPL-3.0\n\npragma solidity >=0.8.2 <0.9.0;\n\ncontract Storage {\n\n    uint256 maliciousNumber;\n\n    function store(uint256 num) public {\n        maliciousNumber = num;\n    }\n}"
    },
    "contracts/4_Source.sol": {
      "content": "// SPDX-License-Identifier: GPL-3.0\n\npragma solidity >=0.8.2 <0.9.0;\n\ncontract Storage {\n\n    uint256 maliciousNumber;\n\n    function store(uint256 num) public {\n        maliciousNumber = num;\n    }\n}"
    }
  },
  "settings": {
    "optimizer": { "enabled": false, "runs": 200 },
    "outputSelection": {
      "*": {
        "": ["ast"],
        "*": [
          "abi",
          "metadata",
          "devdoc",
          "userdoc",
          "storageLayout",
          "evm.legacyAssembly",
          "evm.bytecode",
          "evm.deployedBytecode",
          "evm.methodIdentifiers",
          "evm.gasEstimates",
          "evm.assembly"
        ]
      }
    },
    "remappings": []
  }
}
```

The contract which I deployed is `contracts/1_Source.sol:Storage`. All other three sources are not imported, and thus not necessary for recompilation and verification.

The contract in the repo:
https://repo.sourcify.dev/11155111/0x1Ae02670Ce0365c3cA27a54102520D7EE144D7b9

All sources of the standard json input show up. I could add hundreds of more unrelated sources to the compiled contract and they would all be added to the database and shown in the repo.

You can also see that metadata has only the right source:

```json
  "sources": {
    "contracts/1_Source.sol": {
      "keccak256": "0x950a4e1e6128413ac552d7cc8a723c09cc82fa74b076e88bd908c40481bcb102",
      "license": "GPL-3.0",
      "urls": [
        "bzz-raw://98b00b96f988904d040056b8ab1c1188770702c76e75d7f5d15f90ade6cdfb2f",
        "dweb:/ipfs/QmSAVCQswCx9HRZ7hsYhTtADbRhibYLfsi8XYusuRfQqGs"
      ]
    }
  }
```

## Etherscan case

Etherscan has the same problem. However, it's even worse for them because they don't show the compilation target and no metadata in their web interface.

I verified the same contract via the same json input on Etherscan:
https://sepolia.etherscan.io/address/0x1Ae02670Ce0365c3cA27a54102520D7EE144D7b9#code

As you can see, there are 4 sources of which each has a `Storage` contract, but it is unclear which contract is the deployed one.

marcocastignoli:
## Summary

Fixes #2363

Until now, all sources of a standard JSON input were stored with the verified contract, even sources completely unrelated to the deployed contract. Unrelated sources can hinder contract reviews and could be added maliciously on purpose.

After a successful match, `Verification` now removes the input sources that are not listed in the compilation target's metadata, so only the sources needed for the verification get stored. This covers all entry points (v2 std-json, v1 `/verify/solc-json`, Etherscan import) since they all go through `Verification.verify()`. No storage-layer changes are needed: the storage services already persist `VerificationExport.compilation.sources`.

## The extra file input bug case

As noted in the issue, unused sources cannot always be removed: with the optimizer enabled they can affect the bytecode (extra file input bug, e.g. argotorg/solidity#14494, argotorg/solidity#15686). In that case the sources are needed to reproduce the onchain bytecode and must all be kept.

Since the bug is undetectable when the user submits the full original input (the match simply succeeds), **a check compilation without the unused sources decides whether they are removable**. To keep the cost minimal, it only runs when all of these hold, and is paid once per contract at first verification:

- the verification succeeded (failed jobs never pay for it)
- the input contains unused sources **this gate is the most important one, this will avoid running the recompilation on most cases**
- the optimizer is enabled (the bug cannot occur otherwise)
- solc is below 0.8.22, or the input sets a custom Yul optimizer step sequence

### Why the version and custom-steps gate

Above 0.8.22 the bug only reproduces with a custom Yul optimizer step sequence. That isn't a coincidence: solc names the functions in its intermediate output using a counter that runs across every file in the input, so adding unrelated files renames them. A cleanup pass later shortens those names and makes different choices depending on what it started with. Normally none of this survives, because the optimizer rewrites the code thoroughly enough to erase it — a near-empty sequence like `"u"` skips exactly those passes, so the renaming makes it through to the bytecode.

This was checked against argotorg/solidity#15686 on solc 0.8.36, which still reproduces there but only with `settings.optimizer.details.yulDetails.optimizerSteps` set ([details](https://github.com/argotorg/solidity/issues/15686#issuecomment-5103812547)).

Measured against production traffic, unused sources appear in ~4.2% of recent matches and the optimizer is on for ~53%, so the old gate fired on ~2.7% of verifications. With this gate, modern std-json submissions using default optimizer steps no longer pay for the second compilation at all.

The check compilation itself is cheaper than the main one: it compiles only the reduced source set and outputs only `evm.bytecode.object` and `evm.deployedBytecode.object`.

### DIFF
--- a/packages/lib-sourcify/src/Verification/Verification.ts
+++ b/packages/lib-sourcify/src/Verification/Verification.ts
@@ -867,8 +867,7 @@ export class Verification {
             bytecode: {
               sourceMap: (
                 contractCompilerOutput as
-                  | SolidityOutputContract
-                  | VyperOutputContract
+                  SolidityOutputContract | VyperOutputContract
               )?.evm?.bytecode?.sourceMap,
               linkReferences: (contractCompilerOutput as SolidityOutputContract)
                 ?.evm?.bytecode?.linkReferences,
--- a/packages/lib-sourcify/src/Verification/Verification.ts
+++ b/packages/lib-sourcify/src/Verification/Verification.ts
@@ -1,4 +1,5 @@
 import type { AbstractCompilation } from '../Compilation/AbstractCompilation';
+import { findContractInCompilerOutput } from '../Compilation/AbstractCompilation';
 import { logDebug, logInfo, logWarn } from '../logger';
 import type { SourcifyChain } from '../SourcifyChain/SourcifyChain';
 import { lt } from 'semver';
@@ -425,10 +426,10 @@ export class Verification {
           checkJsonInput,
           forceEmscripten,
         );
-        const checkContract =
-          checkOutput.contracts[compilation.compilationTarget.path][
-            compilation.compilationTarget.name
-          ];
+        const checkContract = findContractInCompilerOutput(
+          checkOutput,
+          compilation.compilationTarget,
+        ) as SolidityOutputContract;
         if (
           `0x${checkContract.evm.bytecode.object}` !==
             compilation.creationBytecode ||
--- a/packages/lib-sourcify/src/Verification/Verification.ts
+++ b/packages/lib-sourcify/src/Verification/Verification.ts
@@ -358,9 +358,9 @@ export class Verification {
   /**
    * Removes input sources that are not listed in the target's metadata.
    * Only runs after a successful match, so failed verifications never pay
-   * for it. With the optimizer enabled, unused sources can change the
-   * bytecode (extra file input bug), so in that case they are only removed
-   * after checking that the bytecode stays the same without them.
+   * for it. When the compiler settings make the extra file input bug possible,
+   * unused sources can change the bytecode, so in that case they are only
+   * removed after checking that the bytecode stays the same without them.
    * See https://github.com/argotorg/sourcify/issues/2363
    */
   private async removeUnusedSources(forceEmscripten: boolean): Promise<void> {
@@ -386,10 +386,20 @@ export class Verification {
       return;
     }
     // A check compilation is only needed when the unused sources could have
-    // affected the bytecode (extra file input bug), which can only happen
-    // with the optimizer enabled. With the optimizer disabled, the unused
-    // sources are removed without recompiling.
-    if (compilation.jsonInput.settings.optimizer?.enabled) {
+    // affected the bytecode (extra file input bug), which requires the
+    // optimizer to be enabled. From 0.8.22 on we additionally require a custom
+    // Yul optimizer step sequence: unrelated sources shift the names solc
+    // generates in the Yul IR, and only a hand-picked sequence skips the passes
+    // that would otherwise normalize that difference away.
+    // See https://github.com/argotorg/solidity/issues/15686#issuecomment-5103812547
+    const settings = compilation.jsonInput.settings;
+    const usesCustomYulOptimizerSteps =
+      settings.optimizer?.details?.yulDetails?.optimizerSteps !== undefined;
+    const mayHitExtraFileInputBug =
+      settings.optimizer?.enabled &&
+      (semver.lt(compilation.compilerVersion, '0.8.22') ||
+        usesCustomYulOptimizerSteps);
+    if (mayHitExtraFileInputBug) {
       try {
         // Compile without the unused sources, only outputting the bytecodes
         const checkJsonInput: SolidityJsonInput = {
--- a/packages/lib-sourcify/src/Verification/Verification.ts
+++ b/packages/lib-sourcify/src/Verification/Verification.ts
@@ -38,9 +38,11 @@ import type {
   SolidityOutputContract,
   FeOutputContract,
   SoliditySettings,
+  SolidityJsonInput,
   Metadata,
 } from '@ethereum-sourcify/compilers-types';
 import { SolidityMetadataContract } from '../Validation/SolidityMetadataContract';
+import { SolidityCompilation } from '../Compilation/SolidityCompilation';
 import type { VyperCompilation } from '../Compilation/VyperCompilation';
 
 function auxdataLacksMetadataOrIntegrityHash(
@@ -295,6 +297,10 @@ export class Verification {
     }
 
     if (this.creationMatch !== null || this.runtimeMatch !== null) {
+      // Remove unused sources so they don't get stored with the verified contract
+      // See https://github.com/argotorg/sourcify/issues/2363
+      await this.removeUnusedSources(forceEmscripten);
+
       logInfo('Verified contract', {
         address: this.address,
         chainId: this.sourcifyChain.chainId,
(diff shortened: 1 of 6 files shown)


kuzdogan:
Approach looks right to me. My main hesitation was how often the check-compilation would actually run, so I pulled some numbers — it's less of a problem than I feared, but I think we can narrow it further.

### How often does it trigger?

The recompile needs Solidity + optimizer enabled + at least one input source missing from `metadata.sources`. I measured against production using stored data as a proxy: input sources = rows in `compiled_contracts_sources`, used sources = keys in `sourcify_matches.metadata->'sources'`.

On the 20,000 most recent matches (~9.5h of live traffic): optimizer is on for 53%, but only **4.2% have unused sources at all** — so the recompile fires on **2.7%**.

A random sample across all history gives 11.3%, but that's dominated by older contracts. The rate has fallen steadily over the past year, from roughly 80% in mid-2025 to 1–2% by early 2026 — presumably as the metadata path took over, since it builds its input *from* `metadata.sources` and structurally can't have extras. Only std-json submissions can trigger this at all.

So the cost is acceptable as-is. But given the trigger population is small and shrinking, I'd rather we also skip the recompile where we're confident the bug can't occur.

### Proposal: also gate on compiler version < 0.8.22

The upstream history:

| issue | affected | status |
|---|---|---|
| [#12281](https://github.com/ethereum/solidity/issues/12281) | 0.6.12, 0.7.0 | closed |
| [#14250](https://github.com/ethereum/solidity/issues/14250) | 0.8.19 | fixed in 0.8.21 |
| [#14494](https://github.com/ethereum/solidity/issues/14494) | 0.8.21 | fixed in 0.8.22 |
| [#15686](https://github.com/ethereum/solidity/issues/15686) | 0.8.26, 0.8.28 | open |

I'd suggest treating 0.8.22 as the cutoff and skipping the check-compile above it. That collapses the trigger population to legacy contracts and takes the recompile off the path for essentially all current traffic.

To be explicit about what we'd be accepting: #15686 is still open and does reproduce above that cutoff. I filed it, so I don't want to pretend otherwise. My read is that it's a narrow tail — one report in over a year, with a fix branch (`yul_node_handles`) already demonstrated to solve it — and that paying a second compile on every triggering verification to cover it isn't a good trade. If we're wrong, the failure mode is a stored source set that doesn't reproduce the on-chain bytecode, which is recoverable by re-verifying once detected.

Note we did version-gate this originally — `=0.6.12 || =0.7.0`, removed in Nov 2023 (`f1be6297`, "instead of checking the version, check that the auxdatas match"). So this would be partially reverting to that shape, with a cutoff informed by what's been fixed since.

### Better ways to filter?

Is there a sharper signal we could use here, beyond compiler version, to decide when the check-compile is actually needed?

One consequence worth stating either way: contracts that do hit the bug keep all their sources, so #2363 ends up partially rather than fully closed.

kuzdogan:
Follow-up on the version gate I proposed above: I reproduced [#15686](https://github.com/argotorg/solidity/issues/15686) on solc 0.8.36 to see how much risk that cutoff actually accepts ([details upstream](https://github.com/argotorg/solidity/issues/15686#issuecomment-5103812547)).

It does still reproduce above 0.8.22 — but only with a **custom Yul optimizer step sequence**. That contract sets `settings.optimizer.details.yulDetails.optimizerSteps: "u"`.

That isn't a coincidence, which is what makes it usable as a gate. Solc names the functions in its intermediate output using a counter that runs across every file in the input, so adding unrelated files renames them. A cleanup pass later shortens those names, and makes different choices depending on what it started with. Normally none of this survives, because the optimizer rewrites the code thoroughly enough to erase it — a near-empty sequence like `"u"` skips exactly those passes, so the renaming makes it through to the bytecode. With solc's default optimizer sequence the bug does not happen at all.

So instead of a plain version cutoff, I'd suggest gating on:

```ts
const usesCustomYulSteps =
  jsonInput.settings.optimizer?.details?.yulDetails?.optimizerSteps !== undefined;

const mayHitExtraFileInputBug = lt(compilerVersion, "0.8.22") || usesCustomYulSteps;
```

and running the double compilation only when that holds (on top of the existing "optimizer enabled" + "has unused sources" gates). That keeps the recompile off essentially all modern traffic while still covering the one case we know reproduces today.

marcocastignoli:
my gate already is based on optimizer enabled ([it has always been about the optimizer](https://github.com/argotorg/solidity/issues/12281#issuecomment-970375808)).

> It does still reproduce above 0.8.22 — but only with a custom Yul optimizer step sequence. That contract sets `settings.optimizer.details.yulDetails.optimizerSteps: "u"`.

this happens only when the optimizer settings are specifically set like `settings.optimizer.details.yulDetails.optimizerSteps: "u"`, or it happens generically even with just `compilation.jsonInput.settings.optimizer?.enabled == true`?

kuzdogan:
It happens when the customer optimizer settings are set in this case only the step `u`. Doesn't happen when optimizer turned on but default optimizer is used

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -349,6 +355,118 @@ export class Verification {
...
+   * for it. When the compiler settings make the extra file input bug possible,
+   * unused sources can change the bytecode, so in that case they are only
+   * removed after checking that the bytecode stays the same without them.
+   * See https://github.com/argotorg/sourcify/issues/2363
+   */
+  private async removeUnusedSources(forceEmscripten: boolean): Promise<void> {
+    const compilation = this.compilation;
+    if (
+      !(compilation instanceof SolidityCompilation) ||
+      compilation.language !== 'Solidity'
```
manuelwedler, line 371: Is this actually also relevant for Yul?

packages/lib-sourcify/src/Verification/Verification.ts
```diff
@@ -349,6 +355,118 @@ export class Verification {
...
+        };
+        const checkOutput = await compilation.compiler.compile(
+          compilation.compilerVersion,
+          checkJsonInput,
+          forceEmscripten,
+        );
+        const checkContract =
+          checkOutput.contracts[compilation.compilationTarget.path][
+            compilation.compilationTarget.name
+          ];
```
manuelwedler: Minor: this direct lookup misses the pre-0.4.9 solcjs quirk that `AbstractCompilation.contractCompilerOutput` handles, where contracts are keyed under `''` instead of the source path. For 0.4.7–0.4.8 contracts with the optimizer enabled, the check compilation will always fail here with a TypeError, so their unused sources are never removed (and every such verification logs the "Error checking unused sources" warning). The failure direction is safe, but if you want the removal to work for those versions too, the `checkOutput.contracts['']?.[name]` fallback from `contractCompilerOutput` could be reused here.

packages/lib-sourcify/src/Verification/Verification.ts
kuzdogan: No — Yul has no metadata. Checked with solc 0.8.26 on the `cas-forwarder` fixture: with `outputSelection: ["*"]` the contract only has `evm`, `ir`, `irOptimized`, no `metadata` field (a Solidity control input does get one). So we'd return at the `!metadataSources` guard below anyway — the `language` check is redundant rather than load-bearing.

manuelwedler: perfect

packages/lib-sourcify/src/Verification/Verification.ts
kuzdogan: Good catch — fixed in 9d91c10f. Since this lookup was being done in several places (and only the `contractCompilerOutput` getter handled the `''` key), I pulled it out into a `findContractInCompilerOutput` function and used it everywhere. That also covers the same latent case in the edited-contract compile for cbor auxdata positions.

manuelwedler (approved):

manuelwedler:
Looks good, thanks :)

Issue #2898, opened 2026-07-21
Pin workflow images by hash


kuzdogan:
We should pin the workflow images (orbs) we use via the hashes intead of versions. If somehow the version image get's overwritten, it can cause an security issue 

E.g.
https://github.com/argotorg/sourcify/blob/63132bb0f24c5d9c8f3418e9d2d9f4dd28319cb0/.circleci/config.yml#L17

Additionally, do this for all sourcify repos + clear signing. There's a repo setting that makes this mandatory for Github Actions https://x.com/pcaversaccio/status/2072285478397825401

PR #2915, opened 2026-08-04
security: pin CI images and actions by digest

kuzdogan:
## Summary

See #2898

Tags are mutable — whoever controls a namespace can repoint `cimg/node:22.22.0` or `postgres:15-alpine` at different content, and a compromised tag lands straight in CI with our secrets. This pins every third-party image to its **multi-arch manifest digest** (so arm64/amd64 both keep working) and the one GitHub Action to a full-length commit SHA.

| What | Change |
|---|---|
| `.circleci/*.yml` | `cimg/base:2024.12`, `cimg/node:22.22.0`, `postgres:15-alpine` executors pinned (35 refs) |
| `services/*/Dockerfile*` | `node:22.22.0-bookworm`, `node:22.22.0-trixie-slim` pinned |
| `services/*/docker-compose*.yml` | `postgres:15-alpine`, `postgres:15-bookworm`, `node:22.22.0-bookworm-slim` pinned |
| `.github/workflows/_fire-routine.yml` | `actions/github-script@v9.0.0` → `@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0` |

## Deliberately not changed

- **CircleCI orbs.** The line the issue points at (`circleci/path-filtering@3.0.0`) is already safe: production orbs published under a full semver version are immutable — they can't be edited or overwritten, only superseded by a new version — and there is no digest syntax for orbs. All four orbs already use exact semver rather than a floating `@3`, `@volatile`, or mutable `@dev:` tag. Worth keeping an eye on in review going forward.
- **`ghcr.io/argotorg/sourcify/server:latest`** in `services/server/docker-compose.yml` — that's the image *we* publish for users to run; pinning it would defeat its purpose.
- **`services/database/database-specs/`** — Verifier Alliance submodule, not ours.

## Renovate

Pinning by digest without an updater is a security *downgrade* — you freeze onto a base image and stop receiving its patches. So this also sets `pinDigests: true`, which keeps digests refreshed going forward.

That required one fix: `ignoreDeps: ["node", "cimg/node"]` would have blocked digest updates too, freezing those images permanently. It's replaced with a rule that disables major/minor/patch bumps for them (preserving the intent — the Node version stays manually controlled) while letting digest refreshes through.

## Follow-ups (not in this PR)

- Enable **Settings → Actions → General → "Require actions to be pinned to a full-length commit SHA"**, ideally at the `argotorg` org level so it covers all Sourcify repos. Do this *after* this merges, or the bot workflows break. Local `uses: ./…` reusable-workflow references are exempt, so the three bot workflows are fine.
- Commit signing ("clear signing" in the issue) is a separate concern — branch protection plus signed Renovate commits. Probably deserves its own issue.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -71,7 +71,7 @@ parameters:
 aliases:
   - &build-base
     docker:
-      - image: cimg/base:2024.12
+      - image: cimg/base:2024.12@sha256:03c91d4730297b5af3718e1e81d43e210eae02be271ed7d370ccb3b994dad2bd
     resource_class: xlarge
     parameters:
       run-build:
@@ -119,7 +119,7 @@ aliases:
           command: cd ~/project/metacoin-source-verify && ./scripts/monitor_e2e.sh
           no_output_timeout: 30m
     docker:
-      - image: cimg/node:22.22.0
+      - image: cimg/node:22.22.0@sha256:2c1320ee6a256628f698938038700911276310d40cc9887cfc8f66ec0daa4343
     resource_class: small
   - &verification-e2e-base
     steps:
@@ -136,7 +136,7 @@ aliases:
           name: verification test
           command: cd ~/project/metacoin-source-verify && ./scripts/verification_e2e.sh
     docker:
-      - image: cimg/node:22.22.0
+      - image: cimg/node:22.22.0@sha256:2c1320ee6a256628f698938038700911276310d40cc9887cfc8f66ec0daa4343
     resource_class: small
 ####################
 #### WORKFLOWS #####
@@ -281,7 +281,7 @@ jobs:
       DOCKERFILE: "services/4byte/Dockerfile"
   publish-multiarch-images:
     docker:
-      - image: cimg/base:2024.12
+      - image: cimg/base:2024.12@sha256:03c91d4730297b5af3718e1e81d43e210eae02be271ed7d370ccb3b994dad2bd
     resource_class: small
     steps:
       - checkout
@@ -310,7 +310,7 @@ jobs:
             ./.circleci/scripts/deploy_to_gcp.sh
   build-and-lint:
     docker:
-      - image: cimg/node:22.22.0
+      - image: cimg/node:22.22.0@sha256:2c1320ee6a256628f698938038700911276310d40cc9887cfc8f66ec0daa4343
     resource_class: medium
     working_directory: ~/sourcify
     steps:
@@ -347,7 +347,7 @@ jobs:
             - .git/modules
   test-bytecode-utils:
     docker:
-      - image: cimg/node:22.22.0
+      - image: cimg/node:22.22.0@sha256:2c1320ee6a256628f698938038700911276310d40cc9887cfc8f66ec0daa4343
     resource_class: small
     working_directory: ~/sourcify
     steps:
@@ -363,7 +363,7 @@ jobs:
       - codecov/upload
   test-compilers:
     docker:
-      - image: cimg/node:22.22.0
+      - image: cimg/node:22.22.0@sha256:2c1320ee6a256628f698938038700911276310d40cc9887cfc8f66ec0daa4343
     resource_class: medium
     working_directory: ~/sourcify
     steps:
@@ -379,7 +379,7 @@ jobs:
       - codecov/upload
   test-lib-sourcify:
     docker:
-      - image: cimg/node:22.22.0
+      - image: cimg/node:22.22.0@sha256:2c1320ee6a256628f698938038700911276310d40cc9887cfc8f66ec0daa4343
     resource_class: medium
     working_directory: ~/sourcify
     steps:
@@ -395,7 +395,7 @@ jobs:
       - codecov/upload
   test-monitor:
     docker:
-      - image: cimg/node:22.22.0
+      - image: cimg/node:22.22.0@sha256:2c1320ee6a256628f698938038700911276310d40cc9887cfc8f66ec0daa4343
     resource_class: medium
     working_directory: ~/sourcify
     steps:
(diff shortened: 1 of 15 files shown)


manuelwedler:
Note that this does not close #2898 since we need to do it for all repos. I updated the initial comment.

services/database/docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     build:
       context: .
       dockerfile_inline: |
-        FROM postgres:15-bookworm
+        FROM postgres:15-bookworm@sha256:b0c5bab0fbba8e0c221f73b1dc6359ec35f8650074377e727299df248fc8ad51
```
manuelwedler: This digest will never be refreshed by Renovate: the docker-compose manager only extracts `image:` values and explicitly skips services that have a `build:` section (`.filter((service) => isString(service?.image) && !service?.build)` in [`lib/modules/manager/docker-compose/extract.ts`](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/docker-compose/extract.ts)), so `FROM` lines inside `dockerfile_inline` are invisible to it. That freezes `postgres:15-bookworm` on today's build permanently — exactly the "pin without an updater is a security downgrade" failure mode the PR description warns about.

Two ways to fix it:
1. Move the inline Dockerfile out to a real file (e.g. `services/database/Dockerfile`) and reference it via `build.dockerfile` — the `dockerfile` manager picks it up like the other Dockerfiles in this PR.
2. Add a `customManagers` regex entry matching `FROM` lines inside compose files, e.g.:
   ```json
   {
     "customType": "regex",
     "managerFilePatterns": ["/(^|/)docker-compose[^/]*\\.ya?ml$/"],
     "matchStrings": ["FROM (?<depName>[^\\s:@]+):(?<currentValue>[^\\s@]+)@(?<currentDigest>sha256:[a-f0-9]+)"],
     "datasourceTemplate": "docker"
   }
   ```

Option 1 is simpler and less fragile.

---

.circleci/continue_config.yml
```diff
@@ -515,7 +515,7 @@ jobs:
             wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
             echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" | sudo tee /etc/apt/sources.list.d/pgdg.list
             sudo apt-get update
-            # Install PostgreSQL 15 client tools to match the postgres:15-alpine server image
+            # Install PostgreSQL 15 client tools to match the postgres:15-alpine@sha256:3d0f7584ed7d04e27fa050d6683a74746608faf21f202be78460d679cc56461f server image
```
manuelwedler: Find-and-replace side effect: this is a prose comment, not an image reference, so the digest is pure noise here — and it will silently go stale, since Renovate only rewrites the actual `- image:` lines, not comments. Suggest reverting to the plain tag:

```suggestion
            # Install PostgreSQL 15 client tools to match the postgres:15-alpine server image
```

---

renovate.json
```diff
@@ -16,9 +16,20 @@
...
       "minimumReleaseAge": "14 days"
+    },
+    {
+      "description": "We choose the Node version by hand, so turn off version bumps for these two images. 'digest' is deliberately not in the list above: rebuilds of the tag we already use still come through, otherwise pinning by digest would freeze us on an unpatched base image forever.",
+      "matchPackageNames": ["node", "cimg/node"],
+      "matchUpdateTypes": ["major", "minor", "patch"],
+      "enabled": false
+    },
+    {
+      "description": "A digest update is upstream rebuilding a tag we already trust, usually to ship OS security patches. The 14-day wait above guards against compromised package releases, which does not apply here and would only delay the patches we pin for.",
```
manuelwedler: Two problems with this description:

1. **The threat-model claim is overstated.** A compromised upstream is not distinguishable from a benign rebuild: an adversary who repoints the tag produces exactly a digest update, and Renovate will propose the malicious digest like any other. A quarantine period would in principle apply to compromised image rebuilds the same way it applies to compromised npm releases. The pin's real protection is that the compromise arrives as a reviewable PR instead of landing silently in the next CI run — which also means digest-update PRs should never be automerged.

2. **The rule is currently a no-op either way.** Renovate doesn't compute a `releaseTimestamp` for docker digests, and releases without a timestamp pass the age check immediately — so the 14-day wait in the rule above never actually gated digest updates, and this `"0 days"` changes no behavior (maintainer-confirmed in [renovate#38212](https://github.com/renovatebot/renovate/discussions/38212), feature gap tracked in [renovate#38656](https://github.com/renovatebot/renovate/issues/38656)). Keeping the rule as explicit documentation of intent is fine, but the description shouldn't imply the wait would otherwise apply.

Suggested rewording that keeps the rule but states the facts:

```suggestion
      "description": "Digest updates go out immediately: they ship the OS patches we pin for, and Renovate currently cannot age-gate docker digests anyway (no releaseTimestamp for digests, see renovatebot/renovate#38656). Caveat: a compromised upstream rebuild also arrives as a digest update — the pin's protection is that it lands as a reviewable PR instead of silently in CI, so digest updates must never be automerged.",
```

---

kuzdogan:
Thanks, all three were right — fixed in `e3e027be`.

**`services/database/docker-compose.yml`** — extracted to a real `services/database/Dockerfile` and referenced via `build.dockerfile` (your option 1). I checked whether the `build:` exclusion bit anything else: `docker-compose.local.yml`'s `server` service also has a `build:`, but it carries no pinned image, so this was the only file affected. `docker compose config` still resolves.

**`.circleci/continue_config.yml`** — reverted, that was a straight bug in my find-and-replace. The substitution matched the bare tag anywhere in the file rather than only on `image:` lines. Worth noting my "everything is pinned" check couldn't have caught it, since it only inspected `image:`/`FROM` lines — exactly the blind spot. I've since checked the inverse (every `sha256:` occurrence sits on an image line) and it's clean.

**`renovate.json`** — you were right that my description was wrong on the facts. Confirmed your read: Renovate doesn't compute a `releaseTimestamp` for docker digests, so `minimumReleaseAge` never gated them and my `"0 days"` rule was a no-op. The description asserted a protection that was never active.

I decided to turn off digest updates entirely (`b366b61b`). This means we'll keep the same digest images indefinitely. Against that we could set up a routine scan over the image digests as well, but I don't think that's needed — we can update the SHAs when needed, and it should be fine to keep using the same version. Regular digest updates can easily become cumbersome. If the image works for us, it works.

manuelwedler (approved):

Issue #2917, opened 2026-08-04
Planner statistics broken: n_distinct off by 50-650x on key FK columns; stats-collector reset starving autovacuum


manuelwedler:
# Planner statistics are broken: n_distinct off by 50–650× on key FK columns; stats-collector reset (2026-07-05) is starving autovacuum

Two related problems found while investigating the similarity-API timeouts (#2891). They affect query planning across the whole database, not just similarity search.

## Problem 1: `n_distinct` is wrong by orders of magnitude on the skewed FK columns

Postgres estimates rows-per-lookup as `table_rows / n_distinct`. `pg_stats` currently claims:

| Column | Claimed `n_distinct` | Proven lower bound (1% sample) | Best estimate of truth | Error |
|---|---|---|---|---|
| `verified_contracts.compilation_id` | 10,028 | 85,142 | ~5.4M (= compilations, FK-complete) | ~500× |
| `contract_deployments.contract_id` | 36,899 | 250,717 | ~24M (≈ all contracts) | ~650× |
| `compiled_contracts_signatures.signature_hash_32` | 16,406 | 357,187 | ~9.1M (all signatures referenced) | ~550× |
| `compiled_contracts_sources.source_hash` | 26,368 | 135,260 | ~5.4M (all sources referenced) | ~200× |
| `compiled_contracts_signatures.compilation_id` | 68,609 | 78,306 | ~5M | ~75× |
| `compiled_contracts_sources.compilation_id` | 96,821 | 64,735 (block-clustered sample) | ~5.9M | ~55× |
| `contracts.runtime_code_hash` | 72,858 | 171,540 | ≥ ~0.7 × rows | ~100×+ |

Proof method: a `TABLESAMPLE` distinct count is a hard lower bound on the true distinct count — five of the seven claims are directly disproven by their own 1% sample. The remaining two follow from FK structure (every `compiled_contracts` / `sources` row is referenced).

Practical meaning: the planner believes one compilation has ~347 source rows (true: ~6.5) and ~2,170 signature rows (true: ~29), one signature appears in ~9,000 compilations (true: ~16), one contract has ~1,000 deployments (true: ~1.6).

[...]

PR #2918, opened 2026-08-04
Fix similarity API database load and timeouts

### LINKED ISSUE #2891: Similarity API: reduce DB load from redundant candidate detail queries
## Problem

Running the similarity verification API in bulk puts very high load on the production DB. Two distinct problems:

### A. The prefix query itself times out (`statement_timeout = 290s`)

Observed in production (chain 56, `0x22a949ee77Cc2940a670592669967a7d87f03b91`):

```
canceling statement due to statement timeout (code 57014)
  at Database.getVerifiedContractsByRuntimeCodePrefix
  at SourcifyDatabaseService.getSimilarityCandidatesByRuntimeCode
```

`getVerifiedContractsByRuntimeCodePrefix` ([Database.ts](services/server/src/server/services/utils/Database.ts#L266)) matches on `substring(code.code FROM 1 FOR 75)`. The expression index `idx_code_code_first_75` makes the lookup itself cheap, but **75 bytes is not selective** — common solc boilerplate / proxy prefixes are shared by a huge number of `code` rows. That set then fans out through `compiled_contracts → verified_contracts → sourcify_matches → contract_deployments`, and `LIMIT 20` (applied on deployment rows, at the end) cannot short-circuit it. A single pathological bytecode burns a full **290s** of a DB connection + CPU before being cancelled — on its own enough to degrade the DB under a bulk script.

### B. Redundant heavy detail queries

When the prefix query does return, `getSimilarityCandidatesByRuntimeCode` ([SourcifyDatabaseService.ts](services/server/src/server/services/storageServices/SourcifyDatabaseService.ts#L1188)) fires up to **20 heavy detail queries in parallel** (`Promise.all` over `getSourcifyMatchByChainAddressWithProperties`, [Database.ts](services/server/src/server/services/utils/Database.ts#L196)), each pulling `std_json_input`/`std_json_output` — a `compiled_contracts_sources` + `sources` JOIN with `GROUP BY` aggregating all source files plus full recompiled bytecode (potentially MBs per row).

Inefficiencies:
1. **Only the first match is used.** The worker loop returns on the first candidate that verifies (`_verifySimilarity`, [verificationWorker.ts](services/server/src/server/services/workers/verificationWorker.ts#L333)), so ~19/20 heavy payloads are fetched and discarded in the common case.
2. **Duplicate compilations.** The prefix query `LIMIT`s on deployment rows with no dedup, so one compiled contract deployed to N addresses produces N identical detail fetches and N identical recompilations.
3. **Unnecessary indirection.** The candidate payload used by `createPreRunCompilationFromStoredCandidate` ([database-util.ts](services/server/src/server/services/utils/database-util.ts#L937)) is entirely compilation-level (no chain/address). The (chain,address) → re-query detour isnt needed.

### C. The timeout error is not handled / surfaced meaningfully

When the prefix query is cancelled, the `catch` in `verifyFromSimilarityViaWorker` ([VerificationService.ts](services/server/src/server/services/VerificationService.ts#L338)) maps everything to a generic `internal_error` job status. The raw pg `57014` (`canceling statement due to statement timeout`) is neither detected nor translated. We should map this to a more user-friendly, specific error (e.g. a dedicated custom code indicating the similarity search timed out / too many candidates) so API consumers get an actionable result instead of an opaque internal error.

## Proposed fixes

**Bound the prefix query (fixes A):**
- Dedup + `LIMIT` early at the compilation level (`SELECT DISTINCT compiled_contracts.id ... LIMIT 20`) **before** fanning out to `verified_contracts`/`sourcify_matches`/`contract_deployments`, so the limit can actually short-circuit.
- Increase prefix selectivity (longer prefix, or match full `code` length) and/or skip similarity for tiny/known-proxy bytecodes where a 75-byte prefix match is meaningless.

**Reduce detail-query load (fixes B):**
- **Fetch candidate details lazily** (one at a time or small batches) in the worker, stopping at first match — turns ~20 heavy queries into ~1–3 typically.
- **Dedup candidates by `compilation_id`** so the 20-slot budget holds distinct compilations.
- (Larger) collapse prefix + detail into a single compilation-keyed query.

**Handle the error (fixes C):**
- Detect pg error code `57014` and map it to a specific, user-friendly custom error code instead of a generic `internal_error`.

**Operational:**
- Run bulk similarity backfills against a read replica.

manuelwedler:
## Summary

Fixes the similarity API's production DB load and timeouts (#2891) in four commits:

- `perf(server)`: similarity candidates are keyed on compilation id; payloads are fetched in batches of 5 with early exit on the first match, instead of 20 full payloads upfront via 20 parallel (chain, address) queries. Creation-tx data is resolved once in the main thread; the verification worker no longer performs RPC calls.
- `feat(server)`: pg statement timeouts (SQLSTATE 57014) from the similarity queries surface as `similarity_search_timeout` instead of an opaque `internal_error`, making the residual timeout rate measurable after deployment.
- `feat(database)`: new `compiled_contracts_runtime_code_prefixes` side table (first 75 bytes of each compilation's runtime code) with an insert trigger, plus an idempotent, resumable backfill script for the existing ~5.2M compilations.
- `feat(server)`: candidate search reads the prefix table — every index entry is a candidate, so `LIMIT 20` short-circuits after ~20 entries instead of walking 142k `code` rows (36s measured in production for a V3-pool bytecode). Bytecodes shorter than 75 bytes are rejected upfront with the new `bytecode_too_short_for_similarity` (400). Library call protection is normalized on the search input so deployed libraries can match their compilations.

Full analysis and production measurements in #2891 (see the plan-summary comment). Related planner-statistics findings: #2917.

## Deployment (order matters)

1. Apply migration `20260803100000_add_compiled_contracts_runtime_code_prefixes.sql` (fast, metadata-only; the trigger covers new compilations from that moment).
2. Run `services/database/schema-updates/backfill-compiled-contracts-runtime-code-prefixes.mjs` (~5.2M rows, ~10–15 min; idempotent and resumable, ends with ANALYZE). See the table in `services/database/README.md`.
3. Deploy the server (3.18.0). Deploying earlier is safe — similarity just sees fewer candidates until the backfill completes.

`idx_code_code_first_75` is intentionally kept: the old query needs it until the new server is deployed, and it is our rollback insurance afterwards. It is dropped in a follow-up migration once the new query is verified in production.

## Notes

- The release CHANGELOG for `sourcify-database` should carry a WARNING line about running the backfill script (wording in #2891, following the 2.13.0 precedent).
- Do not rewrite the candidate query's `LATERAL ... LIMIT 1` as `EXISTS` — see the comment in `Database.ts` and the production plan in #2891.


kuzdogan:
The PR also makes me realize the way we store code is inefficient, which was the main reason behind the extremely poor performing query. We have for instance 142,872 code rows for the onchain bytecode of different Uniswap V3-style pools with different immutables. These could have been just a single row with a shared "executional bytecode" code row (ie. excluding transformation snippets), and then just applied Transformations where needed. 

In essance if we are matching onchain to compiled, we shouldn't really need to store both. We should just store the template (compiled) and the applied transformations which would give you the onchain bytecode. 

However, I also realize there's a tradeoff there. Now you can't easily get code hashes or raw bytecodes. This might be useful for data analysis etc. when you are doing work on direct bytecode. 

So not necessarily something we have to fix but to just a thought and something to keep in mind

services/server/src/server/services/utils/Database.ts
```diff
@@ -263,28 +266,116 @@ ${
...
+        SELECT prefixes.compilation_id
+        FROM ${this.schema}.compiled_contracts_runtime_code_prefixes prefixes
+        CROSS JOIN LATERAL (
+          SELECT 1
+          FROM ${this.schema}.verified_contracts
+          JOIN ${this.schema}.sourcify_matches ON sourcify_matches.verified_contract_id = verified_contracts.id
+          WHERE verified_contracts.compilation_id = prefixes.compilation_id
+          LIMIT 1
+        ) verified
+        WHERE prefixes.runtime_code_prefix = substring($1::bytea FROM 1 FOR ${SIMILARITY_PREFIX_LENGTH_BYTES})
```
kuzdogan, line 300: Do we really need to check a compilation being tied to a `sourcify_mathces` row? Couldn't it be just 

```sql
SELECT compilation_id
FROM compiled_contracts_runtime_code_prefixes
WHERE runtime_code_prefix = substring($1::bytea FROM 1 FOR 75)
LIMIT $2
```

Any verification that could similarity-verify this contract should be fine

services/server/src/server/services/VerificationService.ts
```diff
@@ -361,68 +384,236 @@ export class VerificationService {
       );
     }
 
+    // Similarity candidates are indexed by their first 75 bytes of runtime
+    // code; shorter bytecodes (e.g. minimal proxies) are too weak a
+    // discriminator and are not indexed at all.
+    const runtimeBytecodeLengthBytes = (runtimeBytecode.length - 2) / 2;
+    if (runtimeBytecodeLengthBytes < SIMILARITY_PREFIX_LENGTH_BYTES) {
+      throw new BytecodeTooShortForSimilarityError(
```
kuzdogan, line 395: The 75-byte minimum is a new precondition on this endpoint, but it isn't documented anywhere a client would look. `apiv2.yaml` currently describes it as:

> Starts a verification job that searches the Sourcify database for contracts whose runtime bytecode is similar to the contract deployed at the given address. The job will attempt to verify against each candidate until a match is found or the candidate list is exhausted.

Nothing there hints that a short contract is rejected outright. Suggest appending to that description:

```yaml
Candidates are found by matching the first 75 bytes of runtime bytecode, so the
contract must be at least 75 bytes long. Shorter bytecode — for example EIP-1167
minimal proxies, which are 45 bytes — is rejected with
`400 bytecode_too_short_for_similarity`.
```

Optionally also give this endpoint's `"400"` an inline example for the new code rather than the shared `BadRequestResponse` ref, matching how its 409 and 429 are already spelled out. Not required — the spec never enumerates error codes — but it is the house style here.

Related, on the same check: EIP-1167 minimal proxies are 45 bytes, so they now get a 400. Under the old query `substring(code FROM 1 FOR 75)` on both sides compared a 45-byte contract *in full*, which is an exact match rather than a weak one. The comment above calls sub-75-byte prefixes "too weak a discriminator" — true in general, but not for that case. If dropping proxy support is a deliberate trade, worth saying so explicitly.

services/server/src/server/services/utils/Database.ts
```diff
@@ -263,28 +266,116 @@ ${
...
+        LEFT JOIN ${this.schema}.code as recompiled_runtime_code ON recompiled_runtime_code.code_hash = compiled_contracts.runtime_code_hash
+        LEFT JOIN ${this.schema}.code as recompiled_creation_code ON recompiled_creation_code.code_hash = compiled_contracts.creation_code_hash
+        CROSS JOIN LATERAL (
+          SELECT sourcify_matches.metadata
+          FROM ${this.schema}.verified_contracts
+          JOIN ${this.schema}.sourcify_matches ON sourcify_matches.verified_contract_id = verified_contracts.id
+          WHERE verified_contracts.compilation_id = compiled_contracts.id
+          LIMIT 1
+        ) sourcify_matches
+        WHERE compiled_contracts.id = ANY($1::uuid[])
```
kuzdogan, line 347: It feels a lot of the fields are unnecessary here. Sources, metadata etc. Because in essance we check the onchain vs the compiled bytecode (using transformations etc.) and then link the compilation to the deployment. So we don't need to insert sources, metadata etc. anywhere. Except maybe if we need to insert the metadata at the sourcify_matches table.

I didn't follow the code path fully but in theory these should not be necessary. If necessary in a way, it's ok to keep these. Not a blocker.

services/server/src/server/services/utils/Database.ts
```diff
@@ -263,28 +266,116 @@ ${
...
+   *
+   * Everything a similarity candidate needs is compilation-level, so this keys
+   * on compiled_contracts.id rather than looking each candidate up by
+   * (chain, address) as getSourcifyMatchByChainAddressWithProperties does.
+   * contract_deployments is not involved at all.
+   *
+   * Two details worth knowing before editing:
+   *
+   * - Sources come from a scalar subquery rather than a JOIN + aggregate, so no
+   *   GROUP BY is needed. That matters for a batch query: metadata is a `json`
```
kuzdogan: This and the comment above needs some trimming. We can link to this issue when necessary

kuzdogan (commented):
lgtm thanks.

In general good enough to merge but left couple commetns

services/server/src/server/services/utils/Database.ts
manuelwedler: The filter is intentional. A compilation without a `sourcify_matches` row is stale: re-verification (e.g. `replace-contract`) moved the match to a new `verified_contracts` row, typically because the old verification was wrong. We should not resurrect it as a candidate.

There is also a practical coupling: `getCompilationsByIds` reads `sourcify_matches.metadata` through a `CROSS JOIN LATERAL`, which drops candidates without a match row. Without this filter, such candidates would take `LIMIT` slots here and then silently vanish in the batch fetch. So dropping the filter would also require reworking the payload query, not just simplifying this one.

The check is cheap -- at most `limit` indexed `LIMIT 1` probes; it is part of the measured plan in #2891.

services/server/src/server/services/VerificationService.ts
manuelwedler: Documented in f458dbaf: the endpoint description now states the 75-byte precondition, and the 400 response is spelled out inline with a `bytecode_too_short_for_similarity` example (plus the generic `unsupported_chain` one).

On the EIP-1167 point: dropping sub-75-byte support is deliberate, and the code comment now says so instead of the imprecise "too weak a discriminator" wording. Under the old query a short contract could only ever match byte-identical runtime code -- for a minimal proxy that means a verified compilation with the same implementation address baked in, which is rare. Keeping that case would mean storing variable-length prefixes in the new table; we prefer the fixed 75-byte invariant and a clear 400 over silently returning no candidates.

services/server/src/server/services/utils/Database.ts
manuelwedler: All of these fields are used -- the similarity flow deliberately reuses the generic verification and store paths:

- `std_json_output`, the cbor auxdata columns, `version` and `fully_qualified_name` feed the `PreRunCompilation` for the bytecode comparison (no recompilation in the normal case).
- `std_json_input` incl. sources is needed by the verification logic itself: when the auxdata differs from the onchain code (the normal case for a similarity candidate), `Verification` tries to find perfect metadata from the sources and really recompiles -- that is what can upgrade a candidate to an `exact_match`. The Solidity viaIR bug workaround also forces a real recompilation. On top, the store path writes sources/settings (deduped via `ON CONFLICT`) and RepositoryV2 writes the source files.
- `metadata` fills the new `sourcify_matches` row and restores Vyper metadata, which the Vyper compiler output lacks.

Trimming the payload would need a bytecode-compare-only verification mode plus a link-only store path in every backend, and it would cost us the exact-match upgrade -- so keeping the fields is the cheaper and better option.

services/server/src/server/services/utils/Database.ts
manuelwedler: Trimmed both comment blocks in b9e9a889 -- the analysis and production measurements now live behind a link to #2891. I kept the parts an editor must know before touching the queries: the stale-compilation reason for the LATERAL, the "do not rewrite as EXISTS" warning, the `json` GROUP BY limitation, and the `sourcify_matches` alias trick.

kuzdogan (approved):

PR #2920, opened 2026-08-06
fix(compilers): sandbox native solc cwd to stop import path leaks

0xoasis:
## Summary

Native `solc --standard-json` still resolves missing imports against the process cwd (allowed base path `"."`). From **solc 0.8.8** onward, a user-supplied contract can `import` relative host files (e.g. `./.env`, `package.json`); `ParserError.formattedMessage` then echoes file contents through `compiler_error` API responses.

- `solc <= 0.8.7` rejects the same imports as outside allowed directories (no content leak)
- solc-js is unaffected (no filesystem import callback)
- Production Linux prefers the native binary for solc `>= 0.4.10`, so this path is reachable in normal verification

### Fix

- Run native **solc** in an empty temp directory and delete it after compilation
- Resolve the solc binary path before changing cwd so a relative `solcRepo` still works

### Out of scope

**Vyper is not changed.** Probing `vyper --standard-json` (0.3.7 / 0.3.10 / 0.4.0) with HOST cwd bait files vs empty cwd showed:
- no host-file content in compiler errors
- HOST vs EMPTY behavior identical (`ModuleNotFound` / cannot locate interface)
- so this is a solc-specific issue, not applied to Vyper

## Validation

```bash
# from packages/compilers
npx mocha --exit --grep 'Should not leak host cwd files' test/solidityCompiler.spec.ts
npx mocha --exit test/solidityCompiler.spec.ts
```

### DIFF
--- a/packages/compilers/src/lib/common.ts
+++ b/packages/compilers/src/lib/common.ts
@@ -78,11 +78,21 @@ export function createCompilerTimeoutError(timeoutMs: number): Error & {
   return timeoutError;
 }
 
+export type AsyncExecOptions = {
+  /**
+   * Working directory for the compiler subprocess.
+   * Used by native solc to isolate import resolution from the host process cwd
+   * (solc >=0.8.8 may read relative imports from base path ".").
+   */
+  cwd?: string;
+};
+
 export function asyncExec(
   command: string,
   inputStringified: string,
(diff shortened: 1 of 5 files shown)


marcocastignoli (approved):
lgtm!

Issue #213, opened 2020-07-14
Recreating Solidity's metadata from a subset of the original contracts


alcuadrado:
Hi,

I'm opening this issue to better understand how Sourcify works, so that we can eventually integrate it in Buidler. In particular, I need to know if the `metadata` can be recreated without some of the contracts that were compiled together when creating the deployment build.

I think this is easier to explain with an example. So let's suppose this happens:

* I have a project where:
  * `A.sol` defines the contract `A`
  * `B.sol` imports `A.sol` and defines the contract `B`, which extends `A`
  * `C.sol` defines the contract `C`
* Those files are compiled together
* `B` is deployed

Now, if I want to verify them, it's clear that I can extract the `metadata` from the compilation output to do it. But if I loose it and need to recreate it, what would happen in these cases:

1. I only compile `A.sol` and `B.sol`.
2. I compile `A.sol`, `B.sol`, and `D.sol`, a new file that defines `D` and doesn't import `A.sol` nor `B.sol`.
2. I compile `A.sol`, `B.sol`, and `E.sol`, a new file that imports `A.sol` and defines `E`, which extends `A`.

Of course, the same compiler version and settings are used in all the cases.

Would I be able to extract `B`'s `metadata` in those cases and still verify them? Would it be an identical `metadata`?

Thanks!

chriseth:
For each contract, the Solidity compiler deduces a subset of the sources that are required to compile the contract and only includes those in the metadata. So for `C`, you only need `C.sol` and for `B` you need `A.sol` and `B.sol`. How many files you compile together in one run should not matter, but it does matter what files you import: If `B.sol` imports another file, that file is needed even if the contract `B` does not use any of the imported symbols.

Did you take a look at the metadata in each of the cases?

alcuadrado:
Thanks for your reply, Chris! 

I did wrote these test cases, but wasn't sure if that was guaranteed by how the metadata field works, or if it was just incidental because of the compiler internals.

This confirmation was all I needed. Thanks!

PR #2931, opened 2026-08-18
fix: detect Polygon (Matic) UpgradableProxy in proxyResolution

### LINKED ISSUE #2923: Proxy detection misses Polygon (Matic) UpgradableProxy — e.g. Polygon DAI
# Proxy detection misses Polygon (Matic) `UpgradableProxy` — e.g. Polygon DAI

## Summary

`proxyResolution` returns `isProxy: false` for contracts using Polygon's own proxy pattern, which stores the implementation at `keccak256("matic.network.proxy.implementation")` instead of the EIP-1967 slot.

Example — Polygon DAI (`(PoS) Dai Stablecoin`), chain 137, `0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063`:

https://repo.sourcify.dev/137/0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063
https://polygonscan.com/address/0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063#code

```
GET /v2/contract/137/0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063?fields=proxyResolution

{"proxyResolution":{"isProxy":false,"proxyType":null,"implementations":[]}, ...}
```

But reading the slot directly:

```
slot 0xbaab7dbf64751104133af04abc7d9979f0fda3b059a322a8333f533d3f32bf7f
  → 0x490e379c9cff64944be82b849f8fd5972c7999a7
```

The implementation is itself fully verified on Sourcify (matchId `5195332`), so we have the real ABI — we just never point at it. Consumers get the proxy's ABI (`UChildERC20Proxy`) instead of DAI's.

## Root cause

Three separate problems. The first two are upstream in [whatsabi](https://github.com/shazow/whatsabi); the third is how we call it.

### 1. whatsabi doesn't know the Polygon slot

`slotResolvers` in `src/proxies.ts` covers EIP-1967, ZeppelinOS, PROXIABLE, Gnosis Safe and Diamond. No Polygon entry. Latest whatsabi (0.27.0) still has none.

### 2. The slot value never appears in the bytecode

Even with a table entry this wouldn't fire. The proxy is Solidity 0.6.6:

```solidity
bytes32 constant IMPLEMENTATION_SLOT = keccak256("matic.network.proxy.implementation");
```

solc 0.6.6 does not fold this into a constant — it inlines the expression, so the deployed code carries the **ASCII string** and runs `KECCAK256` on it at execution time. The hash `0xbaab7dbf…` is never pushed.

whatsabi detects proxies purely by literal matching: `PUSH32` operands against known slot hashes (`src/disasm.ts:263`), plus a scan of the trailing data segment for the same values (`src/disasm.ts:544-545`). The string is right there in the data segment, but the table is keyed by hash, not pre-image.

### 3. Two CBOR bugs hide the string from the scan that could find it

Both in the aux-data block at `src/disasm.ts:525-548`:

```ts
const cborLength = valueToOffset(code.bytecode.slice(-2));
endBoundary = -(cborLength + 4); // +4 for the length bytes
```

**(a) Off-by-two.** The metadata suffix is the CBOR blob plus a **2-byte** length, so the trim should be `cborLength + 2`. `code.bytecode` is a `Uint8Array` (`src/disasm.ts:22`), so these are byte units — `+ 4` discards two bytes of real data on every contract. The `4` looks like the hex-character count of a 2-byte length, i.e. a units mix-up.

Usually harmless. Here it is fatal, because the string sits flush against the metadata:

```
total runtime bytes:        2949
CBOR length (last 2 bytes):   51
real metadata starts at:    2896
whatsabi cuts off at:       2894   ← 2 bytes early
matic string occupies:      2862 .. 2895

scanned tail: "...DESTINATION_ADDRESS_IS_NOT_A_CONTRACTmatic.network.proxy.implementati"
```

It clips the `on`.

**(b) The last two bytes are assumed to be a CBOR length.** True for runtime bytecode. Not true for **creation** bytecode, whose tail is constructor arguments. We pass creation bytecode preferentially (`lookup.handlers.ts:112`), so this is the path that actually runs:

```
last 32 bytes: 000000000000000000000000490e379c9cff64944be82b849f8fd5972c7999a7
last 2 bytes (0x99a7) read as cborLength: 39335
endBoundary: -36102  →  slice(3046, -36102) = empty
```

The scan is dead on arrival. Note what those constructor args are: `0x490e…99a7`, the implementation address, sitting in plain sight in the input we hand whatsabi.

The existing guard (`if (!isHalt(finalByte))`) doesn't help — an arbitrary argument byte passes it.

## Verification

Reproduced whatsabi's detection with Sourcify's exact options (`abiLoader: false, signatureLookup: false, followProxies: false`) against the real bytecode:

| input | whatsabi | detected |
|---|---|---|
| runtime | pristine | ✗ |
| runtime | `+2` trim, pre-image entry | ✓ → `0x490e…99a7` |
| creation | pristine | ✗ |
| creation | `+2` trim, pre-image entry | ✗ |
| creation | + CBOR plausibility guard | ✓ → `0x490e…99a7` |

All three changes are needed for the path we actually use.

### Upstream (whatsabi) — two PRs

**PR 1 — CBOR boundary fixes.** Self-contained, defensible independent of Polygon:

- `cborLength + 4` → `cborLength + 2`
- Only treat the tail as metadata if it is plausible: the implied blob must fit inside the data segment **and** start with a CBOR map header (solc emits `0xa1`/`0xa2`/`0xa3`; verified `0xa2 "ipfs" … "solc" 0.6.6` on this contract). Otherwise don't trim at all — failing open is right for a best-effort scan.

**PR 2 — Polygon proxy support:**

- Add `MATIC_IMPL` to `slots` and a `MaticProxyResolver` (a copy of `ZeppelinOSProxyResolver` with a different slot). This alone fixes Polygon proxies built with a newer solc that *does* fold the constant.
- Add a `slotPreimages` table and match it in the same aux-data scan, for the older runtime-hashing ones.

Not doing the fully generic version (hash every printable string, look it up) on purpose: the strings are packed with no separators (`…IS_NOT_A_CONTRACTmatic.network.proxy.implementation…`), so recovering boundaries means hashing every substring — O(n²) keccaks per contract, far too slow for our per-request path.

Related upstream issue: shazow/whatsabi#67 (open since Dec 2023, same family — custom-slot proxies).

### Sourcify

- Bump whatsabi once the above land (note: `services/server/package.json` declares `0.27.0` but the installed tree here is `0.26.0`).
- Add `"MaticProxy"` to the `ProxyType` union in `proxy-contract-util.ts:5-13`. The resolution loop (`:106-123`) is already generic and `SourcifyChain.getStorageAt` exists (`SourcifyChain.ts:797`), so nothing else changes.

## Impact and rollout

`proxyResolution` is computed per request (`lookup.handlers.ts:102-158`), not persisted — the DB mapping in `database-util.ts` stores only the *input* bytecode. So a whatsabi bump fixes existing contracts immediately, with no backfill.

Worth landing the detector fix **before** proxy detection moves into verification and the result becomes stored state (per the TODOs at `lookup.handlers.ts:97-101` and `database-util.ts:582`), otherwise this needs a backfill later.

Cost: one extra `eth_getStorageAt`, only for contracts where detection fires.

Not limited to DAI. `UpgradableProxy` is the standard wrapper for bridged assets on Polygon PoS. Spot-checking the major ones — reading the Matic slot directly vs. what we report:

| token | address | matic slot | our `isProxy` |
|---|---|---|---|
| USDC (PoS) | `0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174` | `0xdd9185db…2226` | `false` |
| USDT (PoS) | `0xc2132D05D31c914a87C6611C10748AEb04B58e8F` | `0x90040487…4229` | `false` |
| WBTC (PoS) | `0x1BFD67037B42Cf73acF2047067bd4F2C47D9BfD6` | `0x7ffb3d63…c1e2` | `false` |
| DAI (PoS)  | `0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063` | `0x490e379c…99a7` | `false` |
| WETH (PoS) | `0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619` | empty (not this pattern) | `false` |

Four of the five checked are proxies we report as non-proxies. A full sweep of chain 137 would give the real count.

## Tracker

- [x] whatsabi PR 1 — CBOR trim off-by-two + plausibility guard (landed in whatsabi 0.28.0, shazow/whatsabi#211 + #213)
- [x] whatsabi PR 2 — `MATIC_IMPL` slot, `MaticProxyResolver`, `slotPreimages` (landed in whatsabi 0.28.0, shazow/whatsabi#214)
- [x] Bump whatsabi in `services/server` (0.27.0 → 0.28.0, #2931)
- [x] Add `"MaticProxy"` to `ProxyType` (#2931)
- [x] Regression test on both runtime and creation bytecode (#2931)
- [ ] Sweep for other custom-slot proxy families worth adding

kuzdogan:
## Summary

`proxyResolution` returned `isProxy: false` for contracts using Polygon's own proxy pattern, which stores the implementation at `keccak256("matic.network.proxy.implementation")` instead of the EIP-1967 slot. This affects bridged assets on Polygon PoS — DAI, USDC, USDT, WBTC, etc. Consumers got the proxy's ABI (`UChildERC20Proxy`) instead of the token's.

whatsabi [0.28.0](https://github.com/shazow/whatsabi/releases/tag/0.28.0) landed both fixes this needed:

- **Polygon (Matic) `UpgradableProxy` support** (shazow/whatsabi#214) — new `MaticProxyResolver` + `slotPreimages` scan for the string that older solc leaves in the aux-data segment.
- **CBOR metadata boundary fix** in the aux-data proxy slot scan (shazow/whatsabi#211, #213) — needed for detection on creation bytecode, which is the input Sourcify passes preferentially.

This PR bumps whatsabi and wires the new proxy type through Sourcify.

## Changes

- Bump `@shazow/whatsabi` `0.27.0` → `0.28.0` in `services/server`. (The installed tree was actually stale at `0.26.0`; the lockfile now genuinely resolves `0.28.0`.)
- Add `"MaticProxy"` to the `ProxyType` union in `proxy-contract-util.ts` and to the `ProxyType` enum in the OpenAPI spec `apiv2.yaml`.
- Add regression tests on both **runtime** and **creation** bytecode, using the real Polygon DAI (chain 137) contract `0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063`.

The resolution loop was already generic — it reads `proxy.name` and calls `getStorageAt` — so no detection logic changed.

## Notes

`proxyResolution` is computed per request and not persisted (the DB stores only the input bytecode), so this fixes existing contracts immediately with **no backfill**. Cost is one extra `eth_getStorageAt`, only for contracts where detection fires.

Closes #2923


services/server/test/unit/utils/proxy-bytecodes.json
```diff
@@ -5,5 +5,7 @@
...
...2b4d4705b57b16e0caae994e214680ee281610988610a08565b604051808373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1681526020018273ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1681526020019250505060405180910390a1610a0581610ac7565b50565b600080600060405180807f6d617469632e6e6574776f726b2e70726f78792e6f776e65720000000000000081525060190190506040518091039020905080549150819250505090565b60008073ffffffffffffffffffffffffffffffffffffffff168273ffffffffffffffffffffffffffffffffffffffff161415610a905760009050610a9e565b6000823b9050600081119150505b919050565b60006040518080610b2e602291396022019050604051809103902090508181555050565b600060405180807f6d617469632e6e6574776f726b2e70726f78792e6f776e657200000000000000815250601901905060405180910390209050818155505056fe44455354494e4154494f4e5f414444524553535f49535f4e4f545f415f434f4e54524143546d617469632e6e6574776f726b2e70726f78792e696d706c656d656e746174696f6ea26469706673582212208262e9eb2153cf52cdbce21b7ed8a72faf3469a061b389bcfdc789246642006364736f6c634300060600336d617469632e6e6574776f726b2e70726f78792e696d706c656d656e746174696f6e000000000000000000000000490e379c9cff64944be82b849f8fd5972c7999a7"
```
manuelwedler: The name `MaticProxyWithSlotOnlyInCreationCode` is not correct. The string `matic.network.proxy.implementation` occurs two times in this bytecode. One occurrence is in the embedded runtime code. One occurrence is in the constructor data segment. Thus, the slot reference is not only in the creation code. This is different from the `EIP1967ProxyWithSlotOnlyInCreationCode` fixture. In that fixture, the slot reference is only in the constructor code.

Recommendation: Change the name to `MaticProxyCreationBytecode`. The test examines the creation-bytecode path with the constructor-arguments tail. The new name will agree with that.

---

manuelwedler (approved):
lgtm

PR #2933, opened 2026-08-19
fix(compilers): jail Fe source writes to the compilation temp dir

SashaMIT:
## Summary

`useFeCompiler` wrote `path.join(tmpDir, sourcePath)`. A Fe verify body whose source keys start with `src/` (the API prefix check) can still be `src/../../../tmp/...`, which `path.join` treats as a write outside the temp ingot.

The attacker already controls `stdJsonInput.sources` keys on the public verify API. The `src/` prefix does not dominate `..`.

Jail is resolve-under-dir fail-closed in `@ethereum-sourcify/compilers` (covers library callers) and the same class is rejected in `validateAndNormalizeFeInput`.

## Test plan

- [x] `mocha --grep 'escape|resolve Fe source'` in `packages/compilers` (2/2)
- [x] Revert-tested: without the prefix check, `Should resolve Fe source paths under the compilation directory` fails (`expected [Function] to throw`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/compilers/src/lib/feCompiler.ts
+++ b/packages/compilers/src/lib/feCompiler.ts
@@ -32,21 +32,14 @@ export function resolveFeSourcePath(root: string, sourcePath: string): string {
   const prefix = rootResolved.endsWith(path.sep)
     ? rootResolved
     : rootResolved + path.sep;
-  if (resolved !== rootResolved && !resolved.startsWith(prefix)) {
+  // A key must name a file under the root. `src/..` resolves to the root
+  // itself and would otherwise hit EISDIR on write.
+  if (!resolved.startsWith(prefix)) {
     throw new Error(
       `Fe source path escapes the compilation directory: ${sourcePath}`,
     );
   }
   return resolved;
-}
-
-function assertFeSourcesStayInside(
-  root: string,
-  sources: FeJsonInput['sources'],
-): void {
-  for (const sourcePath of Object.keys(sources)) {
-    resolveFeSourcePath(root, sourcePath);
-  }
 }
 
 /**
@@ -171,13 +164,6 @@ export async function useFeCompiler(
   if (!fePlatform) {
     throw new Error('Fe compiler is not supported on this machine.');
   }
-
-  // Fail closed before downloading the binary. Public verify accepts Fe
-  // source keys; the API src/ prefix check does not stop src/../...
-  assertFeSourcesStayInside(
-    path.resolve(os.tmpdir(), 'sourcify-fe-jail-root'),
-    feJsonInput.sources,
-  );
 
   const fePath = await getFeExecutable(feRepoPath, fePlatform, version);
 
--- a/packages/compilers/src/lib/feCompiler.ts
+++ b/packages/compilers/src/lib/feCompiler.ts
@@ -17,6 +17,39 @@ const HOST_FE_REPO = 'https://github.com/argotorg/fe/releases/download/';
 const MINIMUM_FE_VERSION = '26.0.0-alpha.12';
 
 /**
+ * Resolve a Fe source key under `root`. `path.join` treats an absolute
+ * segment as a new root, and `src/../..` walks out of the temp ingot.
+ */
+export function resolveFeSourcePath(root: string, sourcePath: string): string {
+  if (sourcePath.includes('\0')) {
+    throw new Error(`Fe source path contains a null byte: ${sourcePath}`);
+  }
+  if (path.isAbsolute(sourcePath)) {
+    throw new Error(`Fe source path must be relative: ${sourcePath}`);
+  }
+  const resolved = path.resolve(root, sourcePath);
+  const rootResolved = path.resolve(root);
+  const prefix = rootResolved.endsWith(path.sep)
+    ? rootResolved
+    : rootResolved + path.sep;
+  if (resolved !== rootResolved && !resolved.startsWith(prefix)) {
+    throw new Error(
+      `Fe source path escapes the compilation directory: ${sourcePath}`,
+    );
+  }
+  return resolved;
+}
+
+function assertFeSourcesStayInside(
+  root: string,
+  sources: FeJsonInput['sources'],
+): void {
+  for (const sourcePath of Object.keys(sources)) {
+    resolveFeSourcePath(root, sourcePath);
+  }
+}
+
+/**
  * Returns the platform-specific asset name for the Fe binary.
  * Asset names follow the pattern: fe_{os}_{arch}[.exe]
  */
@@ -139,6 +172,13 @@ export async function useFeCompiler(
     throw new Error('Fe compiler is not supported on this machine.');
   }
 
+  // Fail closed before downloading the binary. Public verify accepts Fe
+  // source keys; the API src/ prefix check does not stop src/../...
+  assertFeSourcesStayInside(
+    path.resolve(os.tmpdir(), 'sourcify-fe-jail-root'),
+    feJsonInput.sources,
+  );
+
   const fePath = await getFeExecutable(feRepoPath, fePlatform, version);
 
   // Create a unique temp directory to avoid collisions from parallel compilations
@@ -153,7 +193,7 @@ export async function useFeCompiler(
 
     // Write source files (sourcePaths already have src/ prefix, e.g. 'src/lib.fe')
     for (const [sourcePath, source] of Object.entries(feJsonInput.sources)) {
-      const fullPath = path.join(tmpDir, sourcePath);
+      const fullPath = resolveFeSourcePath(tmpDir, sourcePath);
       await fs.promises.mkdir(path.dirname(fullPath), { recursive: true });
       await fs.promises.writeFile(fullPath, source.content);
     }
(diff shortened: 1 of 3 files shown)


packages/compilers/src/lib/feCompiler.ts
```diff
@@ -139,6 +172,13 @@ export async function useFeCompiler(
     throw new Error('Fe compiler is not supported on this machine.');
   }
 
+  // Fail closed before downloading the binary. Public verify accepts Fe
+  // source keys; the API src/ prefix check does not stop src/../...
+  assertFeSourcesStayInside(
+    path.resolve(os.tmpdir(), 'sourcify-fe-jail-root'),
+    feJsonInput.sources,
+  );
```
manuelwedler: This pre-check is redundant and I'd drop it (together with `assertFeSourcesStayInside`) to keep the change minimal. The jail in the write loop below (`resolveFeSourcePath(tmpDir, sourcePath)`) already fails closed before any byte is written, and since escape detection for relative keys is root-independent, this call can never catch anything the loop wouldn't.

Validating against `sourcify-fe-jail-root` — a directory that never exists and isn't the real `tmpDir` — is also confusing to the next reader, who has to convince themself the result transfers to the real root.

The only thing the pre-check buys is erroring before `getFeExecutable`, but the binary is cached after the first download, so that's a one-time cost. If failing before the download is considered important, moving the `mkdtemp` up and validating against the real `tmpDir` would achieve it without the synthetic root.

---

packages/compilers/src/lib/feCompiler.ts
```diff
@@ -16,6 +16,39 @@ import type { JsonFragment } from 'ethers';
...
+  }
+  if (path.isAbsolute(sourcePath)) {
+    throw new Error(`Fe source path must be relative: ${sourcePath}`);
+  }
+  const resolved = path.resolve(root, sourcePath);
+  const rootResolved = path.resolve(root);
+  const prefix = rootResolved.endsWith(path.sep)
+    ? rootResolved
+    : rootResolved + path.sep;
+  if (resolved !== rootResolved && !resolved.startsWith(prefix)) {
```
manuelwedler: The `resolved !== rootResolved` carve-out allows a key that resolves exactly to the root, e.g. `src/..`. That key also slips through the middleware check (it posix-normalizes to `.`, which is neither absolute nor `..`-prefixed), so it reaches the write loop and fails with a raw `EISDIR` from `fs.writeFile(tmpDir)` instead of a clear validation error.

A source key must name a file strictly inside the root, so rejecting equality too would be both simpler and stricter:

```ts
if (!resolved.startsWith(prefix)) {
```

---

services/server/src/server/apiv2/middlewares.ts
```diff
@@ -250,6 +251,30 @@ export function validateAndNormalizeFeInput(
...
+    if (
+      path.posix.isAbsolute(normalized) ||
+      normalized === ".." ||
+      normalized.startsWith("../")
+    ) {
+      throw new InvalidParametersError(
+        "Fe source paths must stay under the compilation directory.",
+      );
+    }
+  }
```
manuelwedler: Strictly speaking this layer isn't needed once the compiler-side jail is in — the write in `useFeCompiler` fails closed regardless. It's still worth keeping in my view, because it turns the rejection into a clean 400 `InvalidParametersError` at request time instead of a failed verification job. But if the goal is a minimal diff, this block is the optional half of the PR.

If it stays, two consistency points:

1. Its rules diverge from `resolveFeSourcePath`: this check treats `\` as a path separator and posix-normalizes, while the compiler jail doesn't — so the two layers accept slightly different key sets (e.g. `src/..` passes here but hits an fs error in the compiler, see the comment there). Since the server already depends on `@ethereum-sourcify/compilers` and the PR exports `resolveFeSourcePath`, aligning the two (or at least a comment noting the middleware is intentionally stricter) would avoid drift.
2. Nit: `const normalized` here shadows the `normalized` sources record declared just above (line 246), and the lead-in comment reads garbled — "does not stop src/../... path.join on the compiler treats that as" is missing a sentence break.

---

packages/compilers/test/feCompiler.spec.ts
```diff
@@ -41,6 +44,50 @@ describe('Verify Fe Compiler', () => {
...
+      thrown = error;
+    }
+
+    expect(thrown, 'expected the compilation to reject').to.be.instanceOf(
+      Error,
+    );
+    expect((thrown as Error).message).to.include(
+      'escapes the compilation directory',
+    );
+    expect(fs.existsSync(marker)).to.equal(false);
```
manuelwedler: The marker assertion is checking a path that doesn't necessarily match the injected key: the key hardcodes `.../tmp/sourcify-fe-escape-probe` while `marker` is built from `os.tmpdir()`. On a machine where `TMPDIR` isn't `/tmp` (macOS: `/var/folders/...`), the two point at different files and `existsSync(marker)` passes trivially. Also, `useFeCompiler` never creates its temp dir under the escape scenario anymore, so the assertion mostly documents intent rather than guarding the write.

Suggestion: either derive the traversal key from `os.tmpdir()` so the marker really is the write target, or drop the marker bookkeeping and keep the simpler "throws with 'escapes the compilation directory'" assertion, which is the part doing the work (and is already covered again by the unit test below).

---

packages/compilers/src/lib/feCompiler.ts
manuelwedler: Remove this pre-check and the helper `assertFeSourcesStayInside`. The reasons:

1. The check in the write loop is sufficient. It stops the compilation before the code writes one byte.
2. Escape detection for a relative key does not change with the root directory. Thus this call cannot find a problem that the loop does not find.
3. The root `sourcify-fe-jail-root` does not exist. It is not the real `tmpDir`. This can confuse the reader.

The pre-check only moves the error before `getFeExecutable`. But the binary is in the cache after the first download. If an early error is important: create the temp directory first, then do the check on the real `tmpDir`.

---

packages/compilers/src/lib/feCompiler.ts
manuelwedler: The condition `resolved !== rootResolved` permits a key that resolves to the root itself. Example: `src/..`.

The middleware also permits this key. There, the key normalizes to `.`. That value is not absolute and does not start with `..`.

The result: `fs.writeFile(tmpDir)` fails with a raw `EISDIR` error, not with a clear validation error.

A source key must point to a file inside the root. Reject the equal path also:

```ts
if (!resolved.startsWith(prefix)) {
```

---

services/server/src/server/apiv2/middlewares.ts
manuelwedler: This block is not necessary for the fix. The jail in `useFeCompiler` stops the write in all cases. But this block gives a clean 400 error at request time, not a failed verification job. Keep it for that behavior. Remove it for the smallest diff.

If the block stays, two points:

1. The rules here are different from `resolveFeSourcePath`. This check changes `\` to `/` and normalizes the path. The compiler jail does not. Thus the two layers accept different key sets. Example: `src/..` passes here, but causes an fs error in the compiler (see the comment there). The server has `@ethereum-sourcify/compilers` as a direct dependency, and the PR exports `resolveFeSourcePath`. Use the same rules in the two places, or add a comment that says the middleware is more strict.
2. The comment text is not clear: "does not stop src/../... path.join on the compiler treats that as". Add the missing sentence break. Also, this loop uses the name `normalized` again. The block above uses the same name for the sources record. The two scopes are separate, but one name with two meanings in one function decreases readability.

---

packages/compilers/test/feCompiler.spec.ts
manuelwedler: The marker path and the injected key can point to different files. The key contains the fixed string `/tmp/sourcify-fe-escape-probe`. The `marker` value comes from `os.tmpdir()`. On macOS, `os.tmpdir()` is `/var/folders/...`. There, `existsSync(marker)` examines a different path, and the assertion passes without value.

Also, the jail throws before the code writes a file. Thus the marker assertion shows intent more than it guards the write.

Two possible corrections:

- Make the traversal key from `os.tmpdir()`. Then the marker is the true write target.
- Or remove the marker steps. Keep the assertion for the message "escapes the compilation directory". That assertion does the work, and the unit test below covers the same case again.

---

packages/compilers/src/lib/feCompiler.ts
SashaMIT: Dropped. The write loop is the jail. No fake root.

packages/compilers/src/lib/feCompiler.ts
SashaMIT: Agreed. src/.. now fails the prefix check, so it is a validation error instead of EISDIR.

services/server/src/server/apiv2/middlewares.ts
SashaMIT: Kept the 400 so a bad key never starts a job. It now calls resolveFeSourcePath, same rules as the write loop.

packages/compilers/test/feCompiler.spec.ts
SashaMIT: Removed the marker. The unit test now covers src/.. and the escape key, and those assertions do the work.

manuelwedler (approved):
Looks good. Thanks for this finding!

PR #2934, opened 2026-08-19
fix(lib-sourcify): find all occurrences of repeated auxdata in bytecode

manuelwedler:
## Summary

Some factory contracts fail verification with `no_match` even though only metadata hashes differ from the on-chain bytecode. This happens when a contract embeds the **same child contract's bytecode multiple times** — for example a factory that embeds a child's creation code in several code paths.

The compiler's `legacyAssembly` lists the child's auxdata **once**, but the bytecode contains it at **multiple positions**. `findAuxdataPositions` mapped each auxdata value from `legacyAssembly` to exactly one position (`if (resultAuxdatas[auxdataKey] !== undefined) continue`). The additional occurrences were never recorded, `extractAuxdataTransformation` left them untouched, and the bytecode comparison failed.

Concrete example: `ZoraCreator1155FactoryImpl` at `0x169d9147dFc9409AfA4E558dF2C9ABeebc020182` on OP Mainnet (solc 0.8.17). The runtime bytecode contains 4 ipfs auxdata blocks (offsets 9377, 10387, 11397, 11514): the factory's own plus the embedded child's auxdata three times. Only 2 positions were found before this change; with it, all 4 are found and the contract verifies as a partial runtime match. The same pattern affects a number of deployments of this factory family across OP Mainnet, Base and Arbitrum.

## Change

`packages/lib-sourcify/src/Compilation/auxdataUtils.ts` — in `findAuxdataPositions`:

- Positions still map 1:1 to the `legacyAssembly` entries first — existing keys and behavior are unchanged, including the handling of multiple identical auxdata entries from #1980.
- When a bytecode diff position hosts an auxdata value that is **already mapped**, it is recorded as an additional occurrence under the next free key (`'3'`, `'4'`, …).
- The security property is preserved: positions are only considered where the edited-source recompilation actually produces a diff, so statically embedded auxdata-like bytes (which do not change when sources are edited) are still not recorded.

`extractAuxdataTransformation` iterates the position entries generically, so the additional occurrences are transformed without further changes.

## Testing

- New unit tests for `findAuxdataPositions` (`test/Compilation/auxdataUtils.spec.ts`): single auxdata, two different auxdatas, **repeated occurrences of one auxdata (regression test — fails without this change)**, multiple identical `legacyAssembly` entries (#1980 case), and no recording at non-diff positions.
- `SolidityCompilation` suite: 14 passing. `Verification` suite: 43 passing.
- End-to-end: the example contract above goes from `no_match` to a partial runtime match.

---
_Generated by [Claude Code](https://claude.ai/code/session_016kwtTmYc4iYJ4cv1JE5vDd)_

### DIFF
--- a/packages/lib-sourcify/src/Compilation/auxdataUtils.ts
+++ b/packages/lib-sourcify/src/Compilation/auxdataUtils.ts
@@ -137,6 +137,11 @@ export function findAuxdataPositions(
 
   // 3.  For every bytecodeDiffPosition, see if it hosts an compilerAuxdataDiff
   const resultAuxdatas: CompiledContractCborAuxdata = {};
+  // An auxdata from `legacyAssembly` can occur more than one time in the bytecode,
+  // e.g. when a contract embeds the same child bytecode in more than one location.
+  // We record each occurrence. Occurrences that go beyond the `legacyAssembly`
+  // entries get the next free keys.
+  let nextExtraAuxdataKey = compilerAuxdataDiffs.length + 1;
   let prevPos = -Infinity;
   for (const pos of bytecodeDiffPositions) {
     // Skip positions that are immediately consecutive to the previous one,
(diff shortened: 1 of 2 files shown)


marcocastignoli (approved):
From my pov lgtm. I gave this to review to Fable xhigh and we also have a confirmation there.

Issue #2916, opened 2026-08-04
Delete wrong Wanchain contracts from HyperEVM (chainId 999)


kuzdogan:
Chain ID 999 is claimed by two different networks: **Wanchain Testnet** (the assignment in [ethereum-lists/chains](https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-999.json), still current) and **HyperEVM Mainnet** (the assignment used by Etherscan and the wider ecosystem). Sourcify served 999 as Wanchain Testnet for years, then silently switched to HyperEVM. As a result, chain 999 today holds a mix of contracts from both networks.

I checked all 2,030 verified contracts on chain 999 against both networks' RPCs. The split is clean and 125 contracts are wrong.

## Summary

| | Count | matchId range | verifiedAt range |
|---|---|---|---|
| **Wanchain Testnet** (wrong) | 125 | `66` → `28822960` | 2024-08-08T10:03:33Z → 2026-05-13T10:11:22Z |
| **HyperEVM Mainnet** (correct) | 1,905 | `29553025` → `43272750` | 2026-05-21T10:12:17Z → 2026-08-04T07:43:13Z |

There is exactly **one transition** — no interleaving, no contracts present on both networks, no contracts present on neither, and all 2,030 addresses are unique.

- Last Wanchain match: `matchId 28822960`, `0x543Cd3Aa98c23783B265D8d7D0307d22C0aC3C27`, 2026-05-13T10:11:22Z
- First HyperEVM match: `matchId 29553025`, `0x6AB31532382Ba5cD5E8b5D343Cf5995906bb8DD8`, 2026-05-21T10:12:17Z

## Why the split exists

Sourcify verifies against whatever bytecode the configured RPC returns, so "which network is this contract on" is really "which RPC was configured that day":

| Period | RPC used for chain 999 | Actual network |
|---|---|---|
| 2023-11 → 2026-05-21 | none configured → fell back to chainid.network's entry, `https://gwan-ssl.wandevs.org:46891/` | Wanchain Testnet |
| 2026-05-21 → now | `https://lb.drpc.org/ogrpc?network=hyperliquid` (from the `sourcify-chains` repo) | HyperEVM Mainnet |

[...]

PR #2940, opened 2026-08-26
fix(scripts): abort BigQuery session on close/interrupt in delete-sourcify-match

kuzdogan:
## Summary

The BigQuery path of `scripts/delete-sourcify-match.ts` had two session bugs that made it fail (and block itself) on any non-trivial run:

- **Session leak.** It opens a BigQuery *session* per invocation, but `close()` was a no-op that relied on auto-cleanup that doesn't happen — sessions persist for ~24h. A batch delete left hundreds of idle sessions behind.
- **Orphaned lock on interrupt.** A run interrupted mid-transaction (Ctrl-C) never reached the `try/finally`, so the session's transaction stayed **open and held a lock** on the target tables. Every later delete then failed with `Transaction is aborted due to concurrent update against table … public_verified_contracts`, naming the same stuck transaction each time.

## Fix

- `close()` now calls `BQ.ABORT_SESSION()` to end the session and roll back any in-progress transaction, instead of relying on non-existent auto-cleanup.
- Register `SIGINT`/`SIGTERM` handlers so an interrupted run rolls back and closes the session before exiting, rather than orphaning a lock-holding transaction.

Postgres behaviour is unchanged (dropping the client already rolls back).

## Context

Surfaced while deleting the 125 wrong-network Wanchain contracts from chain 999 (#2916): interrupted BigQuery runs repeatedly left zombie sessions whose open transactions locked `public_verified_contracts`, blocking all further deletes until each session was manually aborted with `CALL BQ.ABORT_SESSION(...)`.

### DIFF
--- a/scripts/delete-sourcify-match.ts
+++ b/scripts/delete-sourcify-match.ts
@@ -251,8 +251,17 @@ class BigQueryClient implements DatabaseClient {
   }
 
   async close(): Promise<void> {
-    // Session cleanup is handled by BigQuery automatically
-    // when the connection/job completes
+    if (!this.sessionId) return;
+    // BigQuery sessions aren't auto-closed; an open one with an uncommitted
+    // transaction locks the table and blocks later deletes. Abort it explicitly
+    // (also rolls back any in-progress transaction).
+    try {
+      await this.query("CALL BQ.ABORT_SESSION()");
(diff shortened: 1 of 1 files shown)


marcocastignoli (approved):
lgtm

Issue #2924, opened 2026-08-10
The same metadata is stored 8.5 times over: ~139 GB of a 182 GB table


kuzdogan:
## Summary

Every verified contract stores a copy of its compiler metadata. But that metadata describes the **compilation**, not the deployment — so when one compilation is deployed at 3,000 addresses, we store 3,000 identical copies.

Storing one copy per compilation instead would free up about **139 GB**. The `sourcify_matches` table would shrink from 182 GB to roughly 19 GB.

All numbers below were measured against production on 2026-08-10 using a read-only account. The queries are included so anyone can re-run them.

Related: #2891 and #2918 (the similarity-search work that surfaced this), #2917 (planner statistics).

## What the metadata is, and why it's duplicated

When solc compiles a contract it produces, alongside the ABI and the bytecode, a **metadata JSON blob**. It lists the compiler version, the exact settings used, every source file with its hash, and the ABI. Think of it as a receipt for the compilation.

The important thing about that receipt: it depends only on what went *into* the compiler. It says nothing about which chain the contract landed on or what address it got. Two deployments of the same compiled contract have byte-identical metadata.

We store it here:

```ts
// SourcifyDatabaseService.ts:551 — the value comes straight off the compilation
metadata: verification.compilation.metadata as any,
```

...into the `sourcify_matches` table, which has **one row per verified contract address**.

So the receipt for a single compilation gets written once per deployment. A popular proxy compiled once and deployed 3,000 times means 3,000 copies of the same bytes.

The natural home would be `compiled_contracts`, which has one row per compilation. That table has no metadata column, because it follows the shared [Verifier Alliance](https://verifieralliance.org) schema and there's no slot for it there. So it ended up on our own side table instead.

### The table today

Postgres keeps large column values in a separate storage area called **TOAST** (values roughly over 2 KB get moved out of the main table). Since `metadata` is the only large column on `sourcify_matches`, the TOAST size is almost exactly what the metadata costs us:

| | |
|---|---|
| `sourcify_matches` total | **182 GB** |
| ├ main table | 44 GB |
| ├ indexes | 7.4 GB |
| └ **TOAST (large values)** | **130 GB** |

Smaller metadata values stay in the main table rather than going to TOAST — that's another **33 GB**. So:

> **Metadata accounts for 163 GB of the 182 GB table — about 90% of it.**

### How many copies there are

| | |
|---|---|
| Verified contracts (`sourcify_matches` rows) | 42,733,948 |
[...]

PR #2941, opened 2026-08-26
feat: store metadata once per compilation in compiled_contracts_metadata

marcocastignoli:
## Summary

First of two steps for #2924 (option B: a Sourcify-specific side table). Based on #2929 by @peterlodri-sec — same table shape and first-submitter-wins semantics, restructured for a zero-downtime rollout.

- **Migration** creates `compiled_contracts_metadata` (`compilation_id uuid PK → compiled_contracts`, `metadata json NOT NULL`) and nothing else — no in-migration backfill.
- **Backfill** happens out-of-band via `services/database/schema-updates/backfill-compiled-contracts-metadata.mjs`: batched with a keyset cursor, throttled, idempotent (`ON CONFLICT DO NOTHING`), resumable, with `--dry-run` and a `--verify` completeness check. Same pattern as the runtime-code-prefixes backfill from #2918 and the `chain_id` backfill from #2111. The variant picked per compilation is deterministic (lowest `verified_contracts.id`, i.e. first submitter).
- **Server dual-writes**: every stored verification also inserts its compilation's metadata into the side table (first wins). `sourcify_matches.metadata` keeps being written and **remains the only read source** — no read path changes in this PR, so nothing depends on the backfill being complete.

## Rollout

1. Deploy this (migration + dual-write).
2. Run the backfill script; check completeness with `--verify`.
3. Ops: add `compiled_contracts_metadata` to the BigQuery Datastream mirror — the prod stream's include list is an explicit table allowlist, plus the `sourcify_publication` publication on the Postgres side, then an object backfill. **Must happen before step 5**: BigQuery's `public_sourcify_matches` currently mirrors the `metadata` column, so dropping it without mirroring the new table would silently remove metadata from BigQuery.
4. Follow-up PR: switch reads (`STORED_PROPERTIES_TO_SELECTORS`, `getCompilationsByIds`'s LATERAL detour) to the side table and stop writing `sourcify_matches.metadata`.
5. Separate ops task: drop the column and reclaim the ~139 GB with `pg_repack`.

Staging reads through the whole window means no `COALESCE` fallback ever ships.

## Notes

- The column deliberately stays `json`, not `jsonb`: `json` preserves the exact stored text, so the blob keeps hashing to the metadata hash embedded in the onchain bytecode (this property is asserted by the existing test in `assertContractSaved`). The `json` equality limitation doesn't bite here because grouping happens on the table's primary key.
- The dual-write lives in `SourcifyDatabaseService`, not `AbstractDatabaseService`, because the side table is Sourcify-specific and `AllianceDatabaseService` shares the abstract insert path (the Verifier Alliance DB has no such table).
- `assertContractSaved` now also checks the side table, so all existing integration verification tests cover the dual-write; dedicated tests cover first-submitter-wins and the no-metadata (Vyper) case.

## TODO

- [ ] After the migration is deployed, open `sourcify-prod-pg-bq-stream-sourcify-matches` datastream in the GCP console and tick `compiled_contracts_metadata`;

Closes #2929

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,6 +112,19 @@ npm run monitor:start
 
 **The authoritative schema is the committed dump at [`services/database/sourcify-database.sql`](services/database/sourcify-database.sql).** Read it to answer schema questions instead of querying a live database. dbmate regenerates it on `npm run migrate:up`, and it must be committed alongside any new migration (see [`services/database/README.md`](services/database/README.md)).
 
+**Never hand-edit `sourcify-database.sql` — always regenerate it with dbmate.** The `validate-database-schema` CI job applies all migrations to a clean postgres, dumps it, and diffs (comment/blank-line-normalized) against the committed file. pg_dump's object ordering is non-obvious (e.g. constraints sort by constraint name, not table name), so hand-edits get the ordering wrong and fail CI. To regenerate, run the migrations against a disposable local database:
+
+```bash
+# throwaway postgres (matches the version the dump was generated with)
+docker run -d --name schema-regen -e POSTGRES_PASSWORD=password -e POSTGRES_USER=postgres -e POSTGRES_DB=test_db -p 127.0.0.1:55432:5432 postgres:16
+
+cd services/database && DATABASE_URL="postgres://postgres:password@127.0.0.1:55432/test_db?sslmode=disable" DBMATE_SCHEMA_FILE=./sourcify-database.sql npm run migrate:up
+
+docker rm -f schema-regen
+```
+
+dbmate shells out to a local `pg_dump`, which must be version 16 (pgdg `postgresql-client-16`): pg_dump 17 emits `SET transaction_timeout` lines that CI's normalizer does not strip, so its dumps fail the diff.
+
 How the tables join:
 
 - `sourcify_matches.verified_contract_id` → `verified_contracts.id` (Sourcify-specific match info: `creation_match`/`runtime_match` quality, `chain_id`, and the contract's `metadata`)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,24 +112,14 @@ npm run monitor:start
 
 **The authoritative schema is the committed dump at [`services/database/sourcify-database.sql`](services/database/sourcify-database.sql).** Read it to answer schema questions instead of querying a live database. dbmate regenerates it on `npm run migrate:up`, and it must be committed alongside any new migration (see [`services/database/README.md`](services/database/README.md)).
 
-**Never hand-edit `sourcify-database.sql` — always regenerate it with dbmate.** The `validate-database-schema` CI job applies all migrations to a clean postgres, dumps it, and diffs (comment/blank-line-normalized) against the committed file. pg_dump's object ordering is non-obvious (e.g. constraints sort by constraint name, not table name), so hand-edits get the ordering wrong and fail CI. To regenerate, run the migrations against a disposable local database:
-
-```bash
-# throwaway postgres (matches the version the dump was generated with)
-docker run -d --name schema-regen -e POSTGRES_PASSWORD=password -e POSTGRES_USER=postgres -e POSTGRES_DB=test_db -p 127.0.0.1:55432:5432 postgres:16
-
-cd services/database && DATABASE_URL="postgres://postgres:password@127.0.0.1:55432/test_db?sslmode=disable" DBMATE_SCHEMA_FILE=./sourcify-database.sql npm run migrate:up
-
-docker rm -f schema-regen
-```
-
-dbmate shells out to a local `pg_dump`, which must be version 16 (pgdg `postgresql-client-16`): pg_dump 17 emits `SET transaction_timeout` lines that CI's normalizer does not strip, so its dumps fail the diff.
+**Never hand-edit `sourcify-database.sql` — always regenerate it with dbmate.** The `validate-database-schema` CI job applies all migrations to a clean postgres, dumps it, and diffs (comment/blank-line-normalized) against the committed file; pg_dump's object ordering is non-obvious (e.g. constraints sort by constraint name, not table name), so hand-edits get the ordering wrong and fail CI. The regeneration recipe (throwaway `postgres:15` container + `DBMATE_SCHEMA_FILE`) is in [`services/database/README.md`](services/database/README.md) under "Adding a new migration"; pg_dump must be version 15 to match CI (pgdg `postgresql-client-15`), not 17.
 
 How the tables join:
 
 - `sourcify_matches.verified_contract_id` → `verified_contracts.id` (Sourcify-specific match info: `creation_match`/`runtime_match` quality, `chain_id`, and the contract's `metadata`)
 - `verified_contracts.compilation_id` → `compiled_contracts.id`, and `verified_contracts.deployment_id` → `contract_deployments.id`
 - `compiled_contracts_sources` (`compilation_id`, `path`, `source_hash`) is the source set stored for a compilation; join `source_hash` → `sources.source_hash` for the actual content
+- `compiled_contracts_metadata.compilation_id` → `compiled_contracts.id`: one metadata blob per compilation (`sourcify_matches.metadata` is kept only until reads switch over, #2924)
 - `code` holds bytecode, referenced via `creation_code_hash`/`runtime_code_hash` by both `compiled_contracts` (compiled) and `contracts` (onchain); reach the latter through `contract_deployments.contract_id` → `contracts.id`
 
 ### Storage Services
--- a/services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
+++ b/services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
@@ -91,6 +91,14 @@ const firstMatchMetadataLateral = `
     LIMIT 1
   ) first_match`;
 
+// Makes re-runs skip already-filled compilations. Filters batch_metadata, not
+// batch, so the cursor still advances through completed ranges.
+const skipExistingRows = `
+  WHERE NOT EXISTS (
+    SELECT 1 FROM ${schema}.compiled_contracts_metadata ccm
+    WHERE ccm.compilation_id = batch.id
+  )`;
+
 program
   .description(
     "Backfill compiled_contracts_metadata from sourcify_matches.metadata.\n" +
@@ -134,7 +142,7 @@ program
     // the end of a normal run. Use it before switching reads over to the side
     // table.
     "--verify",
-    "Only compare the number of compilations expecting metadata against the side table, then exit.",
+    "Only count compilations still missing their side table row, then exit.",
     false,
   )
   .action(async (options) => {
@@ -148,20 +156,19 @@ program
 
     if (options.verify) {
       logger.info("Verifying compiled_contracts_metadata completeness");
+      // Counts the gap directly: comparing totals instead could hide gaps,
+      // because the side table also keeps rows for compilations that no
+      // current match points to anymore.
       const result = await activePool.query(
-        `SELECT
-           (SELECT count(DISTINCT vc.compilation_id)
-            FROM ${schema}.verified_contracts vc
-            JOIN ${schema}.sourcify_matches sm ON sm.verified_contract_id = vc.id
-            WHERE sm.metadata IS NOT NULL) AS expected_rows,
-           (SELECT count(*) FROM ${schema}.compiled_contracts_metadata) AS actual_rows`,
+        `SELECT count(DISTINCT vc.compilation_id) AS missing_rows
+         FROM ${schema}.verified_contracts vc
+         JOIN ${schema}.sourcify_matches sm ON sm.verified_contract_id = vc.id
+         WHERE sm.metadata IS NOT NULL
+           AND NOT EXISTS (SELECT 1 FROM ${schema}.compiled_contracts_metadata ccm
+                           WHERE ccm.compilation_id = vc.compilation_id)`,
       );
-      const expectedRows = parseInt(result.rows[0].expected_rows, 10);
-      const actualRows = parseInt(result.rows[0].actual_rows, 10);
       logger.info("Verification finished", {
-        expectedRows,
-        actualRows,
-        missingRows: expectedRows - actualRows,
+        missingRows: parseInt(result.rows[0].missing_rows, 10),
       });
       await closePool();
       return;
@@ -206,6 +213,7 @@ program
              SELECT batch.id AS compilation_id
              FROM batch
              ${firstMatchMetadataLateral}
+             ${skipExistingRows}
            )
            SELECT (SELECT count(*) FROM batch) AS batch_rows,
                   (SELECT count(*) FROM batch_metadata) AS insertable_rows,
@@ -231,6 +239,7 @@ program
              SELECT batch.id AS compilation_id, first_match.metadata
              FROM batch
              ${firstMatchMetadataLateral}
+             ${skipExistingRows}
            ),
            inserted AS (
              INSERT INTO ${schema}.compiled_contracts_metadata (compilation_id, metadata)
@@ -307,4 +316,8 @@ program
     await closePool();
   });
 
-program.parse();
+program.parseAsync().catch(async (err) => {
+  logger.error("Fatal error", { error: err.message, stack: err.stack });
+  await closePool();
+  process.exit(1);
+});
--- a/services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
+++ b/services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
@@ -1,0 +1,310 @@
+/**
+ * Backfills compiled_contracts_metadata from sourcify_matches.metadata.
+ *
+ * Required after applying the migration that adds the metadata side table
+ * (20260826100000_add_compiled_contracts_metadata.sql): the server only writes
+ * metadata for verifications stored after the migration, so compilations
+ * verified before it are filled out-of-band by this script. Until it has
+ * finished, reads keep using sourcify_matches.metadata, so nothing is missing
+ * in the meantime.
+ *
+ * For each compilation the metadata of its earliest current sourcify_match
+ * (lowest verified_contracts.id) is stored -- the first submitter wins,
+ * consistent with how sources are deduplicated for a shared compilation.
+ * Compilations without any match, or whose matches all have NULL metadata,
+ * get no row.
+ *
+ * The script is idempotent and resumable: inserts use ON CONFLICT DO NOTHING
+ * and the table is walked in primary-key order with a keyset cursor
+ * (compiled_contracts.id is a uuid, so ranges are cursor-based rather than
+ * numeric). Safe to Ctrl+C and re-run. Rows written concurrently by the live
+ * server are left untouched.
+ *
+ * See: https://github.com/argotorg/sourcify/issues/2924
+ *
+ * Environment Variables Required:
+ *   - POSTGRES_HOST
+ *   - POSTGRES_PORT
+ *   - POSTGRES_DB
+ *   - POSTGRES_USER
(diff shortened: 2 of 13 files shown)


services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -588,6 +588,13 @@ export class SourcifyDatabaseService
         );
       }
 
+      // Temporary dual-write: reads stay on sourcify_matches.metadata until
+      // the backfill completes, then the column gets dropped (issue #2924)
+      await this.database.insertCompiledContractMetadata(poolClient, {
```
kuzdogan: **`replace-metadata` never updates the side table**

The private `replace-metadata` method in `customReplaceMethods.ts` (the repair tool for contracts with wrong stored metadata, #2227) runs only:

```sql
UPDATE sourcify_matches SET metadata = $2 WHERE id = $1
```

It does not touch `compiled_contracts_metadata`, and this insert is `ON CONFLICT DO NOTHING`, so nothing can ever overwrite a side-table row. After the read switch and the column drop, every repair made in the meantime is silently lost.

Suggest that `replaceMetadata` also upserts (`DO UPDATE`) the side-table row for the match's `compilation_id` — here, or as a listed prerequisite of the read-switch PR.

kuzdogan: **Wasted work when the compilation already exists**

This runs for every stored verification. When the compilation was already in `compiled_contracts` (byte-identical contract verified before), the full metadata JSON is sent to Postgres, parsed, and then discarded by `DO NOTHING`.

The base class already computes `isNewCompilation` (it uses it to skip re-inserting sources). Returning that flag from `insertNewVerifiedContract` / `updateExistingVerifiedContract` and only inserting when it is true avoids that. The backfill covers pre-migration compilations anyway.

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
```diff
@@ -0,0 +1,310 @@
...
+    });
+
+    if (options.verify) {
+      logger.info("Verifying compiled_contracts_metadata completeness");
+      const result = await activePool.query(
+        `SELECT
+           (SELECT count(DISTINCT vc.compilation_id)
+            FROM ${schema}.verified_contracts vc
+            JOIN ${schema}.sourcify_matches sm ON sm.verified_contract_id = vc.id
+            WHERE sm.metadata IS NOT NULL) AS expected_rows,
```
kuzdogan: **`--verify` can under-report missing rows**

`expected_rows` counts distinct compilations with a *current* match that has metadata. `actual_rows` is `count(*)` of the side table. But the side table also holds rows for compilations that no current match points to any more:

1. A contract is verified with compilation X after deploy; the dual-write inserts a row for X.
2. A better match with a different compilation Y replaces it. The old `verified_contracts` row for X stays, but no `sourcify_matches` row points to it.
3. `expected_rows` counts Y only; `actual_rows` counts X and Y.

So surplus rows offset real gaps, and `missingRows` can even go negative. Counting the gap directly is exact and should be 0:

```sql
SELECT count(DISTINCT vc.compilation_id)
FROM verified_contracts vc
JOIN sourcify_matches sm ON sm.verified_contract_id = vc.id
WHERE sm.metadata IS NOT NULL
  AND NOT EXISTS (SELECT 1 FROM compiled_contracts_metadata m WHERE m.compilation_id = vc.compilation_id)
```

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
```diff
@@ -0,0 +1,310 @@
+/**
+ * Backfills compiled_contracts_metadata from sourcify_matches.metadata.
```
kuzdogan, line 2: **Missing row in the README's "Available upgrade scripts" table**

`services/database/README.md` tells operators to check that table when they upgrade a live database. This script has no row there. Suggested:

| Script | When to run | What it does |
|---|---|---|
| [`backfill-compiled-contracts-metadata.mjs`](./schema-updates/backfill-compiled-contracts-metadata.mjs) | After applying the migration that adds `compiled_contracts_metadata`, on a server that already dual-writes to it. Run `--verify` before switching reads to the table. | Backfills `compiled_contracts_metadata` from `sourcify_matches.metadata`, one row per compilation (first current match by `verified_contracts.id`). |

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
```diff
@@ -0,0 +1,310 @@
...
+          "ANALYZE failed (needs table owner); run it manually before switching reads to the new table",
+          { error: err.message },
+        );
+      }
+    }
+
+    await closePool();
+  });
+
+program.parse();
```
kuzdogan: **Errors are not caught**

`program.parse()` does not wait for the async action, so a failed query (a bad `--start-id` failing the `::uuid` cast, a dropped connection, a permission error) is an unhandled rejection: raw stack trace, the logger is bypassed, and `closePool()` is skipped.

`post-v2.12-to-v2.13-upgrade.mjs` does it right:

```js
program.parseAsync().catch(async (err) => {
  logger.error(...);
  await closePool();
  process.exit(1);
});
```

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
```diff
@@ -0,0 +1,310 @@
...
+      } else {
+        const result = await activePool.query(
+          `WITH batch AS (
+             SELECT compiled_contracts.id
+             FROM ${schema}.compiled_contracts
+             WHERE compiled_contracts.id > $1::uuid
+             ORDER BY compiled_contracts.id
+             LIMIT $2
+           ),
+           batch_metadata AS (
```
kuzdogan, line 238: **Re-runs redo all the expensive work**

Each batch looks up the matches and reads the metadata blob for every compilation in the batch, and only then `ON CONFLICT` discards the rows that already exist. So a re-run after Ctrl+C costs about as much as the first run.

Adding a `NOT EXISTS` against the side table *here* in `batch_metadata` (not in `batch`, so `last_id` still advances through every id and the loop still ends) makes re-runs skip done compilations. It also makes the `--dry-run` `insertable_rows` accurate on resume.

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    logger.info("Starting backfill of compiled_contracts_metadata", {
+      batchSize: options.batchSize,
+      sleepMs: options.sleepMs,
+      startId: options.startId,
+      dryRun: options.dryRun,
+    });
+
+    const estimate = await activePool.query(
```
kuzdogan, line 184: Nit: the ETA subtracts only this run's `totalProcessed` from the whole-table estimate, so a run resumed with `--start-id` overstates the remaining time. Cosmetic (same in the sibling script).

Also cosmetic: ~150 lines of scaffold (pool, signal handlers, option parsing, batch loop, ANALYZE) are now a third copy of the sibling scripts, and the dry-run and real queries repeat the same first half. Fine to leave if the "each script is self-contained" style is intended.

CLAUDE.md
```diff
@@ -112,6 +112,19 @@ npm run monitor:start
 
 **The authoritative schema is the committed dump at [`services/database/sourcify-database.sql`](services/database/sourcify-database.sql).** Read it to answer schema questions instead of querying a live database. dbmate regenerates it on `npm run migrate:up`, and it must be committed alongside any new migration (see [`services/database/README.md`](services/database/README.md)).
 
+**Never hand-edit `sourcify-database.sql` — always regenerate it with dbmate.** The `validate-database-schema` CI job applies all migrations to a clean postgres, dumps it, and diffs (comment/blank-line-normalized) against the committed file. pg_dump's object ordering is non-obvious (e.g. constraints sort by constraint name, not table name), so hand-edits get the ordering wrong and fail CI. To regenerate, run the migrations against a disposable local database:
+
+```bash
+# throwaway postgres (matches the version the dump was generated with)
+docker run -d --name schema-regen -e POSTGRES_PASSWORD=password -e POSTGRES_USER=postgres -e POSTGRES_DB=test_db -p 127.0.0.1:55432:5432 postgres:16
```
kuzdogan: **Should be PostgreSQL 15, not 16**

CI's `validate-database-schema` job runs `postgres:15-alpine` with `postgresql-client-15`, and the README says 15.13. The dump's version header is a comment line that CI strips, so 15 passes; there is no reason to require 16 here.

Suggest `postgres:15` in the `docker run` and `postgresql-client-15` in the sentence below, wording it as "pg_dump 15, to match CI". Keep the "not 17" warning (`SET transaction_timeout`), that part is correct.

Optional: this recipe could live in `services/database/README.md` under "Adding new migrations", which already describes regeneration and which `CLAUDE.md` links to above.

CLAUDE.md
```diff
@@ -112,6 +112,19 @@ npm run monitor:start
...
+cd services/database && DATABASE_URL="postgres://postgres:password@127.0.0.1:55432/test_db?sslmode=disable" DBMATE_SCHEMA_FILE=./sourcify-database.sql npm run migrate:up
+
+docker rm -f schema-regen
+```
+
+dbmate shells out to a local `pg_dump`, which must be version 16 (pgdg `postgresql-client-16`): pg_dump 17 emits `SET transaction_timeout` lines that CI's normalizer does not strip, so its dumps fail the diff.
+
 How the tables join:
 
 - `sourcify_matches.verified_contract_id` → `verified_contracts.id` (Sourcify-specific match info: `creation_match`/`runtime_match` quality, `chain_id`, and the contract's `metadata`)
```
kuzdogan, line 119: Nit: the "How the tables join" list should mention the new table, e.g. `compiled_contracts_metadata.compilation_id → compiled_contracts.id` (one metadata blob per compilation; `sourcify_matches.metadata` is kept only until the read switch).

services/server/src/server/services/utils/database-util.ts
```diff
@@ -157,6 +157,11 @@ export namespace Tables {
     chain_id: string;
   }
 
+  export interface CompiledContractMetadata {
+    compilation_id: string;
+    metadata: Metadata;
```
kuzdogan, line 162: Nit: `metadata` is declared required, but `insertCompiledContractMetadata` guards `if (!metadata) return` (Vyper has none) and the call site uses `as any` to get `undefined` past the type checker. `metadata?: Metadata` makes the type match reality and removes the cast. (The same `as any` exists on the `sourcify_matches` insert, so this is inherited.)

services/server/src/server/services/utils/Database.ts
```diff
@@ -796,6 +796,28 @@ ${
...
+
+    // ON CONFLICT DO NOTHING keeps the first submitter's metadata when a
+    // compilation is shared by several verified contracts, consistent with how
+    // compiled_contracts_sources are deduplicated
+    await poolClient.query(
+      `INSERT INTO ${this.schema}.compiled_contracts_metadata (
+        compilation_id,
+        metadata
+      ) VALUES ($1, $2)
+      ON CONFLICT (compilation_id) DO NOTHING`,
```
kuzdogan: Nit: every other insert in this file uses `ON CONFLICT ON CONSTRAINT <name> DO NOTHING`; this is the only `ON CONFLICT (column)` form. Same behavior, just inconsistent — `ON CONFLICT ON CONSTRAINT compiled_contracts_metadata_pkey DO NOTHING`.

services/server/test/helpers/assertions.ts
```diff
@@ -217,6 +217,28 @@ export async function assertContractSaved(
...
+
+    // The metadata must also be dual-written once per compilation into
+    // compiled_contracts_metadata (see issue #2924)
+    const compilationMetadataRes = await sourcifyDatabase.query(
+      `SELECT
+          ccm.metadata
+        FROM sourcify_matches sm
+        JOIN verified_contracts vc ON vc.id = sm.verified_contract_id
+        JOIN contract_deployments cd ON cd.id = vc.deployment_id
+        JOIN compiled_contracts_metadata ccm ON ccm.compilation_id = vc.compilation_id
```
kuzdogan: Nit: this repeats the three-table join of the query right above it. A `LEFT JOIN compiled_contracts_metadata ccm ON ccm.compilation_id = vc.compilation_id` on that query, selecting `ccm.metadata AS compilation_metadata`, does the same job with one query (assert non-null instead of `length(1)`).

kuzdogan (changes requested):
The core change looks right: the dual-write sends the same object to `sourcify_matches.metadata` and the side table, the migration only creates a new table, the schema dump is a real dbmate regeneration, and the backfill query uses the existing indexes. Requesting changes for the items below; the rest are nits.

**Should be addressed**

- `replace-metadata` (`customReplaceMethods.ts`) only updates `sourcify_matches.metadata`; nothing can ever overwrite a `compiled_contracts_metadata` row, so repairs made before the read switch are lost. Needs a `DO UPDATE` there, or a listed prerequisite of the read-switch PR.
- `--verify` compares a count of current matches against `count(*)` of the side table, which also holds rows for compilations no current match points to any more, so it can hide gaps. A `NOT EXISTS` gap count is exact (query in the inline comment).
- `services/database/README.md` "Available upgrade scripts" needs a row for the new script (suggested text inline).
- `CLAUDE.md` should say PostgreSQL 15 (what CI uses), not 16.

**Nits** (inline): uncaught errors in the script (`program.parseAsync().catch`), slow re-runs (`NOT EXISTS` in `batch_metadata`), ETA on resume, metadata insert when the compilation already exists, `metadata?: Metadata` instead of `as any`, `ON CONFLICT` style, redundant query in `assertContractSaved`, new table in the `CLAUDE.md` join list.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
marcocastignoli: Done in f7123a67: `replaceMetadata` now also upserts the side table row (`ON CONFLICT ... DO UPDATE`) for the match's compilation, so repairs made before the read switch survive the column drop. Like the `runtime_code_artifacts` repair below it, this affects every contract sharing the compilation.

marcocastignoli: Done in f7123a67: `insertNewVerifiedContract` / `updateExistingVerifiedContract` now return `isNewCompilation` and the dual-write only runs for new compilations.

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
marcocastignoli: Done in f7123a67: `--verify` now runs exactly this `NOT EXISTS` gap query and reports only `missingRows`. Tested against a local database: 0 when complete, the exact count after deleting side table rows.

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
marcocastignoli: Added in f7123a67 with your suggested text.

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
marcocastignoli: Done in f7123a67, same pattern as `post-v2.12-to-v2.13-upgrade.mjs`. A bad `--start-id` now logs through the logger, closes the pool and exits 1.

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
marcocastignoli: Done in f7123a67: the `NOT EXISTS` sits in `batch_metadata` so the cursor still advances through completed ranges. A re-run inserts 0 and resumed `--dry-run` counts are accurate (tested locally).

services/database/schema-updates/backfill-compiled-contracts-metadata.mjs
marcocastignoli: Leaving both as is: the ETA behavior matches the sibling script, and the self-contained style for the schema-update scripts is intentional.

CLAUDE.md
marcocastignoli: Done in f7123a67: the recipe moved to `services/database/README.md` under "Adding a new migration" with `postgres:15` / pg_dump 15, and CLAUDE.md keeps only the never-hand-edit warning plus a pointer. Also regenerated the dump with pg_dump 15.19; only the two version header comments changed, which doubles as a check that the recipe and the committed content are correct.

CLAUDE.md
marcocastignoli: Added in f7123a67.

services/server/src/server/services/utils/database-util.ts
marcocastignoli: Solved slightly differently in 83fd516e: `metadata` stays required so the row type keeps matching the NOT NULL column, and the Vyper check moved to the call site where the `Metadata | undefined` value actually lives. The `as any` is gone and `insertCompiledContractMetadata` lost its guard.

services/server/src/server/services/utils/Database.ts
marcocastignoli: Done in f7123a67.

services/server/test/helpers/assertions.ts
marcocastignoli: Done in f7123a67: one query with the `compiled_contracts_metadata` LEFT JOIN, asserting the joined metadata is non-null and hash-equal.

services/server/src/server/apiPrivate/customReplaceMethods.ts
```diff
@@ -173,12 +173,25 @@ export const replaceMetadata: CustomReplaceMethod = async (
...
+    `UPDATE sourcify_matches
+       SET
          metadata = $2
        WHERE id = $1`,
     [matchId, verification.compilation.metadata],
   );
+
+  // Also repair the side table; DO UPDATE because the dual-write inserts are
+  // DO NOTHING (#2924). Affects every contract sharing the compilation.
+  await sourcifyDatabaseService.database.pool.query(
```
kuzdogan, line 185: Nit: the `sourcify_matches` UPDATE above and this side-table upsert run as two separate statements on the pool, not in one transaction. If this second query fails, the two tables diverge until the repair is re-run. Since this is an admin tool that can simply be re-run, fine as is — but a `BEGIN`/`COMMIT` around both would make the repair atomic.

services/server/src/server/apiPrivate/customReplaceMethods.ts
```diff
@@ -173,12 +173,25 @@ export const replaceMetadata: CustomReplaceMethod = async (
...
+    `INSERT INTO compiled_contracts_metadata (compilation_id, metadata)
+       SELECT vc.compilation_id, $2::json
+       FROM sourcify_matches sm
+       JOIN verified_contracts vc ON vc.id = sm.verified_contract_id
+       WHERE sm.id = $1
+     ON CONFLICT ON CONSTRAINT compiled_contracts_metadata_pkey
+       DO UPDATE SET metadata = EXCLUDED.metadata`,
```
kuzdogan, line 192: Nit: if `verification.compilation.metadata` is undefined (a Vyper compilation), this inserts/updates `NULL` into a `NOT NULL` column and fails mid-repair — after the `sourcify_matches` UPDATE already wrote `NULL`. Inherited shape (the old code also wrote `NULL` into `sourcify_matches.metadata`) and the tool is realistically Solidity-only, so no action needed; an early guard rejecting metadata-less verifications would just make the failure explicit.

kuzdogan (approved):
All review threads are addressed: the dual-write is now gated on `isNewCompilation`, `--verify` counts the gap directly, re-runs skip filled compilations, errors go through the logger, `replace-metadata` repairs the side table, and the docs/dump/README follow-ups are in (including the `.release-todos` file). Verified each fix against the branch; CI is green including `validate-database-schema`.

Two take-it-or-leave-it nits inline on the `replace-metadata` repair — neither blocks the merge.

PR #2942, opened 2026-08-26
feat: switch metadata reads to compiled_contracts_metadata

marcocastignoli:
> [!IMPORTANT]
> **Do not merge until all points are checked** — this PR removes the only fallback to `sourcify_matches.metadata`:
> - [x] Production backfill finished (`backfill-compiled-contracts-metadata.mjs`)
> - [x] `--verify` reports `missingRows=0`
> - [x] `compiled_contracts_metadata` added to the Datastream mirror (sourcify-matches stream)
> - [x] Parquet export updated (sourcifyeth/parquet-export#5)

## Summary

Release 2 of 2 for #2924, following #2941. Reads switch to the side table, writes to the old column stop, and the column is dropped:

- `STORED_PROPERTIES_TO_SELECTORS` (`metadata`, `std_json_output`) now read `compiled_contracts_metadata`; the table is LEFT JOINed only when a requested selector references it, and grouped by its PK where a GROUP BY exists (PK grouping is what makes the `json` column selectable).
- `getSourcifyMatchByChainAddress` drops its metadata selector: its only caller never read it.
- `getCompilationsByIds` (similarity search) joins the side table directly on the compilation PK — the LATERAL detour through `sourcify_matches` that #2924 called out is gone, along with its stale-compilation quirk.
- `insertSourcifyMatch` / `updateSourcifyMatch` no longer take or write `metadata`.
- The repair paths follow: `replace-metadata` only upserts the side table now, and the massive-replace `config-replace-metadata.js` reads the compilation metadata from the side table.
- New migration `20260915115359_drop_sourcify_matches_metadata.sql` drops `sourcify_matches.metadata` (instant). Its down step re-adds the nullable column, also instant, so the previous server version can be redeployed after a rollback and reads NULL metadata until rolled forward.
- Tests read metadata from the side table; `assertContractSaved`'s duplicate check collapses back into one query.
- Docs (CLAUDE.md, database README, backfill script header) describe the read switch instead of the dual-write phase.

## Release order

The `.release-todos` file carries this. The pending migrations (#2963 `created_at` and the drop) must run **after** the deploy, not before: the previous server version still reads and writes `sourcify_matches.metadata`, and this version needs neither migration to run. The `2963-metadata-created-at.md` todo is folded into `2942-metadata-read-switch.md` for that reason. Then `pg_repack` on `sourcify_matches` reclaims the ~139 GB.

## Notes

- Rebased on staging after #2941 and #2963 (`created_at` on the side table) merged.
- Local runs against the docker DB: storageServices, apiPrivate, verification-cases, apiv2 → 167 passing. The one failure (`should fetch a missing source file via IPFS`) fails identically on `staging` from this machine: the test server uses the real ipfs.io gateway instead of the `.env.test` mock host, and ipfs.io rate-limits this IP.


.release-todos/2942-metadata-read-switch.md
```diff
@@ -0,0 +1,6 @@
+# Metadata read switch (#2942)
+
+## before
+
+- This version reads metadata only from `compiled_contracts_metadata`: confirm the backfill has completed on production by running `backfill-compiled-contracts-metadata.mjs --verify` and checking it reports `missingRows: 0`
+- Deploy the parquet-export changes (export `compiled_contracts_metadata`, drop the `metadata` column from the `sourcify_matches` export) BEFORE this version: once this server stops writing `sourcify_matches.metadata`, the old export writes null metadata for new and re-verified matches
```
manuelwedler: The parquet export and the Datastream mirror are both gated, but the Dune export in this repo is a third consumer of `sourcify_matches.metadata` and is not covered anywhere.

- `services/database/dune/fetchRows.ts:166` pulls `SELECT * FROM sourcify_matches`
- `services/database/dune/formatRows.ts:26` maps the row through `JSON.stringify(row.metadata)`
- `services/database/dune/DuneTableClient.ts:385` declares the Dune `metadata` column as `nullable: false`
- `compiled_contracts_metadata` is not in the Dune table set at all

Once this version stops writing the column, every newly synced match lands in Dune with the literal string `"null"` for its metadata, and the real value has no path into Dune. Same failure mode as the parquet export, so it probably wants the same treatment: a line in this file, or the side table added to the export.

services/server/src/server/services/utils/Database.ts
```diff
@@ -168,7 +168,7 @@ export class Database {
           sourcify_matches.created_at,
           sourcify_matches.creation_match,
           sourcify_matches.runtime_match,
-          sourcify_matches.metadata,
+          compiled_contracts_metadata.metadata,
```
manuelwedler, line 171: `getSourcifyMatchByChainAddress` has one caller, `checkByChainAndAddressAndMatch` (`SourcifyDatabaseService.ts:82`), and it reads only `runtime_match`, `creation_match`, `created_at`, `onchain_runtime_code` and `name`.

The metadata selector was already dead before this PR, when it read `sourcify_matches.metadata`. Moving it onto a new LEFT JOIN keeps the dead read and adds a join to the already-verified check that runs on every verification request. The column is `json`, so each lookup detoasts and transfers the blob only to discard it.

Two knock-on points if it stays:

- The join is unconditional here, unlike the `metadataRequested` gating you added to `getSourcifyMatchByChainAddressWithProperties`.
- `GetSourcifyMatchByChainAddressResult` gains `metadata: Metadata` as neither optional nor nullable, but a LEFT JOIN yields null for Vyper compilations and for any compilation the backfill missed. `Nullable<Metadata>` is the convention elsewhere in that file.

Dropping the selector, the join and the type member looks like the cheaper option.

services/server/src/server/services/utils/Database.ts
```diff
@@ -209,6 +210,9 @@ ${
       (property) => STORED_PROPERTIES_TO_SELECTORS[property],
     );
 
+    const metadataRequested =
```
manuelwedler, line 213: `metadataRequested` restates which selectors reference `compiled_contracts_metadata`. That knowledge already lives in `STORED_PROPERTIES_TO_SELECTORS` (`database-util.ts:467` and `:489`). A later selector that reads the side table but is not added to this list produces a query that references the table without the join. The request then fails at runtime with `missing FROM-clause entry`, and TypeScript cannot catch it.

Deriving it keeps one source of truth:

```ts
const metadataRequested = selectors.some((s) =>
  s.includes("compiled_contracts_metadata."),
);
```

Separately, the GROUP BY entry below is worth a word in the code. It works only because `compilation_id` is the primary key of `compiled_contracts_metadata`, and that functional dependency is what makes the `json` column selectable. The comment removed from `getCompilationsByIds` used to carry that reasoning, and it is not obvious from here.

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
```diff
@@ -593,8 +591,7 @@ export class SourcifyDatabaseService
         );
       }
 
-      // Temporary dual-write: reads stay on sourcify_matches.metadata until
-      // the backfill completes, then the column gets dropped (issue #2924).
+      // Metadata is stored once per compilation, first submitter wins (#2924).
```
manuelwedler, line 594: This comment is updated, but the sibling notes that describe the same dual-write phase are left behind, and they become wrong when this merges:

- `CLAUDE.md:119` still lists metadata among the things `sourcify_matches` holds, and `:122` says the column `is kept only until reads switch over, #2924`. This PR is that switch.
- `services/database/README.md:132` describes the backfill as running `on a server that already dual-writes to it`, and tells the operator to run `--verify` `before switching reads to the table`.
- `services/database/schema-updates/backfill-compiled-contracts-metadata.mjs:8` says `reads keep using sourcify_matches.metadata, so nothing is missing in the meantime`. The migration at `20260826100000_add_compiled_contracts_metadata.sql:11` says the same.

The README and the script header matter most. The release script deletes the `.release-todos` file that carries the ordering, so those two are the only operator-facing record that survives.

.release-todos/2942-metadata-read-switch.md
```diff
@@ -0,0 +1,6 @@
+# Metadata read switch (#2942)
+
+## before
+
+- This version reads metadata only from `compiled_contracts_metadata`: confirm the backfill has completed on production by running `backfill-compiled-contracts-metadata.mjs --verify` and checking it reports `missingRows: 0`
```
manuelwedler, line 5: Worth writing down here so it is not lost after the release: dropping `sourcify_matches.metadata` still needs its own dbmate migration. The `pg_repack` that reclaims the ~139 GB is the ops half, but the `ALTER TABLE ... DROP COLUMN` is a normal migration, and per CLAUDE.md it has to ship with a regenerated `sourcify-database.sql` and its own `.release-todos` file.

It should not ride along with this release, for two reasons:

- The column is the rollback net for exactly this change. The previous server version reads and writes it, so a rollback after the column is gone fails on every insert and every read.
- It cannot go before the parquet export and the Datastream mirror stop selecting the column. A NULL column degrades them quietly, a missing column can break them outright, so the drop needs a firmer gate than the one in this file.

So the follow-up is a later PR with a migration plus the repack, and #2924 should stay open until both are done.

.release-todos/2942-metadata-read-switch.md
manuelwedler: The parquet export and the Datastream mirror are both gated, but the Dune export in this repo is a third consumer of `sourcify_matches.metadata` and is not covered anywhere.

- `services/database/dune/fetchRows.ts:166` pulls `SELECT * FROM sourcify_matches`
- `services/database/dune/formatRows.ts:26` maps the row through `JSON.stringify(row.metadata)`
- `services/database/dune/DuneTableClient.ts:385` declares the Dune `metadata` column as `nullable: false`
- `compiled_contracts_metadata` is not in the Dune table set at all

Once this version stops writing the column, every newly synced match lands in Dune with the literal string `"null"` as its metadata, and the real value has no path into Dune. Same failure mode as the parquet export, so it probably wants the same treatment: a line in this file, or the side table added to the export.

services/server/src/server/services/utils/database-util.ts
```diff
@@ -251,6 +250,7 @@ export type GetSourcifyMatchByChainAddressResult = Tables.SourcifyMatch &
   Pick<Tables.CompiledContract, "runtime_code_artifacts" | "name" | "version"> &
   Pick<Tables.ContractDeployment, "transaction_hash"> & {
     onchain_runtime_code: string;
+    metadata: Metadata;
```
manuelwedler, line 253: Every other member of this type is derived from its table type. `metadata` used to arrive for free through `Tables.SourcifyMatch`. Now that it is read from the side table it should come from that table's type instead of being re-declared loose:

```ts
  Pick<Tables.ContractDeployment, "transaction_hash"> &
  Pick<Tables.CompiledContractMetadata, "metadata"> & {
    onchain_runtime_code: string;
  };
```

`Tables.CompiledContractMetadata` already declares `metadata: Metadata` at line 159, so the `Pick` keeps the provenance visible and tracks the table type if it changes.

The same applies to the `metadata: Metadata` added to `GetSourcifyMatchByChainAddressWithPropertiesResult` at line 342, where the outer `Partial` keeps it optional.

Two notes:

- If you take the suggestion on `Database.ts:171` and drop the dead metadata read from `getSourcifyMatchByChainAddress`, this member disappears and only the `WithProperties` one needs the change.
- The `Pick` does not settle nullability. The column is `NOT NULL`, so the table type is correct, but reading it through a LEFT JOIN yields null for Vyper and for any compilation the backfill missed.

services/server/src/server/services/utils/database-util.ts
```diff
@@ -342,6 +339,7 @@ export type GetSourcifyMatchByChainAddressWithPropertiesResult = Partial<
       function_signatures: SignatureRepresentations[];
       event_signatures: SignatureRepresentations[];
       error_signatures: SignatureRepresentations[];
+      metadata: Metadata;
```
manuelwedler, line 342: Same point as on line 253, and this is the one that stays whichever way the other comment goes. Derive the member from the table it is read from instead of re-declaring it inside the inline object:

```ts
    Pick<Tables.CompiledContractMetadata, "metadata"> & {
      ...
      error_signatures: SignatureRepresentations[];
    }
```

The outer `Partial` keeps it optional, so no consumer changes. It matters here in particular because `GetCompilationsByIdsResult` picks `metadata` straight out of this type, so the provenance carries through to the similarity search path.

.release-todos/2942-metadata-read-switch.md
marcocastignoli: Would this work instead of a separate PR? Ship the drop migration in this PR (with the regenerated dump and its own todo) and move the `migrate:up` step to `after` for both pending migrations, this one and #2963. The server does not depend on `created_at`, the insert only writes `compilation_id` and `metadata`, so #2963 can run after the deploy too.

Release sequence: deploy, then `migrate:up` on staging and production, then `pg_repack` on `sourcify_matches`. Between deploy and migration nothing breaks, this version never touches `sourcify_matches.metadata`. The todo would say plainly not to run `migrate:up` before the deploy, since the release script lists pending migrations at the start.

On your two points: the rollback net stays, the migration's down step re-adds the nullable column instantly, so 4.1.0 can be redeployed and reads NULL metadata until rolled forward. And the gate is already met: sourcifyeth/parquet-export#5 (merged 2026-09-09) drops the column from the export, and the Datastream mirror carries the side table.

*Posted with Claude Code*

manuelwedler: We can also include it here. sounds good to me.

services/server/src/server/services/utils/Database.ts
marcocastignoli: Done in 9d159913: selector, LEFT JOIN and the type member are gone, the query is back to what it was before this PR.

*Posted with Claude Code*

services/server/src/server/services/utils/Database.ts
marcocastignoli: Done in af6276dc: `metadataRequested` is now derived from the selectors as you suggested, and the GROUP BY entry has a one-line comment on the PK functional dependency. Small correction: the comment removed from `getCompilationsByIds` said the opposite (that `json` cannot be grouped, hence the scalar subquery), so this reasoning was not written down before.

*Posted with Claude Code*

services/server/src/server/services/storageServices/SourcifyDatabaseService.ts
marcocastignoli: Done in 210c0cf4: CLAUDE.md, the database README row and the backfill script header now describe the read switch and tell the operator to run `--verify` before deploying this version, since its release drops the source column. I left the migration file comment as is: it was true when that migration was applied and rewriting applied migrations seemed worse than a dated comment.

*Posted with Claude Code*

services/server/src/server/services/utils/database-util.ts
marcocastignoli: This member disappeared with the change on `Database.ts:171` (9d159913), only the `WithProperties` one remains, see the other thread.

*Posted with Claude Code*

services/server/src/server/services/utils/database-util.ts
marcocastignoli: Done in 168c7df7: `Pick<Tables.CompiledContractMetadata, "metadata">` in the Pick chain, the outer `Partial` keeps it optional. Nullability left as is: the service already sets `metadata` to null for non-Solidity contracts before returning.

*Posted with Claude Code*

.release-todos/2942-metadata-read-switch.md
marcocastignoli: Went ahead with this in 46ac280e: migration `20260915115359_drop_sourcify_matches_metadata.sql` with regenerated dump (up and down checked on a throwaway postgres 15), and the release todo now runs both pending migrations after the deploy, with the #2963 todo folded in. Happy to split it out again if you still prefer a separate PR.

*Posted with Claude Code*

manuelwedler (approved):
lgtm

Issue #887, opened 2023-01-05
Verify the contract address when verifying with contract creator tx hash


kuzdogan:
One of the ways we are working around verifying contracts is by "scraping" the contract's creator tx hash from the block explorer, fetching the tx with the hash, then checking if `tx.input === recompiled.creationBytecode`.

Here the "source of truth" becomes the block explorer. We assume the block explorer is giving us the correct tx hash for the address we are trying to verify. The block explorer gives us a tx hash, but we don't do another check if the given tx hash is the correct one. Theoretically, we can receive an incorrect tx.hash and end up verifying the contract at the address incorrectly. 

Particularly if we are going to modularize the verification into a library with a function like `verifyWithTxHash([files], chain, address, txHash)` we need to make sure the provided txHash is the correct one. **A verification must always be against a contract at an address and not against a tx.input**

Take the example

Contract A
----
- A.sol
- metadata-A.json
- address: `0xAA...AAA`
- creator address: `0x5A...CDF`
- tx contract created: `0x9A....1234`

Contract B
---
- B.sol
- metadata-B.json
- address: `0xBB...BBB`
- creator address: `0x5B...CDF`
- tx contract created: `0x8B...5678`

Now it becomes possible to incorrectly verify the address `0xAA...AAA` with the Contract B's source code:

`verifyWithTxHash( [B.sol, metadata-B.json], 1, 0xAA...AAA, 0x8B...5678)` will succeed because the provided tx.input will match the recompiled code. **There must be an additional step to check against the address being verified**

And the information for this check is already there. We know that the address of a contract is calculated by having the contract creator address + nonce. This information is already inside the tx receipt. 

In this case when `verifyWithTxHash( [B.sol, metadata-B.json], 1, 0xAA...AAA, 0x8B...5678)` is run:
- compare `tx.input === recompiled.creationBytecode` which will succeed, but also...
- get the `tx.from` which is `0x5B...CDF`
- get the `tx.nonce`,  say `0x02`
- calculate the contract address: `keccak(rlp.encode([0x5B...CDF, 0x02])).substring(24)` => `0xBB...BBB` `!==` `0xAA...AAA` which will fail to verify correctly.

PR #2946, opened 2026-08-27
fix: fall back to traces when a creation tx creates the verified contract internally

### LINKED ISSUE #2932: The creation bytecode cannot be retrieved via the `getContractCreationBytecodeAndReceipt` method of SourcifyChain.ts
EOA creates a token contract → the token’s constructor calls the Router/Factory → the Factory uses CREATE to deploy an LP pair contract. An exception may be thrown when retrieving the LP pair contract’s creation bytecode using the getContractCreationBytecodeAndReceipt method.

https://evm.confluxscan.org/tx/0x6f3a8fb1f14718ccdd2f35d2c966347d0f02ac912c336d4ace61bd0c430c1a38?tab=internal-txns

marcocastignoli:
## Summary

Fixes #2932.

A contract-creation transaction (`to === null`) can create more contracts than the one recorded in `receipt.contractAddress`: any internal `CREATE` during the constructor also deploys a contract under the same tx hash. In the issue's case (tx [`0x6f3a…1a38`](https://evm.confluxscan.org/tx/0x6f3a8fb1f14718ccdd2f35d2c966347d0f02ac912c336d4ace61bd0c430c1a38) on Conflux eSpace), an EOA deploys a token whose constructor calls a DEX factory, which deploys the LP pair — one tx, two contracts.

`getContractCreationBytecodeAndReceipt` treated a non-null `receipt.contractAddress` as "directly deployed, so the verified address must equal it" and threw on mismatch, never reaching the trace fallback. Now the mismatch case falls through to `getCreationBytecodeForFactory`, and the "EOA shortcut" is only taken when the receipt's address actually matches the verified contract.

The protection from #887 (rejecting a tx hash that doesn't create the verified contract) is preserved: the trace path only returns bytecode for a `CREATE` whose result address matches the verified address, and throws otherwise. When the chain has no trace support, an error is thrown naming both possible causes: a wrong tx hash, or an internal creation that can't be checked without traces.

## Notes

- Verified end-to-end against the issue's tx on Conflux eSpace: with a trace-enabled RPC config, the LP pair's creation bytecode is now fetched from `trace_transaction`; the token (directly-deployed) path is unchanged.
- Chain 1030's production config currently declares no `traceSupport`, so this contract also needs sourcifyeth/sourcify-chains#128 to be verifiable on sourcify.dev.
- New unit tests cover: direct deployment (case-insensitive address match), internal-creation fallback to traces, mismatch without trace support, and factory tx without trace support.

### DIFF
--- a/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
+++ b/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
@@ -864,7 +864,7 @@ export class SourcifyChain {
       if (!this.traceSupport) {
         if (txReceipt.contractAddress !== null) {
           throw new Error(
-            `Address of the contract being verified ${address} doesn't match the address ${txReceipt.contractAddress} created by this transaction ${transactionHash}, and chain ${this.chainId} has no trace support to look for internal creations`,
+            `Transaction ${transactionHash} directly created contract ${txReceipt.contractAddress}, not the contract being verified ${address}. Either the transaction hash is wrong or the contract was created by an internal transaction, which can't be checked because chain ${this.chainId} doesn't have trace support`,
           );
         }
         throw new Error(
--- a/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
+++ b/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
@@ -35,6 +35,9 @@ export function createFetchRequest(rpc: FetchRequestRPC): FetchRequest {
 
 export class RpcFailure extends Error {}
 
+/** A conclusive negative answer that no other RPC can change, e.g. the tx provably doesn't create the expected contract. Stops the RPC retry loop. */
+export class DefinitiveError extends Error {}
+
 export type SourcifyChainMap = {
   [chainId: string]: SourcifyChain;
 };
@@ -215,6 +218,9 @@ export class SourcifyChain {
           return result;
         }
       } catch (error) {
+        if (error instanceof DefinitiveError) {
+          throw error;
+        }
         if (error instanceof RpcFailure) {
           logWarn('RPC operation failed, marking as unhealthy', {
             operation: operationName,
@@ -415,7 +421,7 @@ export class SourcifyChain {
           const result = await parityStyleMethod(rpc, ...args);
           return { result };
         } catch (e: any) {
-          if (e instanceof RpcFailure) {
+          if (e instanceof RpcFailure || e instanceof DefinitiveError) {
             throw e;
           }
           logInfo('Failed to fetch from parity traces', {
@@ -438,7 +444,7 @@ export class SourcifyChain {
           const result = await gethStyleMethod(rpc, ...args);
           return { result };
         } catch (e: any) {
-          if (e instanceof RpcFailure) {
+          if (e instanceof RpcFailure || e instanceof DefinitiveError) {
             throw e;
           }
           logInfo('Failed to fetch from geth traces', {
@@ -454,6 +460,29 @@ export class SourcifyChain {
       return { tryNext: true };
     }, operationName);
   };
+
+  /**
+   * Returns a predicate telling whether a Parity style trace reverted, either itself
+   * or through an ancestor frame (a `traceAddress` prefix), whose revert also discards it.
+   */
+  private buildParityRevertedCheck(traces: any[]) {
+    const failedFrames = new Set(
+      traces
+        .filter((trace) => trace.error)
+        .map(
+          (trace) =>
+            `${trace.transactionHash ?? ''}:${(trace.traceAddress ?? []).join('.')}`,
+        ),
+    );
+    return (trace: any): boolean => {
+      const traceAddress: number[] = trace.traceAddress ?? [];
+      for (let i = 0; i <= traceAddress.length; i++) {
+        const frameKey = `${trace.transactionHash ?? ''}:${traceAddress.slice(0, i).join('.')}`;
+        if (failedFrames.has(frameKey)) return true;
+      }
+      return false;
+    };
+  }
 
   /**
    * For Parity style traces `trace_transaction`
@@ -484,16 +513,24 @@ export class SourcifyChain {
       );
     }
 
-    const createTraces = traces.filter((trace: any) => trace.type === 'create');
+    const isReverted = this.buildParityRevertedCheck(traces);
+    // Reverted creates never deployed anything; old OpenEthereum nodes also omit `result` on them
+    const createTraces = traces.filter(
+      (trace: any) => trace.type === 'create' && !isReverted(trace),
+    );
     // This line makes sure the tx in question is indeed for the contract being verified and not a random tx.
     const contractTrace = createTraces.find(
       (trace) =>
-        (trace.result.address as string).toLowerCase() ===
+        (trace.result?.address as string | undefined)?.toLowerCase() ===
         address.toLowerCase(),
     );
     if (!contractTrace) {
-      throw new Error(
-        `Provided tx ${creatorTxHash} does not create the expected contract ${address}. Created contracts by this tx: ${createTraces.map((t) => t.result.address).join(', ')}`,
+      // The trace is immutable, so no other RPC can answer differently
+      throw new DefinitiveError(
+        `Provided tx ${creatorTxHash} does not create the expected contract ${address}. Created contracts by this tx: ${createTraces
+          .map((t) => t.result?.address)
+          .filter(Boolean)
+          .join(', ')}`,
       );
     }
     logDebug('Found contract bytecode in traces', {
@@ -536,9 +573,11 @@ export class SourcifyChain {
       );
     }
 
+    const isReverted = this.buildParityRevertedCheck(traces);
     const createdAddresses: Record<string, string[]> = {};
     for (const trace of traces) {
       if (trace.type !== 'create') continue;
+      if (isReverted(trace)) continue;
       if (!trace.result || !trace.result.address || !trace.transactionHash) {
         logWarn('Found create trace with missing data, skipping.', {
           blockNumber,
@@ -582,7 +621,7 @@ export class SourcifyChain {
       rpc.maskedUrl,
     );
 
-    if (traces?.calls instanceof Array && traces.calls.length > 0) {
+    if (traces?.type) {
       logDebug('Fetched tx traces for creation tx hash', {
         creatorTxHash,
         maskedProviderUrl: rpc.maskedUrl,
@@ -594,9 +633,10 @@ export class SourcifyChain {
       );
     }
 
+    // The root frame is included: for a direct deployment it is itself the CREATE of the contract
     const createCalls: CallFrame[] = [];
     this.findCreateInDebugTraceTransactionCalls(
-      traces.calls as CallFrame[],
+      [traces as CallFrame],
       createCalls,
     );
 
@@ -608,11 +648,12 @@ export class SourcifyChain {
 
     // A call can have multiple contracts created. We need the one that matches the address we are verifying.
     const ourCreateCall = createCalls.find(
-      (createCall) => createCall.to.toLowerCase() === address.toLowerCase(),
+      (createCall) => createCall.to?.toLowerCase() === address.toLowerCase(),
     );
 
     if (!ourCreateCall) {
-      throw new Error(
+      // The trace is immutable, so no other RPC can answer differently
+      throw new DefinitiveError(
         `No CREATE or CREATE2 call found for the address ${address} in the traces of ${creatorTxHash} on RPC ${rpc.maskedUrl} and chain ${this.chainId}`,
       );
     }
@@ -695,9 +736,13 @@ export class SourcifyChain {
     createCalls: CallFrame[],
   ) {
     calls.forEach((call) => {
+      // A reverted frame deployed nothing, including in its subcalls
+      if (call?.error) return;
       if (call?.type === 'CREATE' || call?.type === 'CREATE2') {
         createCalls.push(call);
-      } else if (call?.calls?.length > 0) {
+      }
+      // A created contract's constructor can itself create contracts
+      if (call?.calls?.length > 0) {
         this.findCreateInDebugTraceTransactionCalls(call.calls, createCalls);
       }
     });
@@ -854,7 +899,7 @@ export class SourcifyChain {
     ) {
       // The tx deployed this contract directly, so its input data is the creation bytecode
       creationBytecode = creatorTx.data;
-      logDebug(`Contract ${address} created with an EOA`);
+      logDebug(`Contract ${address} created directly by the transaction`);
     } else {
       // The contract was created by an internal CREATE: either the tx called a factory
       // (contractAddress === null), or the tx deployed another contract whose constructor
@@ -871,7 +916,9 @@ export class SourcifyChain {
           `No trace support for chain ${this.chainId}. No other method to get the creation bytecode`,
         );
       }
-      logDebug(`Contract ${address} created with a factory. Fetching traces`);
+      logDebug(
+        `Contract ${address} not created directly by the transaction, fetching traces`,
+      );
       creationBytecode = await this.getCreationBytecodeForFactory(
         transactionHash,
         address,
--- a/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
+++ b/packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
@@ -845,22 +845,28 @@ export class SourcifyChain {
     if (!creatorTx) creatorTx = await this.getTx(transactionHash);
 
     let creationBytecode;
-    // Non null txreceipt.contractAddress means that the contract was created with an EOA
-    if (txReceipt.contractAddress !== null) {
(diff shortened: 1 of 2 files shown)


packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
```diff
@@ -845,22 +845,28 @@ export class SourcifyChain {
...
+      // The tx deployed this contract directly, so its input data is the creation bytecode
       creationBytecode = creatorTx.data;
       logDebug(`Contract ${address} created with an EOA`);
     } else {
-      // Else, contract was created with a factory
+      // The contract was created by an internal CREATE: either the tx called a factory
+      // (contractAddress === null), or the tx deployed another contract whose constructor
+      // created this one (contractAddress set but different, https://github.com/argotorg/sourcify/issues/2932).
+      // Both need traces, which also check that the tx actually created this contract
+      // and not a random one (https://github.com/argotorg/sourcify/issues/887).
```
kuzdogan, line 908: After this change, a receipt mismatch also goes to the trace helpers. So more kinds of traces reach them. One kind is a constructor that runs a `CREATE` that reverts, before the `CREATE` that makes the verified contract.

Both helpers read a field with no null guard:

```ts
// parity: extractCreationBytecodeFromParityTraceProvider
(trace.result.address as string).toLowerCase() === address.toLowerCase()
// and in its error message: createTraces.map((t) => t.result.address)
// geth: extractCreationBytecodeFromGethTraceProvider
(createCall) => createCall.to.toLowerCase() === address.toLowerCase()
```

Current mainnet nodes do not cause a crash here. I checked tx `0x95daecc900beafa524fd1b2119f1acac0e7de773e4164cf44bfd7bf876ab16af` (block 25482529). Its constructor runs a `CREATE2` that reverts. Geth `callTracer` returns `to: 0x000…000` and `error` for the failed frame. The Parity-format node behind dRPC keeps `result` (with `code: "0x"`) and sets `error: "Reverted"`.

But OpenEthereum-era nodes omit `result` on a failed create. The sibling function `extractCreatedAddressesFromParityTraceProvider` already guards `!trace.result || !trace.result.address`. So the code base treats that shape as possible.

If a node returns that shape, this happens:
1. `.find` throws `TypeError` on the failed frame, before it reaches the matching frame.
2. `getRpcDataViaTraceType` converts the error into `tryNext`.
3. Every RPC fails in the same way.
4. The caller only sees "All RPCs failed or are blocked". That looks like an outage.

A cheap fix is optional chaining in both helpers and in the error message: `trace.result?.address`, `t.result?.address`, `createCall.to?.toLowerCase()`. This is pre-existing, so a follow-up PR is also fine.

packages/lib-sourcify/test/SourcifyChain.spec.ts
```diff
@@ -37,6 +37,105 @@ describe('SourcifyChain', () => {
...
+      const { creationBytecode } =
+        await sourcifyChain.getContractCreationBytecodeAndReceipt(
+          '0xaddress',
+          '0xhash',
+        );
+      expect(creationBytecode).to.equal('0xcreationBytecode');
+    });
+
+    // https://github.com/argotorg/sourcify/issues/2932
+    it('should fall back to traces when the tx deploys another contract whose constructor creates the verified one', async () => {
```
kuzdogan: Suggested tests for the reverted-nested-`CREATE` case (see my comment in `SourcifyChain.ts`). On this branch they fail with `Cannot read properties of null (reading 'address')` and `Cannot read properties of undefined (reading 'toLowerCase')`; with optional chaining in the two helpers they pass.

```ts
it('should skip a reverted CREATE in parity traces and still find the verified contract', async () => {
  sandbox.stub(sourcifyChain, 'getTxReceipt').resolves({ contractAddress: '0xtokenAddress' } as any);
  sandbox.stub(sourcifyChain, 'getTx').resolves({ data: '0xtokenCreationBytecode' } as any);
  const mockProvider = sourcifyChain.rpcs[0].provider!;
  sandbox.stub(mockProvider, 'send').resolves([
    { type: 'create', result: { address: '0xtokenAddress' }, action: { init: '0xtokenCreationBytecode' } },
    // A CREATE that reverted: some parity-style nodes return result: null and an error
    { type: 'create', result: null, error: 'Reverted', action: { init: '0xfailedCreationBytecode' } },
    { type: 'create', result: { address: '0xpairAddress' }, action: { init: '0xpairCreationBytecode' } },
  ]);

  const { creationBytecode } = await sourcifyChain.getContractCreationBytecodeAndReceipt('0xpairAddress', '0xhash');
  expect(creationBytecode).to.equal('0xpairCreationBytecode');
});

it('should skip a reverted CREATE in geth traces and still find the verified contract', async () => {
  (sourcifyChain as any).rpcs = [
    { rpc: 'http://localhost:8545', traceSupport: 'debug_traceTransaction', provider: new JsonRpcProvider('http://localhost:8545') },
  ];
  sandbox.stub(sourcifyChain, 'getTxReceipt').resolves({ contractAddress: '0xtokenAddress' } as any);
  sandbox.stub(sourcifyChain, 'getTx').resolves({ data: '0xtokenCreationBytecode' } as any);
  const mockProvider = sourcifyChain.rpcs[0].provider!;
  sandbox.stub(mockProvider, 'send').resolves({
    type: 'CREATE', to: '0xtokenAddress', input: '0xtokenCreationBytecode',
    calls: [
      // A CREATE that reverted: a callTracer frame without `to`
      { type: 'CREATE', input: '0xfailedCreationBytecode', error: 'execution reverted' },
      { type: 'CREATE', to: '0xpairAddress', input: '0xpairCreationBytecode' },
    ],
  });

  const { creationBytecode } = await sourcifyChain.getContractCreationBytecodeAndReceipt('0xpairAddress', '0xhash');
  expect(creationBytecode).to.equal('0xpairCreationBytecode');
});
```

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
kuzdogan, line 908: Related: the trace helpers do not read the `error` field. So they treat a reverted create as a real one, if the node reports an address for it.

Live example: mainnet tx [`0x95daec…16af`](https://etherscan.io/tx/0x95daecc900beafa524fd1b2119f1acac0e7de773e4164cf44bfd7bf876ab16af) (block 25482529). It is a contract-creation tx. Its constructor runs a `CREATE2` that reverts.

Geth `debug_traceTransaction` with `callTracer` (QuickNode), the failed frame:

```json
{
  "type": "CREATE2",
  "from": "0x9d7f372fb27502d7a8…",
  "to": "0x0000000000000000000000000000000000000000",
  "input": "0x608060405260405161…",
  "error": "execution reverted",
  "gas": "0x8194",
  "gasUsed": "0x4e5d"
}
```

Parity-format `trace_transaction` (dRPC, a Reth/Erigon-style node), the same failed create:

```json
{
  "type": "create",
  "action": { "from": "0x9d7f372fb27502d7a8…", "creationMethod": "create2", "init": "0x608060405260405161…" },
  "result": { "address": "0x9b4bded299ea8a9da0…", "code": "0x", "gasUsed": "0x4e5d" },
  "error": "Reverted",
  "traceAddress": [0]
}
```

I checked all 5 frames in each trace. The 3 successful frames have no `error` key at all. The 2 failed frames have `error` set. So `if (trace.error)` is a correct "failed" check for these two node types.

What goes wrong:
1. `extractCreationBytecodeFromParityTraceProvider` matches on `result.address` only. If the verified address is the would-be address of a failed create, it returns the `action.init` of the failed attempt. That address gets code only if a later `CREATE2` with the same factory, salt, and init code succeeds. Then the bytecode is correct, but the stored creation tx hash and block point to a tx that did not create the contract.
2. `extractCreatedAddressesFromParityTraceProvider` (the block scan of the binary search) records the failed tx as the creator of that address. It guards `!trace.result`, but not `error`.
3. The Geth path is safe only by chance: `to` is the zero address, which never matches.

Fix: skip create frames that have `error`, in both Parity helpers and in `findCreateInDebugTraceTransactionCalls`. Two notes:
- A create under a parent frame that reverted has no `error` of its own (I saw this in block 25747543). A full check must also look at the parent frames.
- Old OpenEthereum nodes omit `result` on a failed create. Add `trace.result?.address` and `createCall.to?.toLowerCase()` as well, or `.find` throws `TypeError` and the lookup fails with "All RPCs failed".

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
```diff
@@ -845,22 +845,28 @@ export class SourcifyChain {
...
       if (!this.traceSupport) {
```
kuzdogan, line 909: Before this change, a wrong `creationTransactionHash` cost one receipt read. The function then threw "Address … doesn't match the address … created by this transaction". After this change, the mismatch goes to `getCreationBytecodeForFactory`. That has two effects.

**1. One full trace fetch per trace RPC.** The trace helper finds no create for the address and throws:

```ts
throw new Error(
  `Provided tx ${creatorTxHash} does not create the expected contract ${address}. Created contracts by this tx: ...`,
);
```

`getRpcDataViaTraceType` catches that and returns `{ tryNext: true }`. So `executeWithCircuitBreaker` asks the next RPC. The trace is immutable, so every RPC gives the same negative answer. The loop only stops after the last trace RPC. `creationTransactionHash` is unauthenticated input, so a client with many stale hashes multiplies the trace load.

**2. The #887 diagnostic is lost.** The precise message above only appears at `info` level, once per RPC. The error that reaches `Verification.verify` is the generic `All RPCs failed or are blocked for getCreationBytecodeForFactory(...)`. That is the same message as an RPC outage. An operator cannot tell "wrong hash" from "trace RPCs are down". On the Geth side, a plain deploy tx with no inner calls has no `calls` list, so the helper logs "received empty or malformed response" and blames the RPC for a correct answer.

Suggested fix: a non-retryable error, in the same way as `RpcFailure`. Add an error class, for example `DefinitiveError`. Throw it from the "no matching create" branch in both helpers (that branch runs only after the "empty or malformed" check, so `tryNext` stays correct for incomplete traces). Rethrow it in the two catch blocks of `getRpcDataViaTraceType`, and in `executeWithCircuitBreaker`:

```ts
} catch (error) {
  if (error instanceof DefinitiveError) throw error; // stop, the answer is final
  if (error instanceof RpcFailure) { ...; continue; }
```

Then a wrong hash costs one trace fetch, and the caller gets the precise message.

packages/lib-sourcify/test/SourcifyChain.spec.ts
kuzdogan, line 58: The new tests cover the success path of the trace fallback. One path has no test: the receipt address is different, `traceSupport` is on, and the trace has no create for the verified address. This is the path that keeps the #887 protection after this change. The expected result is a rejection.

The helper tests (`'no create trace is found in parity traces'`, `'contract address is not found in geth traces'`) call the helpers directly. Nothing asserts that `getContractCreationBytecodeAndReceipt` rejects in this case, or how many RPCs it calls. The wrong-hash test in `Verification.spec.ts` runs on the hardhat chain, which has no `traceSupport`, so it tests the no-trace branch.

Suggested test: a chain with two trace RPCs, `getTxReceipt` resolves `{ contractAddress: '0xtokenAddress' }`, `send` resolves a trace list without `'0xpairAddress'`. Assert that the call rejects. If the non-retryable error from my other comment is added, also assert that `send` is called once.

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
```diff
@@ -845,22 +845,28 @@ export class SourcifyChain {
...
+    // Compare case-insensitively: the receipt's contractAddress is checksummed
+    // on some RPCs, while the caller may pass a lowercase address, and a pure
+    // `!==` would spuriously reject a genuine match.
+    if (
+      txReceipt.contractAddress !== null &&
+      txReceipt.contractAddress.toLowerCase() === address.toLowerCase()
+    ) {
+      // The tx deployed this contract directly, so its input data is the creation bytecode
       creationBytecode = creatorTx.data;
       logDebug(`Contract ${address} created with an EOA`);
```
kuzdogan: Nit: the two debug messages do not match the branches any more. The first branch means "the tx created this contract directly"; the code does not check the sender, so "EOA" is an inference. The else branch also covers a contract that a constructor created (the new comment above says so), not only a factory. Suggested: `created directly by the transaction` and `not created directly by the transaction, fetching traces`. An operator who searches the logs for "factory" then does not read a constructor case as a factory deployment.

kuzdogan (commented):
The fix in `getContractCreationBytecodeAndReceipt` is correct, and the #887 protection holds through the trace helpers. Two related gaps stay open on the server side. Both are pre-existing and out of scope here, but issue #2932 is not fully closed by this PR alone.

**1. The server's own creation-tx search still rejects a constructor-created contract.** When the request has no hash and no explorer API gives a result, `getCreatorTx` falls back to `findContractCreationTxByBinarySearch` in `services/server/src/server/services/utils/contract-creation-util.ts`. That search finds the creation block, loops over the deploy transactions (`tx.to === null`), and accepts a tx only when the receipt names the address:

```ts
if (
  receipt.contractAddress?.toLowerCase() ===
  contractAddress.toLowerCase()
) {
```

For the issue's pair contract, the receipt names the token. So the search skips the outer tx and returns `null`. The pair gets a runtime-only match. This is the same assumption that this PR relaxes in `getContractCreationBytecodeAndReceipt`. On a chain with `traceSupport`, the search could use `getCreatedAddressesFromBlockTraces` (the monitor already uses it in `ChainMonitor.ts`), which also finds factory-created contracts.

**2. Similarity verification stores a hash that was just rejected.** In `resolveSimilarityCreationData` in `VerificationService.ts`, a hash arrives from the request body or from `getCreatorTx`. If `getContractCreationBytecodeAndReceipt` throws because the tx did not create this contract, the catch block returns the hash anyway:

```ts
} catch (error: any) {
  logger.debug("Failed to fetch creation data for similarity verification", { ... });
  return { creationTransactionHash: resolvedCreatorTxHash };
}
```

The mock chain in `_verifySimilarity` never throws, so `Verification` keeps the hash, and `database-util.ts` stores it as `contract_deployments.transaction_hash` with `deployer` and `block_number` empty. `Verification.ts` clears the hash in the same situation, so the two callers disagree. Returning `{}` from the catch block, in the same way as the no-hash case, would align them.

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
marcocastignoli: Fixed in 57148759: `trace.result?.address` in the parity helper (in the match and in the error message, which now also filters out empty entries) and `createCall.to?.toLowerCase()` in the geth helper. Done here rather than in a follow-up since the reverted-create fix touches the same lines anyway. Covered by the new "failed create without result" and "frames without a `to`" test cases.

marcocastignoli: Fixed in 57148759. Parity: both helpers now skip a create whose own frame or any ancestor frame (a `traceAddress` prefix) carries `error`; the block scan keys that check per `transactionHash` so identical `traceAddress`es across txs cannot collide. Geth: `findCreateInDebugTraceTransactionCalls` returns early on a frame with `error`, which also drops its whole subtree, covering the create-under-a-reverted-parent case you saw in block 25747543. Tests cover the reverted CREATE2 that still reports an address, the OpenEthereum shape without `result`, and the reverted-parent case on both trace types.

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
marcocastignoli: Added in 57148759 as suggested: `DefinitiveError` is thrown from the no-matching-create branch of both helpers, rethrown in the catch blocks of `getRpcDataViaTraceType`, and rethrown in `executeWithCircuitBreaker` before the retry logic. A wrong hash now costs one trace fetch and the precise #887 message reaches the caller. The "empty or malformed" branches stay `tryNext`. For the geth side effect you noted: the extractor now includes the trace root frame as a create candidate (the root of a creation tx is itself the CREATE), so a wrong hash pointing at a plain deploy gets the definitive mismatch error instead of blaming the RPC. One consequence worth reviewing: the shared walker now also recurses into CREATE frames, so the geth block scan additionally picks up creates made by a created contract's constructor, which it previously missed.

packages/lib-sourcify/test/SourcifyChain.spec.ts
marcocastignoli: Added in 57148759: a chain with two trace RPCs, `getTxReceipt` resolving a different `contractAddress`, and traces that contain no create for the verified address. It asserts the precise "does not create the expected contract" rejection and that the first RPC's `send` is called exactly once while the second is never asked.

packages/lib-sourcify/src/SourcifyChain/SourcifyChain.ts
marcocastignoli: Done in 57148759 with the suggested wording.

kuzdogan (approved):

PR #2959, opened 2026-09-02
fix: detect Livepeer ManagerProxy in proxyResolution

Strykar:
Livepeer's managers (BondingManager, RoundsManager, TicketBroker,
ServiceRegistry, BondingVotes) sit behind `ManagerProxy`, which keeps no
implementation pointer in a namespaced slot: slot 0 is the Controller, slot 1 the
target's registry key, and the fallback delegates to
`controller.getContract(targetContractId)`. `?fields=proxyResolution` currently
answers `isProxy: false` for all of them.

whatsabi added a resolver in shazow/whatsabi#216, released in v0.29.0. This is
the bump plus the `ProxyType` union, the `apiv2.yaml` enum, a bytecode fixture
and a test.

The bump moves more than #216: `ox` goes 0.14.34 to 1.7.4 and the noble/scure
closure under it moves with it. The commit message itemises exactly what shifts
and what stays pinned. Happy to split the bump into its own PR if you would
rather land it separately.

`npm run test:unit`, against this branch and against `staging` at 1a8ed449:
65 passing vs 64, same 3 failures in both (etherscan and veChain want API keys,
verificationWorker wants node >= 22 for Hardhat).

Verified on Arbitrum One: `0x35Bcf3c30594191d53231E4FF333E8A770453e40` resolves
to `0xbe197fcBfe74DE8F10460EA61644B006cC0f0Bd2`.


manuelwedler (approved):
looks good. Thanks!

Issue #2956, opened 2026-09-02
Update the parquet export to support the new metadata table


marcocastignoli:
With https://github.com/argotorg/sourcify/pull/2941 we added a new table, and with https://github.com/argotorg/sourcify/pull/2942 we are going to remove the existing column. We should update the parquet export script after mergin the latter PR.

manuelwedler:
done in https://github.com/sourcifyeth/parquet-export/pull/5 and reexported the full dataset

PR #2963, opened 2026-09-09
feat(database): add created_at to compiled_contracts_metadata

manuelwedler:
## Summary

Adds a `created_at` column to `compiled_contracts_metadata`, with the same insert and update triggers and the same index as on the other tables.

## Why

The Parquet Export v2 needs a `created_at` column on every exported table to select only the rows added since the last export. See #2956.

## Notes

- Existing rows get the migration time. Rows that the metadata backfill inserts later get the backfill time.
- The schema dump is regenerated with dbmate.


marcocastignoli (approved):
lgtm

PR #2945, opened 2026-08-26
Add release TODO checks to the release script

kuzdogan:
## Summary

Manual steps around a production deploy were easy to forget, because the release script only looked at package versions. Migrations are applied by hand, and the "Deploy latest to production" PR had an empty body. See the discussion on #2937 (drop `idx_code_code_first_75`).

This adds `scripts/release/release_todos.sh`, wired into `main.sh`:

- **`.release-todos/<pr-number>-<short-name>.md`** files with optional `## before` and `## after` sections. `before` items are shown at the start of the release, `after` items once the deploy PR is merged. The script deletes the files on the release branch, so the directory only holds open items.
- **New migration files** since the last deploy are listed automatically, with a reminder to apply them on staging and production.
- **`TODO_RELEASE` markers** in added lines are listed as a safety net.
- All collected items become the body of the "Deploy latest to production" PR, so its reviewer sees the same list.

Everything is scoped to `git diff master...staging`, so an item shows up for exactly one release.

## Notes

- `.release-todos/README.md` documents the format. `services/database/README.md` and `CLAUDE.md` now say that every migration PR must add such a file.
- Not included: a CI check that fails a migration PR without a `.release-todos/` file. Can follow in a separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2970, opened 2026-09-14
Fix release TODO script: lost after items and self-matching markers

### LINKED ISSUE #2955: Errors in release script
Following #2945 two bugs were introduced:

### todo_release folder is included itself.

This is an output of the release script

```
Release TODO (.release-todos/2941-compiled-contracts-metadata.md), before the deploy:

- Run `npm run migrate:up` on the staging and production databases (creates the empty `compiled_contracts_metadata` table, instant)
Done? (y/N): y

TODO_RELEASE markers in changes since the last production deploy:
.release-todos/README.md: - lists added lines that contain `TODO_RELEASE` since the last deploy
scripts/release/main.sh: prompt_execute_or_skip "checking for open release TODOs (migrations, .release-todos/, TODO_RELEASE markers)" check_release_todos
scripts/release/release_todos.sh: #   3. Added lines containing TODO_RELEASE anywhere in the diff (safety net).
scripts/release/release_todos.sh:     /^\+/ && /TODO_RELEASE/ { print file ": " substr($0, 2) }
scripts/release/release_todos.sh:   warn "TODO_RELEASE markers in changes since the last production deploy:"
scripts/release/release_todos.sh:   RELEASE_TODO_TEXT+=$'\n## TODO_RELEASE markers\n'"$hits"$'\n'
All of these done? (y/N): y
Success: checking for open release TODOs (migrations, .release-todos/, TODO_RELEASE markers)


Next: creating GitHub PR. Continue (default: Y), or skip (N)?
```

### After script cannot be displayed because files are deleted before displaying them

```
Next: cleaning up temporary release data. Continue (default: Y), or skip (N)?

Cleaning up temporary package data file: scripts/release/.release_package_data.tmp
Temporary file deleted successfully.
Success: cleaning up temporary release data


Next: cleaning up temporary release TODO data. Continue (default: Y), or skip (N)?

Success: cleaning up temporary release TODO data
```

### General question

Instead of deleting the todo notes, wouldn't it better to keep all the todo notes, and only output the todo notes that differs from master?

marcocastignoli:
## Summary

Fixes #2955: two bugs in the release TODO checks added in #2945.

- **Lost "after" items.** The notes in `.release-todos/` were read from the working tree, and the "after" items were cached in `scripts/release/.release_todos_after.tmp`. The release branch deletes the notes, so running the TODO check again after that step rebuilt the cache from an empty folder, and the "after" step printed "No release TODOs for after the deploy." The script now records the `staging` commit at the start of the release in `scripts/release/.release_todos_ref.tmp` and reads the notes from that commit, so neither the deletion on the release branch nor a pull of `staging` in between can lose them. The commit is recorded only while `staging` still has notes, so a re-run after the deletion keeps the original one. The cleanup step at the end of the release removes the file, and without it the script falls back to the `staging` branch.
- **The script listed its own lines as `TODO_RELEASE` markers.** The marker scan matched lines in `scripts/release/` and `.release-todos/README.md` that mention `TODO_RELEASE`. Both paths are now left out of the scan.

The notes are still deleted on the release branch in every release.

## Notes

- When the after step finds nothing, its message points at the "Deploy latest to production" PR body, which keeps the same list permanently.


scripts/release/release_todos.sh
```diff
@@ -97,26 +106,21 @@ check_release_todos() {
...
   done
 }
 
 # Runs at the end of the release, after the deploy PR is merged.
 show_release_todos_after() {
-  if [ ! -s "$RELEASE_TODOS_AFTER_FILE" ]; then
+  local after
+  after=$(release_todos_after)
+  if [ -z "$after" ]; then
     echo "No release TODOs for after the deploy."
```
kuzdogan: The "after" step only survives a re-run if the local `staging` branch is not pulled in between. But if the script exits after the release PR is merged and you run it again, the second step (`check_branch_sync`) refuses to continue until local `staging` equals `origin/staging`. That pull brings the deletion, and this branch prints "No release TODOs for after the deploy." I reproduced this in a throwaway clone.

So the fix holds within one uninterrupted run, and on a re-run only if the sync step is skipped. The PR description's "re-running a step no longer loses anything" is a bit stronger than that.

Two cheap options:
- Record `git rev-parse staging` at the check step (for example next to the package data in the temp file) and read from that commit instead of the branch name. A commit hash never moves.
- At minimum, make this message point at the deploy PR body, which already holds the "After deploy" section as a permanent copy.

scripts/release/release_todos.sh
```diff
@@ -97,26 +106,21 @@ check_release_todos() {
 # Deletes the release-todo files and stages the deletion for the release branch commit.
 clear_release_todo_files() {
   local file
-  for file in "$RELEASE_TODOS_DIR"/*.md; do
-    [ -f "$file" ] || continue
-    [ "$(basename "$file")" = "README.md" ] && continue
-    git rm -q "$file"
+  for file in $(release_todo_files); do
+    git rm -q --ignore-unmatch "$file"
```
kuzdogan: Nit: with `--ignore-unmatch`, `git rm` succeeds silently when the file is already gone, so the next line prints "Removed <file>" even when nothing was removed. If the flag is there for re-runs, maybe guard the echo with `[ -e "$file" ]` before the `git rm`.

.release-todos/README.md
```diff
@@ -37,6 +37,6 @@ Two optional sections. `before` is shown at the start of the release, before the
 The release script also, without any file here:
 
 - lists new files under `services/database/migrations/` since the last deploy
-- lists added lines that contain `TODO_RELEASE` since the last deploy
+- lists added lines that contain `TODO_RELEASE` since the last deploy (outside `scripts/release/` and this directory)
```
manuelwedler, line 42: nit: Line 10 above still says the script "reads every `.md` file in this directory". After this PR it reads the files from the local `staging` branch, and the working-tree copy is not used. The line is outside this diff, but it is the first thing a reader of this README sees. Could you update it, and maybe add the note from the PR description that the local `staging` branch must not be updated during the release?

scripts/release/release_todos.sh
```diff
@@ -97,26 +106,21 @@ check_release_todos() {
...
     echo "Removed $file"
```
manuelwedler, line 131: nit: With `--ignore-unmatch`, a re-run of this step prints `Removed <file>` for a file that was already deleted. Not harmful, but the output says something that did not happen. `git rm -q --ignore-unmatch "$file" && [ -e ... ]` is not right either since the file is gone in both cases; a simple `[ -e "$file" ] || continue` before the `git rm` keeps the message honest.

scripts/release/release_todos.sh
```diff
@@ -97,26 +106,21 @@ check_release_todos() {
...
   done
 }
 
 # Runs at the end of the release, after the deploy PR is merged.
 show_release_todos_after() {
-  if [ ! -s "$RELEASE_TODOS_AFTER_FILE" ]; then
+  local after
+  after=$(release_todos_after)
```
manuelwedler, line 138: optional: The "after" items are also in the body of the deploy PR (`RELEASE_TODO_TEXT` has a `## After deploy` section). Reading them from `gh pr view <deploy-pr> --json body` here would remove the dependency on the local `staging` branch not moving during the release. It needs the PR number after the PR is merged, so it is more work than this PR. Fine to leave as is, since the PR description documents the limitation.

kuzdogan (commented):
Posting late review :( comments left hanging "Pending"

scripts/release/release_todos.sh
marcocastignoli: You are right, and I reproduced it in a sandbox: once `check_branch_sync` asks for the pull, the after step printed "No release TODOs for after the deploy."

Fixed in 0f3b9a63 with your first option. The check step records `git rev-parse staging` in `scripts/release/.release_todos_ref.tmp`, and the notes are read from that commit instead of the branch name. One addition: the commit is recorded only while `staging` still has notes, so the re-run that happens after the deletion cannot overwrite it with the new `staging`. The cleanup step at the end of the release deletes the file, and if it is missing the script falls back to the `staging` branch.

Verified in the sandbox: the recorded commit survives a re-run of the check step after `staging` moved, and the after items are still listed. The empty message now points at the deploy PR body as you suggested, and the PR description no longer claims more than the code does.

*Posted with Claude Code*

scripts/release/release_todos.sh
marcocastignoli: Fixed in 46b1c174: there is a `[ -f "$file" ] || continue` before the `git rm` now, and `--ignore-unmatch` is gone. A second run of the step prints nothing.

*Posted with Claude Code*

.release-todos/README.md
marcocastignoli: Updated. The first paragraph now says the files are read as committed on the `staging` branch, and that the script records that commit at the start of the release.

I first added the warning you suggested about not pulling `staging` during a release, but 0f3b9a63 removes that limitation, so the paragraph explains the recorded commit instead.

*Posted with Claude Code*

scripts/release/release_todos.sh
marcocastignoli: Fixed in 46b1c174: guarded with `[ -f "$file" ] || continue` before the `git rm`, and `--ignore-unmatch` is gone, so a re-run of the step reports nothing.

*Posted with Claude Code*

scripts/release/release_todos.sh
marcocastignoli: Went with a recorded commit instead (Kaan suggested the same in his review), so the after step no longer depends on the local `staging` branch at all.

The deploy PR body is still used as the permanent copy: when the after step finds nothing, its message now points there.

*Posted with Claude Code*

manuelwedler (approved):
lgtm

PR #2975, opened 2026-09-16
fix(lib-sourcify): fix infinite loop in ipfsHash for empty input

manuelwedler:
## Summary

- `ipfsHash("")` never returned. `ipfsHashData` computed a chunk count of `0` for empty input, so `groupChunksBottomUp` received an empty level. Its loop runs until the level has exactly one element, and `buildNextLevel([])` returns `[]` again, so the loop never ended and pinned the thread at 100% CPU.
- Fix: empty input now produces one empty chunk. The result is `QmbFMke1KXqnYyBBWxB74N4c5SBnJMVAiMNRcGu6x1AwQH`, the CIDv0 of an empty file. It matches the test vector in the Solidity compiler's `IpfsHash` tests.
- Guard: `groupChunksBottomUp` now throws on an empty level instead of looping. This cannot trigger after the fix, but it turns a future regression into an exception instead of a hang.
- `swarmBzzr1Hash("")` returned the Swarm BMT address of an empty chunk. solc has a special case that returns 32 zero bytes for empty input, and Sourcify has to reproduce the `bzz-raw://` URLs that solc writes into the metadata. The function now returns 64 zero hex digits for empty input.
- Tests: new `hashFunctions.spec.ts` with the Solidity test vectors for empty and non-empty input. The empty-input `ipfsHash` cases run in a child process with a timeout, because the hang is a synchronous loop that a mocha timeout cannot stop, and the package runs mocha with `--no-timeout`. A new `tryToFindPerfectMetadata` case with a whitespace-only source covers the caller path.

## How verification reaches the hang

Any verification request whose standard JSON input contains an empty or whitespace-only Solidity file in the import graph of the target contract can trigger it. The request does not need to match the contract.

1. `Verification.verify` fetches the on-chain runtime bytecode and compiles the submitted sources.
2. `checkForPerfectMetadata` runs when the on-chain auxdata differs from the recompiled auxdata. That is the case for every partial match and for every submission with the wrong sources. Only an exact match skips it.
3. It builds a `SolidityMetadataContract` from the recompiled metadata and all submitted sources. The keccak hashes in that metadata are hashes of the submitted content, so every source in the import graph of the target lands in `foundSources`.
4. `tryToFindPerfectMetadata` decodes the on-chain auxdata. That succeeds for any contract that solc compiled with default metadata settings.
5. It generates content variations for each found source. One variator is `trimEnd`, so an empty or whitespace-only file becomes an empty string.
6. For each variation it rebuilds the `dweb:/ipfs/` URL with `ipfsHash(source.content)`. Default solc metadata always has that URL. With the empty string, this call never returned.

A realistic accidental case is a project with a placeholder file, for example an interfaces file that was emptied but is still imported. solc compiles an empty source unit with warnings only and lists it in the metadata `sources`. When the metadata hash of the deployed contract does not match the recompilation, the verification worker spins forever. The same input can also be sent on purpose, so each such request costs one worker until the job timeout kills it.

## Notes

- `lib-sourcify` is published separately. The server only picks this up after a package release, which is not part of this PR.
- Non-empty results are unchanged. The existing test vectors guard that.


marcocastignoli (approved):
lgtm

PR #2978, opened 2026-09-17
feat(server): per-thread CPU in the runtime stats log line

manuelwedler:
## Summary

Follow-up to #2977. The `Worker runtime stats` log line shows the CPU of the process that no event loop and no task accounts for (`cpu.outsideLoops`), but not which thread uses it. This PR adds per-thread CPU to the line, to find threads that use CPU outside the event loops.

- **Thread names**: the pool worker threads set the Linux thread name `sfy-pool-<threadId>`, and the solc-js compiler worker sets `solcjs-<version>` (for example `solcjs-0.4.11`). A thread sets its name with a write to `/proc/thread-self/comm`. The `name` option of `new Worker()` does not do this on Node 22.
- **`threads` and `threadCount` in the stats line**: `logRuntimeStats()` reads `/proc/self/task/<tid>/stat` and `comm` and reports the threads with at least 0.05 cores in the interval, sorted by CPU, at most 8: `{ tid, name, cores, userMs, systemMs, ageS }`. The main thread is reported as `main`. The existing fields do not change.
- **`WORKER_ATOMICS`**: new optional env variable for the Piscina `atomics` option (`sync` | `async` | `disabled`). If it is not set, the option is not passed and the Piscina default (`sync`) stays. Another value stops the server at start-up with an error. The active value is logged when the pool is created.

## Notes

- Linux only. All `/proc` access is in `try/catch`. On other platforms, or if `/proc` is not readable, the server logs one `warn` line and continues without the new fields. The thread name is only a diagnostic aid and fails silently.
- `verificationWorker.ts` is also imported by the main thread (for `filename`), so it sets the thread name only when `!isMainThread`.
- A thread that ends between the `readdir` and the read of its files is skipped. A tid that a new thread uses again is found by its `starttime`.
- The thread CPU is read one time when the stats timer starts, so that the first line does not count the CPU before that point.
- No new dependencies.


marcocastignoli (approved):

pc298335-ops (commented):
Hi

PR #2979, opened 2026-09-17
fix(server): integer worker pool thread counts

manuelwedler:
## Summary

Fixes two defects in the settings of the verification worker pool.

### Fractional thread counts

`VerificationService` computed the pool size like this:

```ts
const minThreads = availableParallelism * 0.5;
const maxThreads = availableParallelism * 1.5;
```

With an odd `availableParallelism` the values are fractions (5 gives 2.5 and 7.5). Piscina starts workers while `workers.size < minThreads` and removes an idle worker at the idle timeout while `workers.size > minThreads`. With 2.5 the two conditions alternate without end: 3 > 2.5 removes a worker, 2 < 2.5 starts a new one. Each start loads the complete module graph of the verification worker. With a short idle timeout an idle server uses a full CPU core for this.

The counts now come from a new pure function, `getWorkerPoolThreadCounts`, which returns integers: `minThreads = max(1, floor(n * 0.5))` and `maxThreads = max(minThreads, ceil(n * 1.5))`. For 5 CPUs this gives 2 and 8.

Reproduction of the loop (any Linux or macOS machine):

```js
const Piscina = require("piscina");
const pool = new Piscina({ filename: "./empty-worker.js", minThreads: 2.5, maxThreads: 7.5, idleTimeout: 30 });
let created = 0;
pool.on("workerCreate", () => created++);
setTimeout(() => console.log(created), 10000); // many workers; with minThreads 3 it prints 3
```

### Wrong unit in the `WORKER_IDLE_TIMEOUT` documentation

`.env.dev` documented `WORKER_IDLE_TIMEOUT` in seconds ("If not set, uses 30 s as default"), but the server passes the value unchanged to Piscina, which reads milliseconds. A value of `30` gives 30 ms, which also makes the loop above very fast. The behaviour does not change: `.env.dev` now says that the value is in milliseconds.

### Pool settings log

The `Initialized the verification worker pool` log at startup now also has `availableParallelism` and `idleTimeoutMs`, next to `minThreads` and `maxThreads`.

## Tests

- Unit test for `getWorkerPoolThreadCounts` with 1 to 8 CPUs.
- Regression test: a `VerificationService` for 5 CPUs with an idle timeout of 20 ms and an idle pool must not remove or start workers for 1 s. Without the fix this test fails.


services/server/src/server/services/VerificationService.ts
```diff
@@ -236,9 +245,16 @@ export class VerificationService {
...
+    // Default values of Piscina. They must be integers: with a fraction,
+    // Piscina stops and starts an idle worker again and again.
+    const { minThreads, maxThreads } =
+      getWorkerPoolThreadCounts(availableParallelism);
+    logger.info("Verification worker pool settings", {
+      availableParallelism,
+      minThreads,
+      maxThreads,
+      idleTimeoutMs: options.workerIdleTimeout || 30000,
+    });
```
marcocastignoli: This log overlaps with the existing `Initialized the verification worker pool` log a few lines below, which already prints `minThreads` and `maxThreads`. At startup we now get two consecutive lines with mostly the same data. I would drop this one and add `availableParallelism` and `idleTimeoutMs` to the existing log instead. That also lets you compute `options.workerIdleTimeout || 30000` once in a `const idleTimeout` and reuse it for both the Piscina option and the log, so the two cannot drift apart.

marcocastignoli (approved):

services/server/src/server/services/VerificationService.ts
manuelwedler: Done in 40b9c23f: the separate log is gone. `Initialized the verification worker pool` now also prints `availableParallelism` and `idleTimeoutMs`, and one `const idleTimeout` feeds the Piscina option and the log.

